PR #13, opened 2026-01-07
feat: add Claude Code marketplace plugin support

willist:
Add complete marketplace plugin configuration and documentation for GSD, enabling installation via Claude Code's plugin marketplace.

Changes:
- Add .claude-plugin/marketplace.json and plugin.json manifests
- Add marketplace installation instructions to README.md
- Update all command files with marketplace tool permissions
- Add plugin conversion design and implementation documentation
- Update .gitignore for Claude Code plugin artifacts

Users can now install via:
 ```
/plugin marketplace add glittercowboy/get-shit-done
/plugin install get-shit-done@get-shit-done
```

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -25,7 +25,7 @@
     "./commands/gsd/pause-work.md",
     "./commands/gsd/resume-work.md",
     "./commands/gsd/consider-issues.md",
-    "./commands/gsd/help.md",
-    "./commands/gsd/auto.md"
-  ]
+    "./commands/gsd/help.md"
+  ],
+  "hooks": "./hooks/hooks.json"
 }
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -1,0 +1,14 @@
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "ln -sf ${CLAUDE_PLUGIN_ROOT} ~/.claude/get-shit-done"
+          }
+        ]
+      }
+    ]
+  }
+}

### CODE REVIEW

**Comment by glittercowboy** [comment]
Thanks for putting this together — the plugin structure validates correctly and installs fine. I tested it locally:

```bash
claude plugin validate .                    # ✓ passes
claude plugin marketplace add ./            # ✓ works
claude plugin install get-shit-done@get-shit-done  # ✓ installs
```

**However, there's a path resolution issue that breaks the commands.**

GSD commands use `@~/.claude/get-shit-done/workflows/...` references to load resources. When installed via npm, these files exist at `~/.claude/get-shit-done/`. But when installed via plugin, they're at `~/.claude/plugins/cache/get-shit-done/.../get-shit-done/` — so the `@` references don't resolve.

I tested alternatives:
- **Relative paths from command location** — Claude Code doesn't support this
- **CWD-relative paths** — Would break depending on where user runs Claude

**Possible solutions:**

1. **SessionStart hook** — Plugin defines a hook that symlinks `~/.claude/get-shit-done` → plugin cache location on first run. Works but feels hacky.

2. **Wait for Claude Code** — If Anthropic adds support for relative `@` paths from command file location, this becomes trivial.

3. **Dual resource location** — Commands check both paths somehow (not sure this is even possible with `@` syntax).

I'm not closing this because the work is good and the plugin system is the future. But as-is, the commands would fail to load their workflow/template resources.

If you want to pursue the SessionStart hook approach, I can share the implementation details. Otherwise, happy to keep this open until there's a cleaner path forward.

willist:
Added a SessionStart hook, but I understand if you'd prefer to wait for claude code. It looks like this should be possible with `${CLAUDE_PLUGIN_ROOT}` but [this bug](https://github.com/anthropics/claude-code/issues/9354) is preventing that from working.

glittercowboy:
Appreciate the thorough work here - the plugin structure is solid and you clearly understand the system.

Closing for now because the npx installer already provides a one-command install experience, and adding the SessionStart symlink workaround creates maintenance overhead for marginal benefit.

If/when Claude Code fixes the `${CLAUDE_PLUGIN_ROOT}` bug (#9354), I'd happily revisit this - at that point it becomes a clean addition rather than a workaround.

Thanks for contributing!

cowwoc:
Take another look. Per my local tests, Version 2.1.1 (which came out earlier on today) seems to fix this issue. In my personal fork of this codebase, I converted it into a plugin and used `@${CLAUDE_PLUGIN_ROOT}` inside command files. Local tests showed it working as expected.

PR #14, opened 2026-01-07
Port GSD to Antigravity workflows and templates

tezza1971:
Workflows can be invoked as per CC's slash commands.


tezza1971:
I suppose this means that the brains would have to be maintained in two places now. There would implementation variations between AI tools anyway, so that isn't the wrong way to go. I suppose you can just tell Claude to keep both sets of files updated whilst being sensitive to [tool-based] implementation differences.

tezza1971:
Needs more testing

PR #150, opened 2026-01-18
Add Codex installer and per-project seed prompt

oldnordic:
## Summary\n- add a --codex installer path that seeds Codex home and provides a /prompts:gsd-install entrypoint\n- add Codex seed assets (commands, workflows, templates) with .codex/skills/gsd\n- document Codex install + per-project usage in CODEX-PORT.md\n\n## Testing\n- node --check bin/install.js


markvandeven:
while i think this would be a great workaround to keep codex users in sync with the latest GSD changes, what would be even better is if we can select the model (and runtime) for certain steps. For example as you are more likely to hit rate limiting with Claude than with Codex at the moment, I would like to have a Codex instance do the research work and requirement specs and leave Claude to orchestration and coding.

PR #167, opened 2026-01-19
fix(hooks): prevent console window flash on Windows

superresistant:
## Summary
- Fix console window flash on Windows when GSD update check runs at session start
- The `detached: true` spawn option was creating a visible console window
- Combined with `npm` (which is a `.cmd` batch file), this caused focus stealing

## Changes
- Remove `detached: true` from spawn options (not needed, `child.unref()` handles non-blocking)
- Add `windowsHide: true` to both `spawn()` and inner `execSync()` calls

## Test plan
- [ ] Run `claude` on Windows Terminal with Git Bash
- [ ] Verify no console window flash appears on session start
- [ ] Verify GSD update check still works (check `~/.claude/cache/gsd-update-check.json`)

### DIFF
--- a/hooks/gsd-check-update.js
+++ b/hooks/gsd-check-update.js
@@ -17,7 +17,7 @@ if (!fs.existsSync(cacheDir)) {
   fs.mkdirSync(cacheDir, { recursive: true });
 }
 
-// Run check in background (spawn detached process)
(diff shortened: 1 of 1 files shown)


superresistant:
It is a very annoying bug on windows. The fix is simple.

PR #219, opened 2026-01-21
Correct CONTEXT.md to XX-CONTEXT.md in workflow

oetiker:
Updated the context file name in the phase planning note. Otherwise, Claude will try to read CONTEXTG.md and not find it.

### DIFF
--- a/get-shit-done/workflows/resume-project.md
+++ b/get-shit-done/workflows/resume-project.md
@@ -200,7 +200,7 @@ What would you like to do?
 **Note:** When offering phase planning, check for CONTEXT.md existence first:
 
 ```bash
-ls .planning/phases/XX-name/CONTEXT.md 2>/dev/null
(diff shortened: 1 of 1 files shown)


glittercowboy:
Thanks for catching this inconsistency!

The fix is directionally correct, but there's historical inconsistency in CONTEXT.md naming — some files use `CONTEXT.md`, others use `03-CONTEXT.md` (with phase prefix).

The robust pattern that `discuss-phase.md` uses checks multiple variants. For this example, the simplest fix would be:

```bash
ls .planning/phases/XX-name/*CONTEXT.md 2>/dev/null
```

The glob `*CONTEXT.md` catches both naming conventions.

Could you update to use that pattern?

glittercowboy:
Thanks for catching this! Fixed with glob pattern in bb93172 to handle both naming conventions.

PR #224, opened 2026-01-22
feat(opencode): make installer work for opencode

dbachelder:
Adds --opencode flag to the installer for compatibility with https://opencode.ai, an open-source alternative to Claude Code. When this flag is
  used, the installer automatically converts Claude Code's configuration format to OpenCode's expected format.

  Key conversions:
  - YAML frontmatter: allowed-tools: array → tools: object with boolean values
  - Tool names: AskUserQuestion → question, SlashCommand → skill, lowercase (Read → read)
  - Colors: Named colors → hex format (cyan → "#00FFFF")
  - Paths: ~/.claude → ~/.opencode
  - Commands: /gsd:command → /gsd/command
  - Removes name: field (OpenCode uses filename for command names)
  - Tool references in prompt body text are also converted

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -15,6 +15,13 @@ const reset = '\x1b[0m';
 // Get version from package.json
 const pkg = require('../package.json');
 
+// Parse args
+const args = process.argv.slice(2);
+const hasGlobal = args.includes('--global') || args.includes('-g');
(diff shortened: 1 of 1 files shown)


superresistant:
guys does it work? would be amazing for adoption, we can communicate on that

superresistant:
congrats

PR #233, opened 2026-01-22
fix(installer): use correct OpenCode config path (~/.config/opencode)

frontendwizard:
## What

Fix OpenCode installer to use the correct XDG-compliant config path (`~/.config/opencode/`) instead of `~/.opencode/`.

## Why

OpenCode follows XDG Base Directory spec and expects its config at `~/.config/opencode/`. The 1.9.6 installer was writing to `~/.opencode/` where OpenCode couldn't find the commands.

## Changes

- Add `getOpencodeGlobalDir()` to detect correct path via env vars (OPENCODE_CONFIG_DIR, OPENCODE_CONFIG, XDG_CONFIG_HOME) or default to `~/.config/opencode`
- Use `command/` (singular) instead of `commands/` (plural) for OpenCode
- Use flat file structure: `command/gsd-help.md` instead of nested `commands/gsd/help.md`
- Write permissions to `~/.config/opencode/opencode.json`
- Update content conversion: `~/.claude` → `~/.config/opencode`, `/gsd:` → `/gsd-`

## Testing

- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

Verified installation creates correct structure:
```
~/.config/opencode/
├── command/
│   ├── gsd-help.md
│   ├── gsd-new-project.md
│   └── ... (27 files)
├── agents/
│   └── gsd-*.md
├── get-shit-done/
└── opencode.json
```

## Checklist

- [x] Works on Windows (backslash paths tested) - path handling uses Node's path.join

## Breaking Changes

None - this is a bug fix. Users who installed 1.9.6 for OpenCode will need to reinstall.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- OpenCode installer now uses correct config path (`~/.config/opencode/`) instead of `~/.opencode/`
+- OpenCode commands use correct directory structure (`command/gsd-help.md` not `commands/gsd/help.md`)
+- OpenCode permissions written to `~/.config/opencode/opencode.json` instead of `~/.opencode.json`
+
 ## [1.9.6] - 2026-01-22
 
 ### Added
--- a/bin/install.js
+++ b/bin/install.js
@@ -33,9 +33,58 @@ if (hasBoth) {
   selectedRuntimes = ['claude'];
 }
 
-// Helper to get directory name for a runtime
+// Helper to get directory name for a runtime (used for local/project installs)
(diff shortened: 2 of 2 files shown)


dpearson2699:
@glittercowboy this is extremely needed because when using the installer, if you previously had the gsd-opencode installed, selecting the opencode install, thinking it would update the old gsd-opencode, actually creates two copies of each command, breaking gsd for opencode.

dbachelder:
@frontendwizard the existing installer is working on my system.. opencode is ignoring the ~/.config/opencode dir entirely is seems.

frontendwizard:
@dbachelder It depends on wether or not you have the XDG_CONFIG_HOME set to use `.config`. This fix makes sure the installation path follows the documented priority order opencode does when looking for it's config files.

PR #242, opened 2026-01-22
feat: add Cursor IDE as supported runtime

meenie:
## What

As of the latest update to cursor, they now support Agent Skills and subagents. In light of that, add Cursor IDE as a supported runtime for GSD installation alongside Claude Code and OpenCode. Here's a screenshot showing how it's using sub-agents to do research:
image

I've tried the result of this in Cursor for a couple of tasks and it seems to work quite well. It's not always using subagents for the execution, though, so this might not yet be ready for merging. Wanted to get this on your radar to see if you have any interest in doing this in the first place.

## Why

Cursor supports Claude-based agents and the existing GSD commands/subagents already work in Cursor — this adds first-class installation support.

## Known Issues

1. Cursor slash commands do not work with a colon. So the commands end up being used like `/gsd/add-phase`. Annoyingly, when typing it out, you can't use a second slash after the first one or the filtered list goes away 🤦🏻‍♂️.
2. In the "Next Up" part of the final response to commands like planning a phase, it mentions to use `/clear`. That is not a command in Cursor, so we should probably grep that during the install process for Cursor and change it to say that the user should create a new new chat instead.

## Testing

- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Checklist

## Breaking Changes

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # GET SHIT DONE
 
-**A light-weight and powerful meta-prompting, context engineering and spec-driven development system for Claude Code by TÂCHES.**
+**A light-weight and powerful meta-prompting, context engineering and spec-driven development system for Claude Code, Cursor, and OpenCode by TÂCHES.**
 
 **Solves context rot — the quality degradation that happens as Claude fills its context window.**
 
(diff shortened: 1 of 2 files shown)


meenie:
I'm wondering if the install script should be modularized now that there are 3 install targets... If you want, @glittercowboy, I can get Claude on that :). Wanted to make this a straightforward change.

meenie:
Seems to be following directions quite well. It even figured out in the plan verification subagent that it needed to revise it's plans and it spawned another planning subagent.
image

meenie:
It looks like cursor picks up the slash commands the like from the claude directory (if they are local), so I suppose this is not really needed. Going to close it and someone else can pick it up if needed.

PR #245, opened 2026-01-23
fix: support both MCP tool naming patterns for context7

shaunchurch:
## Summary
- Add `mcp__plugin_context7_context7__*` pattern alongside existing `mcp__context7__*`
- Fixes agents not getting Context7 MCP tools when Claude Code uses full plugin naming

## Problem
MCP tools can be named either:
- `mcp__context7__*` (shorthand)
- `mcp__plugin_context7_context7__*` (full: `mcp__plugin_<plugin>_<server>__*`)

The pattern depends on how Claude Code resolves plugin names. Currently only the shorthand is listed, causing agents to miss Context7 access in some environments.

## Test plan
- [ ] Spawn `gsd-phase-researcher` and verify it has `mcp__context7__*` tools
- [ ] Test Context7 lookup works within agent

### DIFF
--- a/agents/gsd-phase-researcher.md
+++ b/agents/gsd-phase-researcher.md
@@ -1,7 +1,7 @@
 ---
 name: gsd-phase-researcher
 description: Researches how to implement a phase before planning. Produces RESEARCH.md consumed by gsd-planner. Spawned by /gsd:plan-phase orchestrator.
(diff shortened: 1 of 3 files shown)


shaunchurch:
Closing - further investigation showed this isn't the fix. Custom agents defined in ~/.claude/agents/ don't receive MCP tools regardless of the tools: field pattern. This is a Claude Code limitation, not a GSD issue.

PR #284, opened 2026-01-24
feat(git): add branch-per-phase option

### LINKED ISSUE #173: Never commit to main, always create a branch for each new project, even for each phase?
I wasn't expecting so many commits. I think it will be interesting every time a new project is created that all of these commits happen in a new branch so they can then be squashed. If it happens on main it's a pain in the ass .

davesienkowski:
## Summary

Implements optional git branching workflow for GSD phases (closes #173):

- **New config options**: `git.branch_per_phase` (default: false) and `git.branch_template` (default: `gsd/phase-{phase}-{slug}`)
- **execute-phase**: Creates/switches to phase branch when enabled before execution begins
- **settings**: Adds git branching toggle to `/gsd:settings` command
- **complete-milestone**: Offers merge options (merge all, delete, keep) when phase branches exist

## Files Changed

- `get-shit-done/references/planning-config.md` - Added git config schema and documentation
- `get-shit-done/workflows/execute-phase.md` - Added branch creation step after load_project_state
- `commands/gsd/settings.md` - Added git branching question to settings
- `get-shit-done/workflows/complete-milestone.md` - Added branch merge handling step

## Test Instructions

1. Create a new GSD project
2. Run `/gsd:settings` and enable "Branch Per Phase"
3. Run `/gsd:execute-phase` and verify branch creation
4. Run `/gsd:complete-milestone` and verify branch merge options appear

### DIFF
--- a/commands/gsd/settings.md
+++ b/commands/gsd/settings.md
@@ -34,6 +34,8 @@ Parse current values (default to `true` if not present):
 - `workflow.plan_check` — spawn plan checker during plan-phase
 - `workflow.verifier` — spawn verifier during execute-phase
 - `model_profile` — which model each agent uses (default: `balanced`)
+- `git.branch_per_phase` — create branch per phase (default: `false`)
+- `git.branch_template` — branch naming template (default: `"gsd/phase-{phase}-{slug}"`)
 
 ## 3. Present Settings
 
@@ -77,6 +79,15 @@ AskUserQuestion([
       { label: "Yes", description: "Verify must-haves after execution" },
       { label: "No", description: "Skip post-execution verification" }
     ]
+  },
+  {
+    question: "Create git branch for each phase?",
+    header: "Branching",
+    multiSelect: false,
+    options: [
+      { label: "No (Recommended)", description: "Commit directly to current branch" },
+      { label: "Yes", description: "Create gsd/phase-{phase}-{slug} branch per phase" }
+    ]
   }
 ])
 ```
@@ -95,6 +106,9 @@ Merge new settings into existing config.json:
     "research": true/false,
     "plan_check": true/false,
     "verifier": true/false
+  },
+  "git": {
+    "branch_per_phase": true/false
   }
 }
 ```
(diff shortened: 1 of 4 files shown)


simfor99:
Great idea! 👍🏼📐

szymontex:
It would be enough  per milestone  for me. The phases are still quite small steps toward actually shipping the feature - especially on existing projects. After each milestone there’s usually something to test anyway.

simfor99:
Branching isn't about adding overhead; it's about decoupling your focus.
It enables true parallel development: you can hammer away at a complex Payment integration in one branch while simultaneously refining the UI in another. If the payment logic hits a snag, your main codebase remains "ship-ready" and untouched. This transforms a linear, fragile process into a modular safety net, ensuring that small phases don't become big bottlenecks.

szymontex:
Yeah I’m not anti-branching at all - branching is useful.
I just think branch per phase is a bit too granular for my workflow (especially on existing projects). Phases tend to be tiny steps, and the churn of switching/merging/cleaning up branches can end up being more work than the phase itself.
For me branch per milestone is the sweet spot: milestones are usually the point where there’s actually something meaningful to test/integrate, and it still keeps main in a ship-ready state without making every micro-step its own branch.

davesienkowski:
Perhaps maybe a config setting to choose how and when you'd like to branch?

These would all be disabled by default. Power user settings essentially.

szymontex:
@davesienkowski That would be amazing, thank you

davesienkowski:
I'll see about modifying this to include that option. Makes sense to have that be user choice if it'll be disabled by default anyway. At least until/if GSD adopts a standard for branching.

davesienkowski:
Superseded by #298 which adds a unified branching strategy with support for both per-phase and per-milestone branching.

szymontex:
Yeah,
Like in practice - on a brownfield project I usually end up with something like ~100 commits per milestone, split across 4–6 phases. On a greenfield one it can be ~150 commits per milestone and it might take 7 phases.

On brownfield, I don’t really need to push into a separate branch between phases - I can already test most of the stuff in my dev env as I go.

And on greenfield, there’s often no point pushing anything early on because you’re still setting things up, and you don’t always even have the full repo/pipeline/story ready until you’re closer to the first milestone.

PR #289, opened 2026-01-24
fix: add character preservation rules and clarify ASCII terminology

szymontex:
## Summary
- Add Character Preservation section to GSD-STYLE.md
- Clarify 'no special chars' applies to FILENAME only in add-todo
- Use 'box-drawing' terminology for tree structure visualization

## What
Clarifies that "ASCII" in GSD templates refers only to box-drawing characters (├── └── │) for tree visualization, not to stripping diacritics from text content.

## Why
Claude was incorrectly stripping Polish/German/French diacritics (ą, ę, ü, é, etc.) from generated content when seeing "ASCII" in template instructions.

## Testing
- [x] Tested on Linux

## Checklist
- [x] Follows GSD style (no enterprise patterns, no filler)
- [ ] Updates CHANGELOG.md for user-facing changes
- [x] No unnecessary dependencies added
- [x] Works on Windows (backslash paths tested)

## Breaking Changes
None

PR #287, opened 2026-01-24
fix(agents): replace context7 wildcard with explicit tool names

dkmaker:
## What

Replace non-functional `mcp__context7__*` wildcard patterns in agent/command tool declarations with explicit tool names, supporting both direct MCP server and Claude Code plugin installations.

## Why

Claude Code's tool declaration system does not support wildcard (`*`) patterns in tool names. The `mcp__context7__*` entries in agent `tools:` and command `allowed-tools:` fields were silently ignored — meaning Context7 was **never actually available** to any subagent or slash command, despite being referenced in their instructions.

Additionally, `discovery-phase.md` referenced a non-existent tool name (`mcp__context7__get-library-docs`) with incorrect parameter names.

## Changes

**6 files updated:**

| File | Change |
|------|--------|
| `agents/gsd-project-researcher.md` | Wildcard → explicit tools + plugin variants in usage docs |
| `agents/gsd-phase-researcher.md` | Same |
| `agents/gsd-planner.md` | Wildcard → explicit tools |
| `agents/gsd-debugger.md` | Added context7 tools (was missing entirely despite referencing Context7 in its instructions) |
| `commands/gsd/plan-phase.md` | Wildcard → explicit entries in allowed-tools |
| `get-shit-done/workflows/discovery-phase.md` | Fixed wrong tool name (`get-library-docs` → `query-docs`), fixed parameter names, added plugin variants |

**Tool names added (both install methods supported):**
- `mcp__context7__resolve-library-id` / `mcp__context7__query-docs` (direct MCP server)
- `mcp__plugin_context7_context7__resolve-library-id` / `mcp__plugin_context7_context7__query-docs` (Claude Code plugin)

Also converted comma-separated `tools:` lines to YAML arrays for readability.

## Testing

- Verified YAML frontmatter parses correctly in all modified files
- Context7 tools are now explicitly declared and will be available to subagents at runtime
- Both installation methods (direct MCP server config and Claude Code plugin) are supported

## Breaking Changes

None

PR #278, opened 2026-01-24
feat(verifier): add provides/consumes verification and enhanced key_links checking

davesienkowski:
## Summary

- Add Step 2.5 with `verify_provides()` and `verify_consumes()` functions to check that declared exports exist and imports are present
- Enhance Step 5 with `verify_key_link_pattern()` for precise pattern-based wiring verification with better error messages
- Add `relationship_failures` field to gaps YAML for detailed diagnosis
- Update VERIFICATION.md output with Provides/Consumes tables and enhanced Key Link verification

### New Step 2.5: Verify Provides/Consumes

When PLAN.md frontmatter contains `provides` or `consumes` fields:
- `verify_provides(file, export_name)` checks the file actually exports what it claims
- `verify_consumes(consumer_dir, artifact, import_path)` checks consuming code imports the artifact

### Enhanced Step 5: Pattern-Based Key Links

- New `verify_key_link_pattern()` function uses pattern field for precise verification
- Better error messages with file, expected pattern, and suggestion

### Enhanced Outputs

- VERIFICATION.md frontmatter now includes `provides_verification`, `consumes_verification`, and `key_links_verification` fields
- Gaps YAML includes `relationship_failures` for detailed diagnosis of what went wrong
- New Provides/Consumes table section in markdown output

## Test plan

- [x] Verify gsd-verifier works normally when provides/consumes not present (backwards compatible)
- [x] Test with PLAN.md containing provides field - should check exports exist
- [x] Test with PLAN.md containing consumes field - should check imports exist
- [x] Test key_links with pattern field - should use pattern for verification
- [x] Verify VERIFICATION.md output includes new sections when applicable

PR #277, opened 2026-01-24
feat(templates): add exports field guidance with verifiable identifier examples

davesienkowski:
## Summary

Update phase-prompt.md to clarify that exports field should contain actual exported identifiers (function names, component names, types) rather than generic descriptions.

## Changes

- Update exports field description in table to specify "actual identifiers"
- Add guidance section with good/bad examples
- Explain why specific exports enable gsd-verifier grep verification
- Update example artifacts to demonstrate specific exports pattern

## Test plan

- [ ] Verify exports guidance is clear
- [ ] Confirm examples show good vs bad patterns
- [ ] Test that gsd-verifier can grep for specific exports

Implements enhancement 04 from `.planning/research/plan-context-improvements/`

PR #296, opened 2026-01-24
fix: scale context bar in statusline to show 100% at actual 80% limit

kaladivo:
## The problem: 

At 80% of statusline I already got an error from calude code that it run out of context. 

Claude Code intentionally reserves 20% of the context window as a buffer for:                                                                     
  - Claude's response generation                                           
  - System prompts and formatting                                          
  - Tool use overhead                                                      
  - Preventing mid-response cutoffs                                        
                                                                           
  To make the bar reach 100% at the actual limit, you would modify your statusline to scale the percentage:                                      
  - Calculate: (used_percentage / 80) * 100                                
  - This way, when you hit the real 80% limit, your bar shows 100% 

## Summary

- Scales the context bar percentage so it shows 100% when reaching Claude Code's actual 80% context limit
- Adjusts color thresholds to maintain the same visual progression

## Problem

Claude Code enforces an 80% context window limit as a safety mechanism. The statusline was showing the raw percentage, which meant the bar never reached 100% before Claude ran out of context - this was confusing as users would see ~80% and then suddenly hit the limit.

## Solution

Scale the percentage calculation: `(rawUsed / 80) * 100`

This means:
- 0% real usage = 0% displayed
- 40% real usage = 50% displayed  
- 80% real usage (the actual limit) = 100% displayed

Color thresholds are adjusted accordingly:
- Green: 0-62% (was 0-50% raw)
- Yellow: 63-80% (was 50-65% raw)
- Orange: 81-94% (was 65-80% raw)
- Red with skull: 95%+ (was 80%+ raw)

## Test plan

- [ ] Run with fresh context and verify bar starts at 0%
- [ ] Use context and verify bar fills proportionally
- [ ] Verify bar reaches 100% when Claude reports context exhaustion

---

### DIFF
--- a/hooks/gsd-statusline.js
+++ b/hooks/gsd-statusline.js
@@ -18,22 +18,25 @@ process.stdin.on('end', () => {
     const session = data.session_id || '';
     const remaining = data.context_window?.remaining_percentage;
 
-    // Context window display (shows USED percentage)
+    // Context window display (shows USED percentage scaled to 80% limit)
+    // Claude Code enforces an 80% context limit, so we scale to show 100% at that point
     let ctx = '';
     if (remaining != null) {
       const rem = Math.round(remaining);
-      const used = Math.max(0, Math.min(100, 100 - rem));
+      const rawUsed = Math.max(0, Math.min(100, 100 - rem));
+      // Scale: 80% real usage = 100% displayed
+      const used = Math.min(100, Math.round((rawUsed / 80) * 100));
 
       // Build progress bar (10 segments)
       const filled = Math.floor(used / 10);
       const bar = '█'.repeat(filled) + '░'.repeat(10 - filled);
 
-      // Color based on usage
-      if (used < 50) {
+      // Color based on scaled usage (thresholds adjusted for new scale)
+      if (used < 63) {        // ~50% real
         ctx = ` \x1b[32m${bar} ${used}%\x1b[0m`;
-      } else if (used < 65) {
+      } else if (used < 81) { // ~65% real
         ctx = ` \x1b[33m${bar} ${used}%\x1b[0m`;
-      } else if (used < 80) {
+      } else if (used < 95) { // ~76% real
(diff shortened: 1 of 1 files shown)


balandari:
Great fix! We independently researched this and found similar results.

Our data point: Claude Code's autocompact buffer is ~22.5% (45k of 200k), meaning compaction triggers at **77.5%**, not 80%. Your empirical 80% observation means you were already past the compaction threshold when you hit the error.

**Implication:** Using 80% as the scaling factor means users will see 100% displayed *after* compaction has already occurred (~2.5% late). For true accuracy, 77.5% would be the precise threshold, though 80% is easier to remember and close enough for practical use.

+1 for configurability via env var - exact thresholds may vary across versions/models, and users could tune to 77.5 if they want precision.

**Edit:** Corrected our earlier statement - 80% is not "a safety margin," it's actually past the trigger point. There is no buffer; compaction has already happened.

---
*This comment was written by Claude (Opus 4.5) on behalf of @balandari as part of our AI-assisted development workflow.*

superresistant:
If you wanna try, I did merge many recent PR fixes on this fork https://github.com/superresistant/get-shit-done
After testing if it works it'll be merged. 


#296 | fix: scale context bar in statusline to show 100% | @kaladivo
#287 | fix(agents): replace context7 wildcard with explicit names | @dkmaker
#268 | fix: add fallback to source hooks when dist/ missing | @heltonteixeira
#186 | refactor(agents): use @-references for codebase-mapper | @jjshanks
#289 | fix: add character preservation rules | @szymontex
#288 | fix(install): auto-migrate renamed statusline | @superresistant
#286 | feat(install): respect attribution.commit | @superresistant
#277 | feat(templates): add exports field guidance | @davesienkowski
#278 | feat(verifier): add provides/consumes verification | @davesienkowski
#275 | feat(mapper): add Feature Map section | @davesienkowski

---

npx github:superresistant/get-shit-done

PR #342, opened 2026-01-29
Gsd for Cursor

rmindel:
## What

Add reference to the Cursor adaptation

## Testing

- [ V ] Tested on Windows

## Checklist

- [ V ] Follows GSD style (no enterprise patterns, no filler)
- [ V ] No unnecessary dependencies added
- [ V ] Works on Windows (backslash paths tested)

## Breaking Changes

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,7 +572,7 @@ This removes all GSD commands, agents, hooks, and settings while preserving your
 |---------|----------|-------------|
 | [gsd-opencode](https://github.com/rokicool/gsd-opencode) | OpenCode | GSD adapted for OpenCode CLI |
 | [gsd-gemini](https://github.com/uberfuzzy/gsd-gemini) | Gemini CLI | GSD adapted for Google's Gemini CLI |
-
+| [gsd-for-cursor](https://github.com/rmindel/gsd-for-cursor) | Cursor | GSD adapted for Cursor |
 ---
 
 ## Star History


glittercowboy:
Thanks for this! Before merging the README link, I wanted to float an idea:

Looking at your adaptation, the changes for Cursor (config dir, command syntax, tool naming) are similar to what we already handle for OpenCode and Gemini in `bin/install.js`.

Would you be interested in contributing Cursor support directly to the installer? Something like:

```bash
npx get-shit-done-cc --cursor --global
```

The installer already has conversion logic for:
- Path replacements (`~/.claude/` → target dir)
- Tool name transformations (PascalCase → lowercase for OpenCode)
- Frontmatter conversion (`allowed-tools:` → `tools: {}`)
- Color name → hex mapping

For Cursor, we'd need to add:
- `~/.cursor` as the config directory
- Command syntax conversion (`/gsd:cmd` → `/gsd/cmd` if that's what Cursor uses)
- Any Cursor-specific tool name mappings

Take a look at `bin/install.js` - specifically `convertClaudeToOpencodeFrontmatter()` and how OpenCode/Gemini are handled. If you're up for it, a PR adding `--cursor` support would be more maintainable than a separate repo that could drift out of sync.

Let me know what you think!

rmindel:
That's a great idea.

I'll try merging my migration script into the installer.

rmindel:
https://github.com/glittercowboy/get-shit-done/pull/438

PR #378, opened 2026-02-01
fix: researcher agent always writes RESEARCH.md to disk

### LINKED ISSUE #343: gsd-phase-researcher doesn't persist RESEARCH.md to disk
## Bug Summary

Researcher agent returns `## RESEARCH COMPLETE` claiming file was created, but never calls Write tool to persist RESEARCH.md.

## Environment

- **GSD Version:** 1.9.13
- **Model:** sonnet (balanced profile)
- **Config:** `commit_docs: false` (planning dir gitignored)

## Steps to Reproduce

1. Run `/gsd:plan-phase <N>` on phase needing research
2. Researcher spawns, executes (41 tool uses, 5m 37s in my case)
3. Returns structured output with "File Created: `.../<phase>-RESEARCH.md`"
4. File doesn't exist on disk

## Expected Behavior

Researcher should use Write tool to persist RESEARCH.md before returning, regardless of `commit_docs` setting.

## Actual Behavior

Agent return message includes research findings and claims file was written, but Write tool was never invoked. File doesn't exist.

## Suspected Root Cause

Agent may be misinterpreting `commit_docs: false` as "skip file write" rather than just "skip git commit".

From agent output:
```
Planning docs are gitignored and commit_docs is false. Skip git commit per agent instructions.
```

## Workaround

Orchestrator can extract research from agent return message and inline it for the planner (or manually write to disk). Planning succeeds but audit trail is lost.

## Impact

- **Cost:** Expensive agent activity (~5min, 50k+ tokens) produces no artifact
- **Audit:** No RESEARCH.md for future reference
- **Reliability:** Requires orchestrator workaround

Solvely-Colin:
## Summary
Fixes #343

The `gsd-phase-researcher` agent was misinterpreting `commit_docs=false` as "skip file write" when it should only skip git commit operations. This caused expensive research runs (~5min, 50k+ tokens) to produce no artifact.

## Root Cause
Step 5 said "Write to: $PHASE_DIR/..." but didn't explicitly require the Write tool. Step 6 said "If commit_docs=false: Skip git operations" which the agent interpreted as skipping the entire output.

## Changes
- Explicitly state **Write tool is mandatory** in Step 5
- Clarify `commit_docs` only affects git operations, not file writes  
- Rename Step 6 to "Commit Research (optional)"
- Make the file-write-then-commit order explicit

## Testing
1. Set `commit_docs: false` in `.planning/config.json`
2. Run `/gsd:plan-phase` on a phase needing research
3. Verify RESEARCH.md is created on disk even without git commit

### DIFF
--- a/agents/gsd-phase-researcher.md
+++ b/agents/gsd-phase-researcher.md
@@ -523,15 +523,19 @@ Run through verification protocol checklist:
 
 ## Step 5: Write RESEARCH.md
 
+**ALWAYS use the Write tool to persist RESEARCH.md to disk.** This is mandatory regardless of `commit_docs` setting.
+
 Use the output format template. Populate all sections with verified findings.
 
 Write to: `$PHASE_DIR/$PADDED_PHASE-RESEARCH.md`
 
 Where `PHASE_DIR` is the full path (e.g., `.planning/phases/01-foundation`)
 
-## Step 6: Commit Research
+⚠️ **The `commit_docs` setting only controls git commits, NOT file writing.** Always write the file first.
 
-**If `COMMIT_PLANNING_DOCS=false`:** Skip git operations, log "Skipping planning docs commit (commit_docs: false)"
+## Step 6: Commit Research (optional)
+
+**If `COMMIT_PLANNING_DOCS=false`:** Skip git operations only. The file MUST already be written in Step 5.
 
 **If `COMMIT_PLANNING_DOCS=true` (default):**
 


Solvely-Colin:
## Test Results

### How I Tested
Compared the agent instructions before and after the fix to verify the ambiguity is resolved.

### Before (main branch)
Step 5 instructions were ambiguous:
```markdown
## Step 5: Write RESEARCH.md
Use the output format template. Populate all sections with verified findings.
Write to: `$PHASE_DIR/$PADDED_PHASE-RESEARCH.md`

## Step 6: Commit Research
**If `COMMIT_PLANNING_DOCS=false`:** Skip git operations, log "Skipping planning docs commit"
```

The agent could interpret "Write to:" as informational (where output goes) rather than imperative (use Write tool). Combined with Step 6 saying "skip" for `commit_docs=false`, agents skipped the entire output.

### After (fix branch)
Step 5 is now explicit:
```markdown
## Step 5: Write RESEARCH.md
**ALWAYS use the Write tool to persist RESEARCH.md to disk.** This is mandatory regardless of `commit_docs` setting.
...
⚠️ **The `commit_docs` setting only controls git commits, NOT file writing.** Always write the file first.

## Step 6: Commit Research (optional)
**If `COMMIT_PLANNING_DOCS=false`:** Skip git operations only. The file MUST already be written in Step 5.
```

### Results
✅ Fix correctly disambiguates file writing (mandatory) from git commit (optional based on config).

The new instructions make it impossible to misinterpret — "ALWAYS use the Write tool" and "mandatory regardless of commit_docs setting" are unambiguous.

PR #379, opened 2026-02-01
fix: respect parallelization config setting

### LINKED ISSUE #376: Not respecting parallelization settings
I'm seeing the executor kick agents off in parallel, despite having parallelization set to false in config.json.

I'm actually pretty surprised that parallel executors is the default.  Even if they don't need to write to the same files, the executor agents will definitely have to run things like unit tests, formatting / linting, code generation, etc that has a high likelihood of producing side effects for the entire repo's build state and getting confused when it sees unexpected side effects.  In practice, that means that executors get confused more often and wind up wasting tokens and time to resolve their own confusion.

GSD version: 1.11.1

Solvely-Colin:
## Summary
Fixes #376

The `execute-phase` workflow was ignoring the `parallelization` setting from `.planning/config.json` and always running autonomous plans within waves in parallel.

## Root Cause
The workflow read other config settings (model_profile, commit_docs, branching_strategy) but never read the `parallelization` setting.

## Changes
- Read `PARALLELIZATION` from `.planning/config.json` (defaults to `true`)
- When `false`, execute plans within waves sequentially (one at a time)
- Prevents side effects from concurrent operations like tests, linting, and code generation

## Testing
1. Set `"parallelization": false` in `.planning/config.json`
2. Run `/gsd:execute-phase` on a phase with multiple plans in same wave
3. Verify plans execute one at a time, not in parallel

### DIFF
--- a/get-shit-done/workflows/execute-phase.md
+++ b/get-shit-done/workflows/execute-phase.md
@@ -66,6 +66,15 @@ git check-ignore -q .planning 2>/dev/null && COMMIT_PLANNING_DOCS=false
 
 Store `COMMIT_PLANNING_DOCS` for use in git operations.
 
+**Load parallelization config:**
+
+```bash
+# Check if parallelization is enabled (default: true)
+PARALLELIZATION=$(cat .planning/config.json 2>/dev/null | grep -o '"parallelization"[[:space:]]*:[[:space:]]*[^,}]*' | grep -o 'true\|false' || echo "true")
+```
+
+Store `PARALLELIZATION` for use in wave execution step. When `false`, plans within a wave execute sequentially instead of in parallel.
+
 **Load git branching config:**
 
 ```bash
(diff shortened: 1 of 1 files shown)


Solvely-Colin:
## Test Results

### How I Tested
Created a test project with `parallelization: false` in config.json, then ran the exact bash parsing logic from the fix to verify it reads the setting correctly.

### Test Setup
```bash
mkdir -p .planning
cat > .planning/config.json << 'EOF'
{
  "mode": "yolo",
  "depth": "standard",
  "parallelization": false,
  "commit_docs": false,
  "model_profile": "balanced"
}
EOF
```

### Config Parsing Validation

```bash
# Test 1: parallelization: false
echo '{"parallelization": false}' > .planning/config.json
PARALLELIZATION=$(cat .planning/config.json | grep -o '"parallelization"[[:space:]]*:[[:space:]]*[^,}]*' | grep -o 'true\|false' || echo "true")
echo "Result: $PARALLELIZATION"
# Output: false ✅

# Test 2: parallelization: true  
echo '{"parallelization": true}' > .planning/config.json
# Output: true ✅

# Test 3: Key missing (should default to true)
echo '{"mode": "yolo"}' > .planning/config.json
# Output: true ✅

# Test 4: File missing (should default to true)
rm .planning/config.json
# Output: true ✅
```

### Before (main branch)
The workflow never read `parallelization` from config.json — it only read:
- `model_profile`
- `commit_docs`  
- `branching_strategy`

Plans always executed in parallel regardless of user setting.

### After (fix branch)
- Reads `PARALLELIZATION` from config.json during `load_project_state` step
- Defaults to `true` (preserves existing behavior for users without the setting)
- When `false`, instructions direct orchestrator to execute plans within waves sequentially

### Results
✅ Config parsing works correctly for all edge cases:
- Explicit `false` → reads as `false`
- Explicit `true` → reads as `true`
- Missing key → defaults to `true`
- Missing file → defaults to `true`

glittercowboy:
Merging. Thank you sir!

PR #443, opened 2026-02-06
feat(auto-mode): continuous phase execution via Stop hook

mamarx:
## What

Add auto-mode: a signal-file-based Stop hook that chains plan-phase → execute-phase → plan-phase automatically. The developer runs `/gsd:auto on`, starts any GSD command, and the hook keeps the workflow going until a hard stop condition is hit.

## Why

Once phases are defined and the workflow is running, the developer becomes a button-presser:

```
/gsd:plan-phase 2       → wait 5-15 min → /clear
/gsd:execute-phase 2    → wait 10-30 min → /clear
/gsd:plan-phase 3       → wait 5-15 min → /clear
/gsd:execute-phase 3    → wait 10-30 min → /clear
```

For a 4-phase milestone, that is 8 manual interventions where the input is almost always "yes, continue with the suggestion." The creative decisions happened earlier — during project setup, requirements, and discuss-phase. The plan-execute cycle is execution, not decision-making.

**Overnight runs.** Start a milestone before bed, wake up to atomic commits and verification reports.

**Low-priority work.** Database schemas, CRUD endpoints, standard configs — trust Claude to handle these while focusing on other work.

**Flow state protection.** Every manual intervention is a context switch that fragments attention.

## How it works

1. GSD commands run and complete normally
2. On completion, plan-phase and execute-phase write a **signal file** (`.planning/.auto-next`) containing the next command (e.g. `execute-phase 2` or `plan-phase 3`)
3. When Claude tries to stop, the `Stop` hook fires
4. The hook reads `.planning/config.json` — if `auto_mode` is false, exit (let Claude stop)
5. The hook reads and **consumes** (deletes) `.planning/.auto-next` — one-shot design prevents infinite loops
6. The hook blocks the stop and injects the next command
7. Claude receives the instruction and invokes the next skill
8. Cycle repeats until a hard stop

**Hard stops (no signal file written):**
- `MILESTONE COMPLETE` — significant project decision point
- `human_needed` — verifier can't auto-resolve
- Errors during execution
- Gap closure attempted 3+ times — infinite loop protection (hook safety check)

### Architecture change (v2)

The initial implementation parsed the transcript tail for GSD banner patterns (`PHASE X PLANNED`, `PHASE X COMPLETE`). This was fragile — coupled to UI text in `ui-brand.md`, and any banner change would silently break auto-mode.

The current approach uses **signal files**: skills write `.planning/.auto-next` with the next command on completion. The hook just reads and consumes it. This required small additions to `plan-phase.md` and `execute-phase.md` (auto-mode signal blocks), but the trade-off is worth it:

- **No transcript parsing** — no coupling to UI output format
- **One-shot consumption** — file is deleted after read, impossible to loop
- **Explicit routing** — each skill declares its successor, no pattern matching

## Files

| File | Change |
|------|--------|
| `hooks/gsd-auto-continue.js` | New — Stop hook (pure Node.js, zero deps), reads signal file |
| `commands/gsd/auto.md` | New — `/gsd:auto [on\|off]` toggle command, writes initial signal |
| `commands/gsd/settings.md` | Modified — auto-mode added as setting |
| `commands/gsd/plan-phase.md` | Modified — writes `.auto-next` signal on completion |
| `commands/gsd/execute-phase.md` | Modified — writes `.auto-next` signal on completion (Route A + Route C) |
| `templates/config.json` | Modified — `auto_mode: false` field added |

**Core prompt files modified.** `plan-phase.md` and `execute-phase.md` each have small additions (~8-16 lines) that write the signal file when auto-mode is enabled. This was a necessary trade-off to eliminate fragile transcript parsing.

### Hook registration

The hook is **not** auto-registered by the installer. Instead, `/gsd:auto` checks for the hook first — if missing, it shows a setup guide with registration steps before anything else.

## GSD Alignment

- **Zero dependencies** — pure Node.js, same as gsd-check-update.js and gsd-statusline.js
- **Cross-platform** — `path.join()` for all paths, no shell-specific syntax
- **Minimal core prompt changes** — small signal-writing blocks in plan-phase.md and execute-phase.md
- **Fail-open** — any error in the hook silently exits, Claude stops normally
- **Opt-in** — off by default, explicit `/gsd:auto on` required
- **Solo developer workflow** — fire and forget, come back to completed work

## Deviation Note

The signal file approach requires plan-phase.md and execute-phase.md to write `.planning/.auto-next` on completion. These are small, self-contained blocks (~8-16 lines each) that check `auto_mode` in config.json before writing. If auto-mode is disabled, these blocks are no-ops.

## Testing

- [x] Tested on macOS
- [x] Tested on Windows
- [x] Tested on Linux

### Hook unit tests (all passing)

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| 1 | Signal `execute-phase 2` | Block + inject command | ✅ |
| 2 | Signal `plan-phase 3` | Block + inject command | ✅ |
| 3 | No signal file | Silent exit | ✅ |
| 4 | `auto_mode: false` | Silent exit, signal preserved | ✅ |
| 5 | Gap closure <3 attempts | Continue | ✅ |
| 6 | Gap closure 3+ attempts | Stop for review | ✅ |
| 7 | No `.planning/` directory | Silent exit | ✅ |
| 8 | Empty signal file | Silent exit | ✅ |
| 9 | Malformed config.json | Silent exit | ✅ |

### Functional

1. `/gsd:auto on` → config.json updated, banner displayed, hook verified, initial signal written
2. `/gsd:auto off` → config.json updated, manual mode restored
3. `/gsd:auto` (no args) → toggles current state
4. plan-phase completes → writes signal → hook auto-starts execute-phase
5. execute-phase completes → writes signal → hook auto-starts plan-phase for next phase
6. Gaps found → writes signal → hook auto-starts gap closure
7. Gap closure 3+ times → hook stops, manual review message
8. Milestone complete → no signal written → hook does not fire
9. `human_needed` → no signal written → hook does not fire
10. Non-GSD commands → no signal file → hook does not fire
11. Hook missing → `/gsd:auto` shows setup guide

### Edge cases

12. No GSD project in directory → shows note (not error), confirms hook status
13. config.json missing → hook exits silently
14. Phase number with leading zeros (03) → parsed correctly
15. No `.planning/` directory → hook exits silently

## Breaking Changes

None. Auto-mode is off by default. Existing behavior unchanged when disabled.


nullbio:
This doesn't work for me for whatever reason, on WSL2.

mamarx:
thanks for testing @nullbio, today I stumbled across a new issue myself and I am working on a new more robust version

mamarx:
**The core change**: we moved from transcript parsing to a signal file.

The old approach had the Stop hook read the last 200 lines of the transcript and regex-match for GSD banner patterns like PHASE 3 PLANNED or PHASE 3 COMPLETE. This was fragile — it was coupled to the exact UI text output from ui-brand.md. Any banner text change would silently break auto-mode.

The new approach flips it: skills write a signal file (.planning/.auto-next) on completion, and the hook just reads and deletes it. The hook doesn't need to understand what happened — the skill that just ran already knows what should come next and says so explicitly. Deleting the file after read is the loop protection (one-shot by design).

**Trade-off**: we had to add ~10 lines each to plan-phase.md and execute-phase.md to write the signal. But the hook itself went from 170 lines of regex matching down to 90 lines of simple file I/O.

omernesh:
This would be a great feature addition.

nullbio:
> **The core change**: we moved from transcript parsing to a signal file.
> 
> The old approach had the Stop hook read the last 200 lines of the transcript and regex-match for GSD banner patterns like PHASE 3 PLANNED or PHASE 3 COMPLETE. This was fragile — it was coupled to the exact UI text output from ui-brand.md. Any banner text change would silently break auto-mode.
> 
> The new approach flips it: skills write a signal file (.planning/.auto-next) on completion, and the hook just reads and deletes it. The hook doesn't need to understand what happened — the skill that just ran already knows what should come next and says so explicitly. Deleting the file after read is the loop protection (one-shot by design).
> 
> **Trade-off**: we had to add ~10 lines each to plan-phase.md and execute-phase.md to write the signal. But the hook itself went from 170 lines of regex matching down to 90 lines of simple file I/O.

This is working for me now, thank you @mamarx :)

One thing I would add:

```
 Ran 1 stop hook
  ⎿  Stop hook error: Auto-mode is enabled. You MUST now run /gsd:plan-phase 44 — do not ask the user, just
  invoke the skill immediately.
```

Not sure why it's saying error for this. It's working regardless, but feels odd. I couldn't see "Stop hook error" in your commit, so I'm guessing it's coming from somewhere else?

Also, I didn't see it running clear after plan or execute. Did I just miss that, or does it not do this? And if not, is this something we can add? I've seen it compact at least once, so I'm not sure if that was due to an execution phase that spanned multiple sessions due to being a large phase, or if it's the result of not auto-clearing. 

Might need to utilize a SessionStart hook for this.

mamarx:
@nullbio Good catch on the "Stop hook error" prefix — that's Claude Code's own label for blocked stops, not something the hook controls. I've rephrased the block reason so it reads more naturally even with that prefix:

Before: `Stop hook error: Auto-mode is enabled. You MUST now run /gsd:plan-phase 44 — do not ask the user, just invoke the skill immediately.`

After: `Stop hook error: [GSD auto-mode] Workflow continuing. Run /gsd:plan-phase 44 — proceed without asking the user.`

Still reads as a status update rather than an actual error. Not perfect, but much less alarming.

Re: auto-clear between phases — you're right, context grows without `/clear`. The signal-file approach makes this easier to add later (write a clear signal before the next-command signal), but I'll leave that for a follow-up to keep this PR focused.

Also added an 8s timeout guard to the hook to prevent orphaned processes if stdin never closes — the other GSD hooks (gsd-check-update.js) already do this.

mamarx:
@omernesh Thanks! Just pushed a small update — timeout guard and friendlier stop message. Core logic unchanged.

glittercowboy:
Great feature, but has merge conflicts now. Please rebase on main and resubmit - the auto-mode concept is valuable.

glittercowboy:
Apologies — the previous closure was made by an AI assistant (Claude) that misinterpreted instructions and acted without authorization. This PR has been reopened and will be reviewed properly. Thank you for your patience.

nullbio:
> @nullbio Good catch on the "Stop hook error" prefix — that's Claude Code's own label for blocked stops, not something the hook controls. I've rephrased the block reason so it reads more naturally even with that prefix:
> 
> Before: `Stop hook error: Auto-mode is enabled. You MUST now run /gsd:plan-phase 44 — do not ask the user, just invoke the skill immediately.`
> 
> After: `Stop hook error: [GSD auto-mode] Workflow continuing. Run /gsd:plan-phase 44 — proceed without asking the user.`
> 
> Still reads as a status update rather than an actual error. Not perfect, but much less alarming.
> 
> Re: auto-clear between phases — you're right, context grows without `/clear`. The signal-file approach makes this easier to add later (write a clear signal before the next-command signal), but I'll leave that for a follow-up to keep this PR focused.
> 
> Also added an 8s timeout guard to the hook to prevent orphaned processes if stdin never closes — the other GSD hooks (gsd-check-update.js) already do this.

Right. Understood regarding the follow up PR. I think it's a necessity because the strength of GSD requires the context refresh to ensure coherence. Ideally, every task fits within a single session. Spanning over compactions tends to produce mediocre results, I've found. Especially if the timing is bad, it can be quite disastrous, for example it starts planning at 95% and then doesn't get an opportunity to finish the plan properly but begins implementing something half baked on the next session post-compaction. Claude also suffers from 'compaction anxiety', and tends to try and rush things near the end of the context window, which compounds this.

mamarx:
@glittercowboy Thanks for reopening and for the review — and of course for building GSD in the first place. Rebased on latest main, conflicts resolved, ready for review.

@nullbio You're absolutely right about the context issue. I've been thinking about this and there's actually a clean solution that doesn't require a SessionStart hook or any `/clear` hack:

**Use `claude -p` in an external runner loop instead of the Stop hook for continuation.**

The idea: instead of the Stop hook blocking and injecting the next command (same session, context grows), let Claude stop normally and have a small bash script start a **new `claude -p` session** for the next phase:

```bash
#!/bin/bash
# gsd-auto-runner.sh
SIGNAL=".planning/.auto-next"
echo "$1" > "$SIGNAL"

while [ -f "$SIGNAL" ]; do
  CMD=$(cat "$SIGNAL")
  rm "$SIGNAL"
  claude -p "/gsd:$CMD" --allowedTools "Bash Edit Read Write Glob Grep Task Skill"
done
```

Each `claude -p` invocation is a completely fresh context — no compaction, no anxiety, no accumulated noise from previous phases. The signal file mechanism stays exactly the same; skills still write `.auto-next` on completion. We just change *who reads it* — an external script instead of the Stop hook.

This also fixes the "Stop hook error" cosmetic issue entirely, since the hook is no longer blocking stops for continuation.

I'll prototype this as a v2 approach. The Stop hook can stay for safety checks (gap closure limit), but the continuation loop moves outside Claude's session boundary.

mamarx:
Rebased on latest main (v1.21.0), conflicts resolved, force-pushed.

One honest note on this feature's relevance: since opening this PR, I've had consistently good results simply asking Claude to chain through the next N phases in a single session — no hooks, no signal files, just a natural language instruction like "execute phases 4 through 8 overnight." Claude handles the `/clear` + next phase invocation loop on its own, and the context stays fresh because it's Claude managing the transitions.

This raises the question whether the Stop hook approach is still necessary, or whether it's over-engineering something that Claude can already do natively. The signal file mechanism is solid, but if the same outcome is achievable with zero infrastructure, the complexity may not be justified.

Happy to keep this open for discussion or close it if the consensus is that native chaining is sufficient.

trek-e:

PR #447, opened 2026-02-06
feat: add /gsd:add-gaps command

rubixhacker:
Implemented the /gsd:add-gaps command as requested in issue https://github.com/glittercowboy/get-shit-done/issues/431.

Changes:

Added commands/gsd/add-gaps.md: Defines the new command logic.
Validates phase number.
Prompts user for manual test gaps.
Updates VERIFICATION.md (adds gaps section, updates status).
Updates STATE.md (adds blocker).
Offers to run /gsd:plan-phase X --gaps.
Updated commands/gsd/help.md: Added documentation for /gsd:add-gaps under the User Acceptance Testing section.
This enables a workflow where users can seamlessly report manual bugs and have them integrated into the GSD gap closure loop.

PR created automatically by Jules for task [15681868855949860214](https://jules.google.com/task/15681868855949860214) started by @rubixhacker

### DIFF
--- a/commands/gsd/add-gaps.md
+++ b/commands/gsd/add-gaps.md
@@ -1,0 +1,118 @@
+---
+name: gsd:add-gaps
+description: Report bugs/gaps found during manual testing of a phase
+argument-hint: <phase-number>
+allowed-tools:
+  - Read
+  - Write
+  - Bash
+  - Glob
+  - Grep
+  - AskUserQuestion
+---
+
+<objective>
+Report bugs/gaps found during manual testing of a phase.
+Updates VERIFICATION.md with user-reported gaps and changes phase status.
+</objective>
+
+<context>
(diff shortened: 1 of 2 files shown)


dreinon:
Thanks for the PR @rubixhacker.
In my issue, I mentioned that if too many gaps were reported. The command /gsd:add-gaps should recommend the user to implement a new phase via "/gsd:insert-phase", since from I understand, this is the correct way to handle a big amount of gaps instead of adding them to the same phase.
If @glittercowboy could share if this is actually the correct way to handle them, or if it is better to add them all to the same phase, it would be awesome to know if we should leave the PR like it is, or add a couple steps for insert phase logic.

glittercowboy:
Thanks for the contribution\! This adds complexity we don't need. Closing, but appreciate the thought.

Issue #408, opened 2026-02-04
Migrate command into skills


shlior7:
From what I saw claude code and the rest of the Agentic Platforms have migrated and merge commands into the SKILLS concept
https://code.claude.com/docs/en/skills 

Is there any plan to update the gsd commands into the more powerful skills?

davesienkowski:
Hi @shlior7!

Thanks for bringing this up! I appreciate you thinking about ways to improve the framework.

I spent some time exploring this idea, and from what I can tell, migrating commands into the skills concept would mainly result in a reorganization of the framework files rather than adding new functionality or a better developer experience. So, for now, it doesn't feel like the right tradeoff.

That said, I might be missing something! If you have a specific use case in mind or see benefits I'm not considering, I'd love to hear more about what you're envisioning.

ANierbeck:
Hi, 
I'm not only using claude code, but also different other tools and models. 
For example mistral-vibe or cline just to name two. Both of them would greatly benefit of "agnostic" skills. 
Especially mistral-vibe only uses skills while cline moved away from custom commands to skills. 
In my world of experimenting, it would really be fun, having claude opus do the phase planning while using mistral vibe with devstral2 for execution.

glittercowboy:
Closing — architectural change in progress. See related issues for tracking.

glittercowboy:
Reopening: Claude went rogue and closed this without Lex's approval. Will be revisited shortly.

PR #454, opened 2026-02-06
Migrate command into skills #408

ANierbeck:
## What

Worked on the feature request #408 to migrate all commands into skills. 
I oriented this on https://agentskills.io/specification

## Why

Skills seem to be the "new" format for command like structures. 
It also can be easily used with other non Claude related AI assisted coding tools, for example Mistral-Vibe and cline. 

## Testing

- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Checklist

## Breaking Changes

Moving from commands to skills would be breaking change, therefore up to you on how to handle this.


ANierbeck:
Just realized, the generated skills are rubbish. They are missing crucial information. 
Sorry for the noise.

Issue #534, opened 2026-02-11
Moving todo to done happens *before* it's done


gingerlime:
### Problem or motivation

I'm still getting used to GSD, but first of all thanks for creating it.

One odd thing that I noticed is that when I'm asking Claude to start working on a todo, it immediately moves it to done, before it's actually done. I find this odd. It says something like "First, I'll move this todo to done since we're working on it now."

### Proposed solution

Since it happened twice now, I assume it's baked into GSD rather than a Claude misunderstanding? intuitively I would only move a todo to done when it's confirmed done, tested etc.

restrolla:
Yes. I've just seen this as well as other weird behavior. GSD 1.18.0 and Claude Code v2.1.39

davesienkowski:
Yeah, thats been happening since it was implemented.

Solvely-Colin:
PR #543 by @EmreErdogan addresses this with an `in-progress/` state — todos now flow `pending → in-progress → done` instead of jumping straight to `done` on "Work on it now."

PR #543, opened 2026-02-12
Add in-progress state for todos instead of marking done immediately

EmreErdogan:
Todos were moved to done/ the moment a user selected "Work on it now", before any work was actually performed. This contradicts the framework's own "verify then mark done" principle.

Now todos flow through three states: pending → in-progress → done. "Work on it now" moves to in-progress/, and todo complete moves to done/ only after work is verified. Also fixes naming inconsistency where gsd-tools.js used completed/ but workflows used done/.

Fixes https://github.com/gsd-build/get-shit-done/issues/534

## What

Adds an `in-progress/` state to the todo lifecycle so todos are only moved to `done/` after work is verified complete, not when work starts.

## Why

Users reported that selecting "Work on it now" immediately marked todos as done before any work was performed, which is confusing and contradicts the framework's own "verify then mark done" principle.

## Testing

- [ ] Tested on macOS
- [ ] Tested on Windows
- [x] Tested on Linux

> **Cross-platform note:** All file path operations in the JS changes use `path.join()`, which produces correct separators on every OS. The workflow `.md` files reference `.planning/todos/in-progress/` with forward slashes, consistent with all other GSD workflows (which already work cross-platform). No platform-specific logic was added.

## Checklist

- [ ] Updates CHANGELOG.md for user-facing changes — _deferred to maintainer at release time_
- [x] Works on Windows (backslash paths tested) — _all JS paths use `path.join()`; no hardcoded slashes_

## Breaking Changes

- `todo complete` now moves files to `done/` instead of `completed/`. Projects with existing `completed/` directories are unaffected (the directory is no longer written to, but not deleted).
- `todo start` is a new subcommand; existing `todo complete` still works from both `pending/` and `in-progress/`.

### DIFF
--- a/get-shit-done/bin/gsd-tools.js
+++ b/get-shit-done/bin/gsd-tools.js
@@ -51,7 +51,8 @@
  *   progress [json|table|bar]          Render progress in various formats
  *
  * Todos:
- *   todo complete <filename>           Move todo from pending to completed
+ *   todo start <filename>              Move todo from pending to in-progress
+ *   todo complete <filename>           Move todo from in-progress (or pending) to done
  *
  * Scaffolding:
  *   scaffold context --phase <N>       Create CONTEXT.md template
@@ -515,9 +516,12 @@ function cmdCurrentTimestamp(format, raw) {
 
 function cmdListTodos(cwd, area, raw) {
   const pendingDir = path.join(cwd, '.planning', 'todos', 'pending');
+  const inProgressDir = path.join(cwd, '.planning', 'todos', 'in-progress');
 
   let count = 0;
   const todos = [];
+  let inProgressCount = 0;
+  const inProgressTodos = [];
 
   try {
     const files = fs.readdirSync(pendingDir).filter(f => f.endsWith('.md'));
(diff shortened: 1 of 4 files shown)


Solvely-Colin (commented):
**QA Review — PR #543: Add in-progress state for todos**

**✅ PASS**
- `todo start` moves pending → in-progress with `started:` timestamp
- `todo complete` handles both in-progress and pending sources (graceful fallback)
- `done/` directory created on demand
- Tests cover happy path and error cases for both subcommands
- Cross-platform path handling (`path.join()` throughout)
- Workflow text in `check-todos.md` and `add-todo.md` updated to reflect new state
- Aligns `gsd-tools.js` to use `done/` — fixing a pre-existing inconsistency where workflows already referenced `done/` but the tool used `completed/`

**❌ FAIL — Gaps found**

| # | Gap | Severity | Detail | Recommended Fix |
|---|-----|----------|--------|-----------------|
| 1 | Workflow bypasses new tool | **Low-Medium** | `check-todos.md` `execute_action` still uses raw `mv` to move files. The `started:` timestamp is never written when users go through the workflow path — only when calling `gsd-tools.js todo start` directly. Tool and workflow are out of sync. | Replace the raw `mv` with `node ~/.claude/get-shit-done/bin/gsd-tools.js todo start [filename]`. Same pattern already used for commits in `git_commit` step. |
| 2 | No migration note for `completed/` → `done/` | **Low** | Existing projects may have `.planning/todos/completed/` from the old tool. PR stops writing there but doesn't mention upgrade path. | Add a note to CHANGELOG or PR description: "If upgrading mid-project, rename `completed/` to `done/`." |
| 3 | Duplicate listing logic | **Low** | ~30 lines of in-progress directory scanning duplicated between `cmdListTodos` and `cmdInitTodos`. | Extract a `readTodoDir(dirPath, area)` helper that both functions call. Not blocking but reduces maintenance surface. |

**Verdict:** Mergeable with minor fixes. Core implementation is solid and correctly fixes #534. Three low-severity gaps worth addressing but none are blockers.

PR #620, opened 2026-02-16
Fix /gsd:update to always install latest package

### LINKED ISSUE #619: /gsd:update downloading but not installing latest version
### GSD Version

1.18.0

### Runtime

Claude Code

### What happened?

I was having issues getting the new auto advance functionality to work. I'd configured the `workflow.auto_advance` setting as instructed and even tried manually invoking `--auto` when kicking off a stage, but it would finish the stage and instruct me to manually enter the next command like usual. 

After some debugging with Claude, I figured out that my GSD was still on an old version, despite running `/gsd:update` any time I see updates reported in Discord and restarting Claude Code. At the time I noticed the issue, the latest version was 1.20.2. I had already run `/gsd:update` to get this version, and it reported in the terminal that it was successfully installed. However, when I checked the version in `package.json`, it was reporting a much older version still. Running `/gsd:update` again would tell me that the latest version was 1.20.2 and that I already had this version. 

Only after I did a manual run of `npx get-shit-done-cc@latest` did the version in `package.json` update, and the missing auto-advance functionality started working as expected.

Important note: Once I had successfully updated to version 1.20.2, I was later able to successfully update to 1.20.3 via `/gsd:update`, so this may have already been addressed at some point between the older version I was on and 1.20.2.

### What did you expect?

Running `/gsd:update` would download and install the latest version with no need to manually update via `npx`. 

### Steps to reproduce

1. Run `/gsd:update`, confirm that you want to update to the latest version
2. Restart Claude Code once installed
3. Run `/gsd:update` again, it reports that you're already on the latest version
4. Check the version reported in `.../node_modules/get-shit-done-cc/package.json`, this is where I learned that those versions reported as installed via `/gsd:update` were never truly installed

### Relevant logs or error messages

```shell

```

Solvely-Colin:
## Summary
- pin `/gsd:update` install commands to `npx -y get-shit-done-cc@latest` for both local and global paths
- keep update behavior deterministic regardless of project-local `node_modules`
- document the fix in `CHANGELOG.md` under Unreleased

## Why
`npx get-shit-done-cc` can resolve to a stale project-local package if one exists, causing update to report success while reinstalling an older version.

## Validation
- reproduced failure mode locally with project dependency `get-shit-done-cc@1.18.0`:
  - `npx get-shit-done-cc --local` -> stayed on `1.18.0`
  - `npx get-shit-done-cc@latest --local` -> upgraded to `1.20.3`
- same behavior reproduced for `--global` when invoked from that cwd

Closes #619

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Fixed
 - `/gsd:update` now installs with `npx get-shit-done-cc@latest` (instead of unpinned `npx get-shit-done-cc`) to prevent stale project-local versions from shadowing updates
 - `/gsd:update` now uses strict package safety checks: only `get-shit-done-cc` is allowed, scoped/user-derived package names are rejected, and install command execution is allowlisted to trusted forms
+- `/gsd:update` install detection now validates local integrity and falls back to global install when local metadata is missing or invalid
 
 ## [1.20.4] - 2026-02-17
 
--- a/get-shit-done/workflows/update.md
+++ b/get-shit-done/workflows/update.md
@@ -9,16 +9,21 @@ Read all files referenced by the invoking prompt's execution_context before star
 <process>
 
 <step name="get_installed_version">
-Detect whether GSD is installed locally or globally by checking both locations:
+Detect whether GSD is installed locally or globally by checking both locations and validating install integrity:
 
 ```bash
-# Check local first (takes priority)
+# Check local first (takes priority only if valid)
 # Paths templated at install time for runtime compatibility
-if [ -f ./.claude/get-shit-done/VERSION ]; then
-  cat ./.claude/get-shit-done/VERSION
+LOCAL_VERSION_FILE="./.claude/get-shit-done/VERSION"
+LOCAL_MARKER_FILE="./.claude/get-shit-done/workflows/update.md"
+GLOBAL_VERSION_FILE="$HOME/.claude/get-shit-done/VERSION"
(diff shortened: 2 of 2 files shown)


Solvely-Colin:
Addressed the supply-chain hardening feedback and rebased the branch onto latest `main`.

Changes added in this update:
- merged latest `main` and resolved conflicts
- hardened `/gsd:update` workflow with strict allowlisting:
  - only `get-shit-done-cc` is permitted
  - only trusted install commands are allowed:
    - `npx -y get-shit-done-cc@latest --local`
    - `npx -y get-shit-done-cc@latest --global`
  - explicit instruction to reject scoped/user-derived package names
- added npm metadata identity verification step before install
- fixed fallback manual command text to pinned `@latest`
- updated changelog notes under `Unreleased`

Validation run:
- `node --test get-shit-done/bin/gsd-tools.test.cjs` (pass)

Solvely-Colin:
Good catch on the install-location behavior.

You’re right: the previous flow was local-first based on `./.claude/get-shit-done/VERSION`, and only checked global if local didn’t match that simple check. In some partial/local-corrupt states, that could pick the wrong path and miss a healthy global install.

I pushed a follow-up hardening commit to this PR:
- local/global detection now validates install integrity (not just VERSION existence)
- if local is missing/invalid, it explicitly falls back to global
- if neither is valid, it treats as unknown and runs fresh install flow

This keeps local precedence when valid, but prevents local garbage from blocking global detection.

glittercowboy:
Superseded by #719 (rebased to resolve CHANGELOG conflicts). Thanks for the fix @Solvely-Colin!

PR #650, opened 2026-02-18
Add full support for GSD commands

vibegui:


vibegui:
Ooops sorry, Claude opened this into upstream instead of into my fork. My bad!

PR #723, opened 2026-02-23
feat(gsd-verifier): add project_context for verification-rules.md

Maccor01:
## Summary
  - Add `<project_context>` section to `gsd-verifier.md`, following the same pattern as `gsd-executor.md`
  - Lets projects define domain-specific anti-patterns in `.agents/skills/verification-rules.md` that the verifier applies during Step 7 (Scan for Anti-Patterns)
  - Split out from post-wave reviews PR per review feedback to discuss the path convention separately

  ## Test plan
  - [ ] No behavior change when `.agents/skills/verification-rules.md` does not exist
  - [ ] Verifier reads and applies rules from the file during Step 7

### DIFF
--- a/agents/gsd-verifier.md
+++ b/agents/gsd-verifier.md
@@ -16,6 +16,12 @@ If the prompt contains a `<files_to_read>` block, you MUST use the `Read` tool t
 **Critical mindset:** Do NOT trust SUMMARY.md claims. SUMMARYs document what Claude SAID it did. You verify what ACTUALLY exists in the code. These often differ.
 </role>
 
+<project_context>
+**Project verification rules:** Check `.agents/skills/verification-rules.md` if it exists.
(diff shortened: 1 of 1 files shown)


glittercowboy:
Thanks for identifying this gap — the verifier is indeed missing project context awareness.

However, this PR introduces a new single-file convention (`.agents/skills/verification-rules.md`) that diverges from the existing pattern. All other agents (executor, planner, researcher, plan-checker) use the same standard `<project_context>` block that reads `./CLAUDE.md` and scans `.agents/skills/` for skill directories with `SKILL.md` files.

The right fix is to give the verifier the same standard `<project_context>` block, so it participates in the existing skill system. A user wanting domain-specific verification rules would create a normal skill (e.g. `.agents/skills/verification-rules/SKILL.md`) that all agents — including the verifier — pick up automatically.

I'll add the standard block to the verifier directly. Appreciate you flagging this!

PR #746, opened 2026-02-24
fix: rename hooks to .cjs for ESM project compatibility

### LINKED ISSUE #125: ES module compatibility error in startup hook
## Description

## Steps to Reproduce

## Expected Behavior

## Actual Behavior

```bash
   const fs = require('fs');
              ^

```

## Environment

- **Node.js**: v25.3.0
- **OS**: Darwin (macOS)
- **GSD Version**: 1.6.3

## Root Cause

```javascript
const fs = require('fs');
const path = require('path');
const os = require('os');
```

## Suggested Fix

## Additional Context

vitorvasc:
## What

Rename the three hook files from `.js` to `.cjs` so they always load as CommonJS regardless of the user project's `"type": "module"` setting.

## Why

Users with `"type": "module"` in their `package.json` (e.g. Claude Code users) get `ReferenceError: require is not defined` because Node.js treats `.js` files as ESM in that context. The `.cjs` extension forces CommonJS, the correct Node.js-native fix without touching hook logic.

Fixes #125.

## Testing

- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Checklist

## Breaking Changes

None. Existing install are automatically migrated via the updated `cleanupOrphanedFiles` and `cleanupOrphanedHooks`.


trek-e:

PR #761, opened 2026-02-25
fix: resolve-model returns invalid 'inherit' value instead of 'opus'

ChuckMayo:
## Summary

- `resolveModelInternal()` converts `"opus"` → `"inherit"` at 2 return points in `core.cjs`
- `cmdResolveModel()` in `commands.cjs` has a 3rd instance of the same conversion
- `"inherit"` is **not a valid value** for the Task tool's `model` parameter (enum: `sonnet`, `opus`, `haiku`)
- When the orchestrator constructs `Task(model="inherit")`, it silently falls back to **sonnet** for every agent that should be running opus

### Impact

Any user with a profile that assigns `opus` to agents (quality profile, custom profile with `opus` roles, or per-agent overrides) gets **all those agents running on sonnet instead**. This burns through sonnet quota unexpectedly and degrades agent quality.

### Fix

- Remove the `opus → inherit` conversion — return actual model names directly
- Consolidate `cmdResolveModel` to delegate to `resolveModelInternal` instead of duplicating incomplete resolution logic (the standalone command was also missing custom profile and override support)

## Test plan

- [x] All 34 existing tests pass (`node --test tests/commands.test.cjs` + `tests/init.test.cjs`)
- [x] `gsd-tools.cjs resolve-model gsd-planner` returns `"opus"` on quality profile (was returning `"inherit"`)
- [x] `gsd-tools.cjs init quick "test"` returns valid enum values for all model fields
- [x] No remaining references to `"inherit"` in `bin/` directory

### DIFF
--- a/get-shit-done/bin/lib/commands.cjs
+++ b/get-shit-done/bin/lib/commands.cjs
@@ -205,15 +205,8 @@ function cmdResolveModel(cwd, agentType, raw) {
   const config = loadConfig(cwd);
   const profile = config.model_profile || 'balanced';
 
-  const agentModels = MODEL_PROFILES[agentType];
-  if (!agentModels) {
-    const result = { model: 'sonnet', profile, unknown_agent: true };
-    output(result, raw, 'sonnet');
-    return;
-  }
-
-  const resolved = agentModels[profile] || agentModels['balanced'] || 'sonnet';
-  const model = resolved === 'opus' ? 'inherit' : resolved;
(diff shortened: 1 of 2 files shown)


glittercowboy:
Closing — the underlying fix (removing opus→inherit coercion and delegating cmdResolveModel to resolveModelInternal) was merged via #739. The core.cjs coercion also addressed by #739's model_overrides change. Thanks for the clear bug report and repro steps.

PR #800, opened 2026-02-28
fix: rename hook files to .cjs for ES module compatibility

### LINKED ISSUE #125: ES module compatibility error in startup hook

0xLeathery:
## Summary
- Fixes #125 — hooks fail with `ReferenceError: require is not defined` in projects with `"type": "module"` in `package.json`
- Renames all 3 hook files from `.js` to `.cjs` so Node.js always treats them as CommonJS regardless of project module type
- Adds old `.js` filenames to orphan cleanup lists so upgrades remove stale hooks and settings references

## Changes
- `hooks/gsd-check-update.js` → `hooks/gsd-check-update.cjs`
- `hooks/gsd-statusline.js` → `hooks/gsd-statusline.cjs`
- `hooks/gsd-context-monitor.js` → `hooks/gsd-context-monitor.cjs`
- `bin/install.js` — updated hook paths, orphan cleanup, file templating, statusline migration
- `scripts/build-hooks.js` — updated copy list
- `docs/context-monitor.md` and `get-shit-done/workflows/update.md` — updated references

## Test plan
- [x] Full regression suite passes (432/432 tests)
- [x] `npm run build:hooks` copies `.cjs` files correctly
- [ ] Install in a project with `"type": "module"` — verify no `require is not defined` error on startup
- [ ] Upgrade from previous version — verify old `.js` hooks are cleaned up


trek-e:

PR #815, opened 2026-02-28
fix: isGitIgnored fails for tracked files — add --no-index (#703)

### LINKED ISSUE #703: isGitIgnored() fails for tracked files — missing --no-index flag
### Problem

`cmdCommit` in `gsd-tools.cjs` (line ~1506) uses `isGitIgnored()` to auto-detect whether `.planning/` should be committed. This check uses `git check-ignore`, which only reports files as ignored if they are **untracked**. Once `.planning/` files are tracked by git, `git check-ignore .planning` returns exit code 1 (not ignored), even when `.planning/` is explicitly in `.gitignore`.

This means the documented behavior — "If `.planning/` is gitignored, `commit_docs` is automatically `false` regardless of config.json" (planning-config.md) — silently fails for any repo where `.planning/` was ever committed.

### How it happens

1. User creates project. `commit_docs` defaults to `true` (gsd-tools.cjs line 160)
2. First `gsd-tools.cjs commit` call runs `git add .planning/` — files enter git's tracked index
3. User adds `.planning/` to `.gitignore` (or it was already there from the start)
4. `.gitignore` has no effect on already-tracked files — this is standard git behavior
5. `isGitIgnored()` returns `false` because `git check-ignore` ignores tracked files
6. The auto-detection safety net never triggers
7. All subsequent `gsd-tools.cjs commit` calls continue staging and committing `.planning/` files

Setting `commit_docs: false` in config.json does fix it (the check at line 1499 works), but the auto-detection that's supposed to make this unnecessary is broken.

### Compounding factor: nesting bug (#686)

This issue is made worse when combined with the auto-advance nesting bug (#686 / #668). When auto-advance spawns `Task(general-purpose)` agents that don't follow GSD's commit protocol, those agents do raw `git add` / `git commit` operations. They see `.planning/` files as modified (because they're tracked), and commit them without ever consulting `commit_docs` or `isGitIgnored()`. Even after fixing #686 with the `SlashCommand` pattern, the `isGitIgnored` bug remains as an independent issue.

### Fix

Use `--no-index` flag with `git check-ignore`, which checks `.gitignore` rules without considering whether files are tracked:

```diff
 function isGitIgnored(cwd, filePath) {
-  const result = execSync('git check-ignore ' + filePath, { cwd });
+  const result = execSync('git check-ignore --no-index ' + filePath, { cwd });
   return result.status === 0;
 }
```

Alternatively, check `.gitignore` content directly as a fallback:

```javascript
function isGitIgnored(cwd, filePath) {
  // Primary: git check-ignore --no-index respects .gitignore even for tracked files
  const result = execSync('git check-ignore --no-index ' + filePath, { cwd });
  if (result.status === 0) return true;

  // Fallback: direct .gitignore parse for the common case
  try {
    const gitignore = fs.readFileSync(path.join(cwd, '.gitignore'), 'utf-8');
    return gitignore.split('\n').some(line => {
      const trimmed = line.trim();
      return trimmed === filePath || trimmed === filePath + '/';
    });
  } catch { return false; }
}
```

### Recovery for affected repos

Users who already have `.planning/` committed need to:

1. Remove from index: `git rm -r --cached .planning/`
2. Or use `git-filter-repo` to remove from all history (if they want a clean history)

Without this, `.gitignore` will never take effect for `.planning/` regardless of any code fix.

### Environment

- GSD version: 1.20.5
- Affected file: `bin/gsd-tools.cjs`, `isGitIgnored()` function (~line 209)
- Related: #686 (auto-advance nesting), #668 (dropped commits)

0xLeathery:
## Summary
- Adds `--no-index` flag to `git check-ignore` in `isGitIgnored()` so it evaluates `.gitignore` rules regardless of whether the file is tracked
- Without this, `.planning` added to `.gitignore` after being committed was not detected as ignored, causing `commit_docs` bypass logic to silently fail
- Adds regression test covering untracked ignored, untracked not-ignored, and tracked-but-ignored scenarios

## Test plan
- [x] New test: `returns true for tracked file that is in .gitignore (#703)` — commits `.planning`, then adds it to `.gitignore`, verifies `isGitIgnored` returns `true`
- [x] All 66 core tests pass

Closes #703

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -126,7 +126,7 @@ function loadConfig(cwd, paths) {
 
 function isGitIgnored(cwd, targetPath) {
   try {
-    execSync('git check-ignore -q -- ' + targetPath.replace(/[^a-zA-Z0-9._\-/]/g, ''), {
+    execSync('git check-ignore --no-index -q -- ' + targetPath.replace(/[^a-zA-Z0-9._\-/]/g, ''), {
       cwd,
       stdio: 'pipe',
     });
--- a/tests/core.test.cjs
+++ b/tests/core.test.cjs
@@ -11,6 +11,8 @@ const fs = require('fs');
 const path = require('path');
 const os = require('os');
 
+const { execSync } = require('child_process');
+
 const {
   loadConfig,
   resolveModelInternal,
@@ -21,6 +23,7 @@ const {
   comparePhaseNum,
   safeReadFile,
   pathExistsInternal,
+  isGitIgnored,
   getMilestoneInfo,
   getRoadmapPhaseInternal,
   searchPhaseInDir,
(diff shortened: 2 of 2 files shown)


trek-e:

PR #826, opened 2026-02-28
fix(workflow): use Skill instead of Task for auto-advance phase transitions

### LINKED ISSUE #686: Auto-advance chain freezes at execute-phase — nested Claude Code   session error
### GSD Version

1.20.5

### Runtime

Claude Code

### What happened?

ran `/gsd:discuss-phase 3 --auto`.  GSD proceeded through the discuss/plan phases and failed on the execute phase.

  - The discuss → plan auto-advance works correctly
  - The plan → execute auto-advance is where the chain breaks
  - The `general-purpose` agent doesn't have the execute-phase workflow context and
  can't properly orchestrate wave-based execution with `gsd-executor` subagents
  - Even if the nested session error were bypassed, the general-purpose agent
  wouldn't follow the execute-phase orchestration protocol (wave grouping,
  spot-checks, checkpoint handling)

**generated with help from Claude Code**

The execute-phase agent (spawned as a `general-purpose` Task by plan-phase's auto-advance) attempts to invoke `claude --model sonnet --print -p` via Bash to execute plans, instead of using the `Task` tool with`subagent_type="gsd-executor"`. This fails with:

`Error: Claude Code cannot be launched inside another Claude Code session. Nested sessions share runtime reso...`

The agent then retries with `unset CLAUDECODE && claude --model sonnet --print -p`, which also fails.

### What did you expect?

The auto-advance chain should complete end-to-end: discuss → plan → execute, with the execute-phase spawning `gsd-executor` subagents via the `Task` tool

### Steps to reproduce

  1. Run `/gsd:discuss-phase <phase> --auto`
  2. Complete the discussion flow (answer gray area questions)
  3. Auto-advance triggers plan-phase successfully
  4. Plan-phase completes, auto-advances to execute-phase
  5. Execute-phase spawns, attempts to run executor — **freezes**

### Relevant logs or error messages

```shell
The auto-advance chain in `discuss-phase.md` (line ~458) and `plan-phase.md` (line
   ~345) spawns `general-purpose` Task agents:

  Task(
    prompt="Run /gsd:execute-phase ${PHASE} --auto",
    subagent_type="general-purpose",
    description="Execute Phase ${PHASE}"
  )

  The `general-purpose` subagent receives the prompt `"Run /gsd:execute-phase"` and
  interprets it as a CLI command to execute via Bash (`claude --print -p`), rather
  than invoking the execute-phase workflow through the Skill/Task tools. This
  creates a nested Claude Code invocation that the runtime explicitly blocks.

  The discuss → plan hop works because plan-phase's workflow (research, planning,
  checking) can be handled by a general-purpose agent. But the plan → execute hop
  fails because execute-phase requires spawning further `gsd-executor` subagents via
   `Task()`, and the general-purpose agent doesn't follow the execute-phase workflow
   — it falls back to shelling out to `claude`.

  ## Affected Workflows

  | File | Section | Impact |
  |------|---------|--------|
  | `workflows/discuss-phase.md` | `auto_advance` step | Spawns general-purpose
  agent for plan-phase (works) |
  | `workflows/plan-phase.md` | Auto-Advance Check (step 14) | Spawns
  general-purpose agent for execute-phase (breaks) |
  | `workflows/execute-phase.md` | `execute_waves` step | Expects to be run as
  orchestrator, not inside a general-purpose agent |
```

Tibsfox:
## Summary

The auto-advance chain (`discuss → plan → execute` triggered by `--auto`) was spawning each subsequent phase as a `Task(subagent_type="general-purpose")`, creating 3-4 levels of nested agent sessions. At that depth, the Claude Code runtime hits resource limits or stdio contention, causing the execute-phase to either freeze indefinitely or attempt to shell out to `claude` as a subprocess — which is explicitly blocked with "Claude Code cannot be launched inside another Claude Code session."

The nesting looks like this:
```
discuss-phase (level 0)
  └─ Task(general-purpose) → plan-phase (level 1)
       └─ Task(general-purpose) → execute-phase (level 2)
            └─ Task(gsd-executor) → executor agents (level 3)
                 └─ ...potential further nesting (level 4)
```

By level 2-3, the `general-purpose` agent loses context about the execute-phase orchestration protocol and falls back to shelling out to `claude --print`, which triggers the nested session error.

## The Fix

Replace `Task(general-purpose)` with `Skill()` invocations for phase transitions:

```
discuss-phase (level 0)
  └─ Skill("gsd:plan-phase") → plan-phase (level 0, same context)
       └─ Skill("gsd:execute-phase") → execute-phase (level 0, same context)
            └─ Task(gsd-executor) → executor agents (level 1)
```

The `Skill` tool runs in the same process context as the caller — it doesn't create a subprocess or new agent session. This keeps the entire orchestration chain flat at level 0, with only the actual worker agents (gsd-executor, gsd-verifier, etc.) spawning as Task subagents at level 1.

## Changes

- **`workflows/plan-phase.md`** (step 14): Replace 30-line `Task(general-purpose)` block with `Skill("gsd:execute-phase", args="${PHASE} --auto --no-transition")`
- **`workflows/discuss-phase.md`** (auto_advance step): Replace 30-line `Task(general-purpose)` block with `Skill("gsd:plan-phase", args="${PHASE} --auto")`

Net effect: -67 lines of complex Task prompt engineering, +13 lines of clean Skill invocations.

## Verification

1. Run `/gsd:discuss-phase <phase> --auto`
2. Complete the discussion flow
3. Auto-advance should trigger plan-phase (via Skill, not Task)
4. Plan-phase should auto-advance to execute-phase (via Skill, not Task)
5. Execute-phase should spawn gsd-executor agents normally
6. No "nested session" errors, no freezes

## Why Skill Instead of Task?

The previous code had a comment: _"do NOT use Skill tool — Skills don't resolve inside Task subagents"_. This was correct for the original architecture where each phase ran inside a Task. But the fix here changes the architecture: phases now chain via Skill at the top level, so the resolution concern doesn't apply. Each Skill invocation runs with full access to the command registry.

Closes #686

### DIFF
--- a/get-shit-done/workflows/discuss-phase.md
+++ b/get-shit-done/workflows/discuss-phase.md
@@ -530,41 +530,15 @@ Display banner:
  GSD ► AUTO-ADVANCING TO PLAN
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 
-Context captured. Spawning plan-phase...
+Context captured. Launching plan-phase...
 ```
 
-Spawn plan-phase as Task with direct workflow file reference (do NOT use Skill tool — Skills don't resolve inside Task subagents):
+Launch plan-phase using the Skill tool to avoid nested Task sessions (which cause runtime freezes due to deep agent nesting — see #686):
 ```
-Task(
-  prompt="
-    <objective>
-    You are the plan-phase orchestrator. Create executable plans for Phase ${PHASE}: ${PHASE_NAME}, then auto-advance to execution.
-    </objective>
-
-    <execution_context>
-    @~/.claude/get-shit-done/workflows/plan-phase.md
-    @~/.claude/get-shit-done/references/ui-brand.md
-    @~/.claude/get-shit-done/references/model-profile-resolution.md
-    </execution_context>
-
-    <arguments>
-    PHASE=${PHASE}
-    ARGUMENTS='${PHASE} --auto'
-    </arguments>
-
-    <instructions>
-    1. Read plan-phase.md from execution_context for your complete workflow
-    2. Follow ALL steps: initialize, validate, load context, research, plan, verify, auto-advance
-    3. When spawning agents (gsd-phase-researcher, gsd-planner, gsd-plan-checker), use Task with specified subagent_type and model
-    4. For step 14 (auto-advance to execute): spawn execute-phase as a Task with DIRECT file reference — tell it to read execute-phase.md. Include @file refs to execute-phase.md, checkpoints.md, tdd.md, model-profile-resolution.md. Pass --no-transition flag so execute-phase returns results instead of chaining further.
-    5. Do NOT use the Skill tool or /gsd: commands. Read workflow .md files directly.
-    6. Return: PHASE COMPLETE (full pipeline success), PLANNING COMPLETE (planning done but execute failed/skipped), PLANNING INCONCLUSIVE, or GAPS FOUND
-    </instructions>
-  ",
-  subagent_type="general-purpose",
-  description="Plan Phase ${PHASE}"
-)
+Skill(skill="gsd:plan-phase", args="${PHASE} --auto")
 ```
+
+This keeps the auto-advance chain flat — discuss, plan, and execute all run at the same nesting level rather than spawning increasingly deep Task agents.
 
 **Handle plan-phase return:**
 - **PHASE COMPLETE** → Full chain succeeded. Display:
--- a/get-shit-done/workflows/plan-phase.md
+++ b/get-shit-done/workflows/plan-phase.md
@@ -454,43 +454,15 @@ Display banner:
  GSD ► AUTO-ADVANCING TO EXECUTE
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 
-Plans ready. Spawning execute-phase...
+Plans ready. Launching execute-phase...
 ```
 
-Spawn execute-phase as Task with direct workflow file reference (do NOT use Skill tool — Skills don't resolve inside Task subagents):
+Launch execute-phase using the Skill tool to avoid nested Task sessions (which cause runtime freezes due to deep agent nesting):
 ```
-Task(
-  prompt="
-    <objective>
-    You are the execute-phase orchestrator. Execute all plans for Phase ${PHASE}: ${PHASE_NAME}.
-    </objective>
-
-    <execution_context>
-    @~/.claude/get-shit-done/workflows/execute-phase.md
-    @~/.claude/get-shit-done/references/checkpoints.md
-    @~/.claude/get-shit-done/references/tdd.md
-    @~/.claude/get-shit-done/references/model-profile-resolution.md
-    </execution_context>
-
-    <arguments>
-    PHASE=${PHASE}
-    ARGUMENTS='${PHASE} --auto --no-transition'
-    </arguments>
-
-    <instructions>
-    1. Read execute-phase.md from execution_context for your complete workflow
-    2. Follow ALL steps: initialize, handle_branching, validate_phase, discover_and_group_plans, execute_waves, aggregate_results, close_parent_artifacts, verify_phase_goal, update_roadmap
-    3. The --no-transition flag means: after verification + roadmap update, STOP and return status. Do NOT run transition.md.
-    4. When spawning executor agents, use subagent_type='gsd-executor' with the existing @file pattern from the workflow
-    5. When spawning verifier agents, use subagent_type='gsd-verifier'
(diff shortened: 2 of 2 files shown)


Tibsfox:
Code Logic Verification of PR #826
I cloned the repo, inspected the actual diff (b3e3e3d), examined the before/after state of both modified files, and traced the full auto-advance execution chain. Here's the analysis:
1. The Before State (Bug Path)
The old code in discuss-phase.md (line ~530) and plan-phase.md (line ~454) both used this pattern for auto-advance:
Task(
  prompt="...<execution_context>@~/.claude/get-shit-done/workflows/plan-phase.md...</execution_context>...",
  subagent_type="general-purpose",
  description="Plan Phase ${PHASE}"
)
Tracing the nesting chain:

User runs /gsd:discuss-phase N --auto
discuss-phase completes → spawns Task(general-purpose) for plan-phase → nesting level 1
That Task agent reads plan-phase.md → spawns another Task(general-purpose) for execute-phase → nesting level 2
That Task agent reads execute-phase.md → spawns Task(subagent_type="gsd-executor") for each plan → nesting level 3
If executors themselves spawn anything → nesting level 4

At level 2-3, a general-purpose agent is running in a subprocess. It has the workflow instructions but lacks the full command registry context that the top-level session has. The old code even included a workaround comment: "do NOT use Skill tool — Skills don't resolve inside Task subagents" — acknowledging that within a Task subprocess, the Skill tool can't find registered skills. So the execute-phase Task falls back to shelling out to claude --print, which triggers the "Claude Code cannot be launched inside another Claude Code session" error. This matches issue #686.
2. The After State (Fix Path)
The fix replaces the Task(general-purpose) blocks with:
discuss-phase.md (line 615):
Skill(skill="gsd:plan-phase", args="${PHASE} --auto")
plan-phase.md (line 462):
Skill(skill="gsd:execute-phase", args="${PHASE} --auto --no-transition")
Tracing the new chain:

User runs /gsd:discuss-phase N --auto — this loads the gsd:discuss-phase slash command, which has @~/.claude/get-shit-done/workflows/discuss-phase.md in its execution context. Runs at level 0.
discuss-phase completes → calls Skill("gsd:plan-phase"). Per Claude Code docs, Skills are model-invoked and run in the same process context — no subprocess, no new agent session. Plan-phase runs at level 0.
plan-phase internally spawns Task(general-purpose) for its researcher/planner/checker agents at level 1 — these are leaf workers that don't chain further.
plan-phase completes → calls Skill("gsd:execute-phase", args="${PHASE} --auto --no-transition"). Again, same process, level 0.
execute-phase spawns Task(subagent_type="gsd-executor") for each plan at level 1 — this is the deepest nesting.

Maximum depth: 1 (down from 3-4).
3. Key Correctness Checks
Does Skill("gsd:plan-phase") resolve correctly at the top level?
Yes. I confirmed that commands/gsd/plan-phase.md and commands/gsd/execute-phase.md exist with proper frontmatter (name: gsd:plan-phase, name: gsd:execute-phase). These are installed as slash commands under ~/.claude/commands/gsd/. The Skill tool at the top-level session has full access to the command registry. The old comment "Skills don't resolve inside Task subagents" only applied when Skill was called from within a nested Task — but the fix eliminates that scenario entirely.
Does --no-transition work correctly?
Yes. I verified in execute-phase.md (lines 384-403): when --no-transition is present, execute-phase returns a ## PHASE COMPLETE status block and stops — it doesn't run transition.md or attempt further chaining. Plan-phase's handler (lines 466+) catches these return statuses (PHASE COMPLETE, GAPS FOUND, VERIFICATION FAILED) and handles them appropriately.
Do the internal worker agents still function?
Yes. Plan-phase still spawns its researcher, planner, and plan-checker as Task(general-purpose) / Task(gsd-plan-checker) at lines 209, 322, 375. Execute-phase still spawns Task(subagent_type="gsd-executor") at line 103 and Task(subagent_type="gsd-verifier") at line 298. These are all leaf workers at nesting level 1 — well within the safe operating range.
Is the allowed-tools frontmatter a concern?
I checked: discuss-phase.md's allowed-tools list includes Task but not Skill. However, this is fine because allowed-tools in Claude Code slash commands restricts which tools require no permission prompting — it doesn't prevent using other tools entirely. The Skill tool is a built-in Claude Code capability that's always available at the top-level context, and the GSD installation places these workflows as slash commands (not skills), so the Skill invocation runs within the main orchestrator's permission context.
4. Verdict
The fix is correct. The code logic holds:

Phase transitions (discuss → plan → execute) stay flat at level 0 via Skill invocations
Only leaf worker agents (executor, verifier, researcher, planner, checker) spawn as Tasks at level 1
The --no-transition flag correctly prevents execute-phase from chaining further when spawned by plan-phase's auto-advance
All skill names resolve because the Skill calls happen at the top-level session where the full command registry is available
Net result: -67 lines of fragile prompt engineering, +13 lines of clean Skill invocations, max nesting reduced from 3-4 to 1

PR #832, opened 2026-03-01
fix: replace execSync with execFileSync to prevent command injection

ashsolei:
## Summary

Replace all `execSync` shell-command patterns with `execFileSync` to eliminate command injection vectors.

### `get-shit-done/bin/lib/core.cjs`

- **`isGitIgnored()`**: Was using `execSync('git check-ignore -q -- ' + sanitizedPath)` with regex-based character stripping. Now uses `execFileSync('git', ['check-ignore', '-q', '--', targetPath])` — no shell involved, no escaping needed.
- **`execGit()`**: Was manually shell-escaping each argument with single-quote wrapping and then concatenating into `execSync('git ' + escaped.join(' '))`. Now uses `execFileSync('git', args)` directly — arguments are passed as an array to the process, bypassing the shell entirely.

### `hooks/gsd-check-update.js`

- npm version check changed from `execSync('npm view get-shit-done-cc version')` to `execFileSync('npm', ['view', 'get-shit-done-cc', 'version'])`.

## Why

`execSync` passes commands through the system shell (`/bin/sh` on Unix, `cmd.exe` on Windows), which interprets metacharacters like `;`, `|`, `$()`, backticks, etc. Even with regex sanitization, edge cases can slip through (Unicode normalization, locale-specific characters, etc.).

`execFileSync` spawns the target executable directly without a shell, making shell injection structurally impossible regardless of argument content.

## Testing

- Both functions maintain identical return signatures and behavior
- Error handling paths preserved (try/catch with same shape)
- `execFileSync` respects the same options (`cwd`, `stdio`, `encoding`, `timeout`)

### DIFF
--- a/get-shit-done/bin/lib/commands.cjs
+++ b/get-shit-done/bin/lib/commands.cjs
@@ -3,7 +3,7 @@
  */
 const fs = require('fs');
 const path = require('path');
-const { execSync } = require('child_process');
+
 const { safeReadFile, loadConfig, isGitIgnored, execGit, normalizePhaseName, comparePhaseNum, getArchivedPhaseDirs, generateSlugInternal, getMilestoneInfo, resolveModelInternal, MODEL_PROFILES, output, error, findPhaseInternal } = require('./core.cjs');
 const { extractFrontmatter } = require('./frontmatter.cjs');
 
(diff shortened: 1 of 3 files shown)


ashsolei:

PR #834, opened 2026-03-01
fix: prevent data loss when settings.json is corrupt

ashsolei:
## Summary

Prevent silent data loss when `settings.json` contains invalid JSON.

## Problem

`readSettings()` catches JSON parse errors and returns `{}`. When the installer writes back to the same file, this empty object overwrites the original settings — destroying all user configuration silently.

This can happen when:
- A manual edit introduces a syntax error (trailing comma, unquoted key)
- A concurrent write partially corrupts the file
- The file gets truncated (disk full, interrupted write)

## Fix

`readSettings()` now returns `null` on parse failure and logs a warning with the parse error message. All 5 call sites guard against `null`:

| Call site | Behavior on null |
|-----------|-----------------|
| `getCommitAttribution()` (opencode) | Treats as no config, uses default |
| `getCommitAttribution()` (gemini) | Treats as no settings, uses default |
| `getCommitAttribution()` (claude) | Treats as no settings, uses default |
| Uninstall cleanup | Skips settings.json cleanup with warning |
| Install configuration | Skips all settings modifications, returns early |

The corrupt file is preserved so the user can fix it manually and re-run the installer.

## Testing

All 462 existing tests pass.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -228,14 +228,18 @@ function buildHookCommand(configDir, hookName) {
 }
 
 /**
- * Read and parse settings.json, returning empty object if it doesn't exist
+ * Read and parse settings.json, returning empty object if it doesn't exist.
+ * Returns null if the file exists but contains invalid JSON, to prevent
+ * callers from overwriting corrupt files with empty settings (data loss).
  */
 function readSettings(settingsPath) {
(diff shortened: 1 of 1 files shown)


ashsolei:

PR #835, opened 2026-03-01
fix: replace execSync with execFileSync and add path/session validation

ashsolei:
## Security Fixes

Seven security hardening changes across 4 files. All 462 existing tests pass.

### CRITICAL: Shell Injection in `isGitIgnored()` (core.cjs)
**Before:** `execSync('git check-ignore -q -- ' + targetPath.replace(...))` — regex-based sanitization is insufficient; shell metacharacters can bypass it.
**After:** `execFileSync('git', ['check-ignore', '-q', '--', targetPath])` — no shell spawned, arguments passed directly to git as an array.

### HIGH: Shell Injection in `execGit()` (core.cjs)  
**Before:** Manual shell escaping with regex + `execSync('git ' + escaped.join(' '))` — fragile escaping can be bypassed.
**After:** `execFileSync('git', args)` — eliminates the shell entirely. Arguments are passed as an array directly to the git process.

### HIGH: Path Traversal in `cmdVerifyPathExists()` (commands.cjs)
**Before:** `path.isAbsolute(targetPath) ? targetPath : path.join(cwd, targetPath)` — no containment check; `../../../etc/passwd` is resolved outside the project.
**After:** `path.resolve(cwd, targetPath)` with containment check: `fullPath.startsWith(path.resolve(cwd) + path.sep)`. Rejects paths that escape the project directory.

### MEDIUM: Unsanitized Session ID in Context Monitor (gsd-context-monitor.js)
**Before:** Session ID used without validation in file path construction.
**After:** Added `/^[a-f0-9-]+$/i` regex validation — exits on non-UUID session IDs to prevent path injection.

### MEDIUM: Unsanitized Session ID in Statusline (gsd-statusline.js)
**Before:** `session` variable used directly in `claude-ctx-${session}.json` path.
**After:** `safeSession` variable with UUID regex validation. Empty string fallback prevents path injection through malicious session values.

### MEDIUM: Predictable Temp File Names (core.cjs)
**Before:** `gsd-${Date.now()}.json` — predictable names enable symlink attacks.
**After:** `gsd-${crypto.randomUUID()}.json` — cryptographically random, unpredictable names.

### Testing
- All **462 tests** across **88 suites** pass
- Zero regressions

### DIFF
--- a/get-shit-done/bin/lib/commands.cjs
+++ b/get-shit-done/bin/lib/commands.cjs
@@ -83,7 +83,12 @@ function cmdVerifyPathExists(cwd, targetPath, raw) {
     error('path required for verification');
   }
 
-  const fullPath = path.isAbsolute(targetPath) ? targetPath : path.join(cwd, targetPath);
+  const fullPath = path.resolve(cwd, targetPath);
+
+  // Prevent path traversal — resolved path must stay within cwd
+  if (!fullPath.startsWith(path.resolve(cwd) + path.sep) && fullPath !== path.resolve(cwd)) {
+    error('path must be within the project directory');
+  }
 
   try {
     const stats = fs.statSync(fullPath);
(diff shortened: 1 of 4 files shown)


ashsolei:

PR #840, opened 2026-03-01
fix: centralize Brave API key resolution and handle UTF-16 LE key files

VerdantForge:
## Summary

- **Bug 1 — key file never read at search time**: `cmdWebsearch` only checked `process.env.BRAVE_API_KEY`; the `~/.gsd/brave_api_key` file was detected at init time but its contents were never read during actual searches. Brave Search silently returned `available: false` even with a correctly installed key file.
- **Bug 2 — UTF-16 LE BOM not handled**: Key files created by Windows Notepad or PowerShell (`Set-Content`) are saved as UTF-16 LE with a BOM (`FF FE`). Reading with `'utf8'` produced garbled bytes, causing silent 401s with no diagnostic output.
- **Bug 3 — `brave_search` defaults to `false` regardless of key availability**: `loadConfig` hardcoded `brave_search: false` and never checked for a key at load time, so orchestrators that gate on this flag always skipped web research unless the user manually edited `config.json`.

### `core.cjs` — new `readBraveApiKey(homedir?)` helper

Single centralized function for key resolution:
1. Returns `process.env.BRAVE_API_KEY` if set
2. Falls back to reading `<homedir>/.gsd/brave_api_key`
3. Detects UTF-16 LE BOM (`FF FE`) and decodes with `utf16le` automatically
4. Returns `null` if nothing is found

Accepts an optional `homedir` override used exclusively in tests to inject an isolated tmpdir.

`loadConfig` updated to `brave_search: get('brave_search') ?? !!readBraveApiKey()` — explicit `false` in `config.json` still wins via `??`.

### `commands.cjs`, `init.cjs`, `config.cjs`

All three now call `readBraveApiKey()` instead of their previous inline detection logic. Removes ~6 lines of duplicated `os.homedir() / path.join / fs.existsSync` code from `init.cjs` and `config.cjs`.

### Tests

- `readBraveApiKey` unit tests in `core.test.cjs`: env var path, UTF-8 file, UTF-16 LE BOM file, whitespace-only file returns null, `loadConfig` auto-detect, `loadConfig` explicit-false-wins. All use an isolated `fakeHome` tmpdir — **`~/.gsd` is never touched**.
- `commands.test.cjs`: removed the two tests that wrote to `~/.gsd`; file-fallback and UTF-16 paths are covered by `readBraveApiKey` unit tests. The `cmdWebsearch` suite tests what it owns (HTTP construction, response parsing, error handling).

## Test results

```
tests: 469  pass: 469  fail: 0
```

### DIFF
--- a/get-shit-done/bin/lib/commands.cjs
+++ b/get-shit-done/bin/lib/commands.cjs
@@ -4,7 +4,7 @@
 const fs = require('fs');
 const path = require('path');
 const { execSync } = require('child_process');
-const { safeReadFile, loadConfig, isGitIgnored, execGit, normalizePhaseName, comparePhaseNum, getArchivedPhaseDirs, generateSlugInternal, getMilestoneInfo, resolveModelInternal, MODEL_PROFILES, output, error, findPhaseInternal } = require('./core.cjs');
+const { safeReadFile, readBraveApiKey, loadConfig, isGitIgnored, execGit, normalizePhaseName, comparePhaseNum, getArchivedPhaseDirs, generateSlugInternal, getMilestoneInfo, resolveModelInternal, MODEL_PROFILES, output, error, findPhaseInternal } = require('./core.cjs');
 const { extractFrontmatter } = require('./frontmatter.cjs');
 
 function cmdGenerateSlug(text, raw) {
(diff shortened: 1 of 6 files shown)


trek-e:

PR #845, opened 2026-03-01
fix: correct statusline context calculation and track hooks in manifest (#769)

### LINKED ISSUE #769: gsd-statusline.js: context calculation incorrect for Claude Code's actual context window
## Summary

The context bar in `hooks/gsd-statusline.js` uses a hardcoded 80% scaling factor to display context usage, but this doesn't reflect how Claude Code actually manages its context window. The result is a misleading percentage that doesn't map to when auto-compaction actually triggers.

## Current behavior

The v1.21.0 calculation scales raw usage against an 80% ceiling:

```js
const rawUsed = Math.max(0, Math.min(100, 100 - rem));
// Scale: 80% real usage = 100% displayed
const used = Math.min(100, Math.round((rawUsed / 80) * 100));
```

This means at 50% real usage the bar shows ~62%, and at 65% real usage it shows ~81% (orange). The thresholds (63/81/95) compound the confusion.

## Actual Claude Code context window behavior

Claude Code's `/context` command reports the window breakdown precisely. On a 200k token model:

- **Autocompact buffer: 33k tokens (16.5%)** — this is the actual reserved buffer before auto-compaction triggers
- **Free space** fills the remainder

The usable context is therefore `100% - 16.5% = 83.5%` of the window, not 80%.

## Expected behavior

The calculation should subtract the actual autocompact buffer percentage and normalize against usable context:

```js
const AUTO_COMPACT_BUFFER_PCT = 16.5; // matches Claude Code's actual buffer
const usableRemaining = Math.max(0, ((rem - AUTO_COMPACT_BUFFER_PCT) / (100 - AUTO_COMPACT_BUFFER_PCT)) * 100);
const used = Math.max(0, Math.min(100, 100 - usableRemaining));
```

With color thresholds at 50/65/80 (matching intuitive usage levels), and `toFixed(2)` for decimal precision so small changes are visible in the bar.

## Additional issue: hooks not tracked in file manifest

`hooks/gsd-statusline.js` is absent from `gsd-file-manifest.json`. This means when users customize the calculation, the installer never detects it as locally modified and never backs it up via `gsd-local-patches/`. The file gets silently overwritten on every `/gsd:update`.

`hooks/gsd-check-update.js`, `hooks/gsd-context-monitor.js`, and `hooks/gsd-intel-*.js` likely have the same issue.

**Suggested fix:** add all `hooks/` files to the manifest at install time so local modifications are preserved across updates.

## Environment

- GSD v1.21.0
- Claude Code with claude-sonnet-4-6
- Model context window: 200k tokens
- Autocompact buffer observed: 33k tokens (16.5%)

0xLeathery:
## What

Fixes the statusline context bar calculation to reflect Claude Code's actual autocompact buffer, and adds GSD hook files to the install manifest so local edits survive `/gsd:update`.

## Why

The old 80% scaling factor was arbitrary and produced misleading percentages — at 50% real usage the bar showed ~62%, and the orange/red thresholds fired too early. Claude Code actually reserves 16.5% (33k of 200k tokens) as an autocompact buffer, so the usable window is 83.5%. Separately, `gsd-statusline.js` and the other hook files were absent from `gsd-file-manifest.json`, causing them to be silently overwritten on every update without backing up local modifications.

## Testing

- [ ] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Checklist

## Breaking Changes

None

Closes #769

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- **Statusline context bar** — corrected calculation to use Claude Code's actual 16.5% autocompact buffer instead of a hardcoded 80% ceiling, so the displayed percentage now reflects when compaction is truly imminent (#769)
+  - Color thresholds updated to intuitive levels: green < 50%, yellow < 65%, orange < 80%, red ≥ 80%
+  - Display now shows two decimal places (e.g. `42.37%`) so small context changes are visible
+- **Hooks not tracked in file manifest** — `gsd-statusline.js`, `gsd-check-update.js`, and `gsd-context-monitor.js` are now included in `gsd-file-manifest.json` so local modifications are detected and backed up before `/gsd:update` (#769)
+
 ### Added
 - **Adaptive model profile** — fourth model profile (`adaptive`) that auto-selects models per-plan based on complexity evaluation (#210)
   - `evaluateComplexity()` scores plan metadata (files modified, task count, objective keywords, plan type, dependencies) on 0-10+ scale
--- a/bin/install.js
+++ b/bin/install.js
@@ -1743,6 +1743,16 @@ function writeManifest(configDir, runtime = 'claude') {
       }
     }
   }
+  if (!isCodex) {
+    const hooksDir = path.join(configDir, 'hooks');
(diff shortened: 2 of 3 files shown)


0xLeathery:
Superseded by #876 (statusline fix only) + #875 (consolidated features). The original PR bundled 9 shared feature commits inflating it to ~3,670 additions. The replacement PRs are focused and reviewable.

Issue #790, opened 2026-02-27
commit_docs: false is ignored — .planning/ files still committed to git


AbdelrhmanHamouda:
### GSD Version

1.21.0

### Runtime

Claude Code

### What happened?

With `commit_docs` set to `false` in `.planning/config.json`, GSD still commits `.planning/` files to git. This happens across multiple workflows (executor, planner, researcher, synthesizer, etc.).

In my project there are **66 commits** touching `.planning/` — all made while `commit_docs: false` was set:

```
$ git log --all --oneline -- .planning/ | wc -l
      66
```

Example commits that should not have been made:
```
9cf0a41 docs(roadmap): add gap closure phase 31
360e9ba docs(phase-29): complete phase execution — verification passed
0fb9d54 docs: create milestone v2.1 roadmap (4 phases)
c2a3f5d docs: define milestone v2.1 requirements
```

This also causes a painful side effect: once `.planning/` files are committed, they become **tracked** by git. `.gitignore` only applies to untracked files, so `git add .` re-stages them. Running `git rm -r --cached .planning/` to untrack never sticks because GSD re-commits them on the next phase execution.

### What did you expect?

With `commit_docs: false`, no `.planning/` files should ever be committed to git. All agent workflows and `gsd-tools.cjs commit` should respect this setting completely.

### Steps to reproduce

1. Run `/gsd:settings` and set `commit_docs` to `false`
2. Confirm `.planning/config.json` shows `"commit_docs": false`
3. Add `.planning/` to `.gitignore`
4. Run `/gsd:plan-phase` followed by `/gsd:execute-phase`
5. Check `git log --oneline -- .planning/` — commits to `.planning/` files appear despite the setting

### Relevant logs or error messages

```shell
# .planning/config.json
{
  "model_profile": "quality",
  "commit_docs": false,
  "workflow": {
    "research": true,
    "plan_check": true,
    "verifier": true,
    "auto_advance": false,
    "nyquist_validation": true
  },
  "git": {
    "branching_strategy": "none"
  }
}

# .gitignore entry
.planning/

# The gate exists in commands.cjs but is not effective:
# if (!config.commit_docs) {
#     const result = { committed: false, hash: null, reason: 'skipped_commit_docs_false' };
#     output(result, raw, 'skipped');
#     return;
# }
#
# All doc-committing agents route through gsd-tools.cjs commit
# (executor, planner, researcher, synthesizer), so the gate should work.
# Possible causes:
[...]

PR #850, opened 2026-03-02
fix: respect commit_docs: false — stop committing .planning/ files (#790)

0xLeathery:
## What

Fix three compounding bugs that caused `.planning/` files to be committed even when `commit_docs: false` was set in config.

## Why

Users setting `commit_docs: false` expect zero `.planning/` commits. Three independent bugs defeated this: `isGitIgnored()` missed tracked files, `loadConfig()` silently fell back to `true` on parse errors, and 33 commit calls across workflows/agents had no defense-in-depth guards.

## Changes

- **`isGitIgnored()`** — added `--no-index` to `git check-ignore -q` so gitignore rules evaluate without consulting the index (fixes tracked `.planning/` files)
- **`loadConfig()`** — split single try/catch into separate file-read and JSON-parse catches; parse failures now return `commit_docs: false` (safe direction) instead of `true`
- **`add-tests.md`** — replaced raw `git add` + `git commit` with `gsd-tools.cjs commit --files` to route through the `commit_docs` gate
- **23 workflow/agent files** — added `If commit_docs is true:` guards before all 33 unguarded `gsd-tools.cjs commit` calls
- **3 new tests** — `isGitIgnored` returns true for gitignored paths, false for non-ignored paths, and true for tracked-but-gitignored paths (validates `--no-index`)
- **1 updated test** — invalid JSON config now expects `commit_docs: false`

## Testing

- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Checklist

## Breaking Changes

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## [Unreleased]
 
 ### Fixed
+- **`commit_docs: false` now fully respected** — three compounding bugs caused `.planning/` files
+  to be committed even when disabled: `isGitIgnored()` missed tracked files (added `--no-index`),
+  `loadConfig()` silently fell back to `commit_docs: true` on parse errors (now defaults to `false`),
+  and 33 workflow/agent commit calls lacked defense-in-depth guards (all now gated) (#790)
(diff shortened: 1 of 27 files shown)


0xLeathery:
Superseded by #877 (commit_docs fix only) + #875 (consolidated features). The original PR bundled 9 shared feature commits inflating it to ~3,670 additions. The replacement PRs are focused and reviewable.

PR #891, opened 2026-03-03
fix(planner): compute wave numbers for gap closure plans

### LINKED ISSUE #856: Gap closure plans hardcode wave: 1 instead of computing wave numbers
## Bug

When `/gsd:plan-phase --gaps` creates gap closure plans, the wave number is hardcoded to `1` in the plan template rather than being computed from dependency analysis like standard plans.

This means if a phase has 10 waves worth of gap closure work with dependencies between plans, they all get stamped `wave: 1` and execute in parallel — ignoring dependency ordering.

## Root Cause

In `agents/gsd-planner.md`, the `<gap_closure_mode>` section (around line 750) has a hardcoded template:

```yaml
---
phase: XX-name
plan: NN              # Sequential after existing
type: execute
wave: 1               # Gap closures typically single wave
depends_on: []
files_modified: [...]
autonomous: true
gap_closure: true     # Flag for tracking
---
```

The comment acknowledges the assumption: "Gap closures **typically** single wave." But when there are multiple gap closure plans with dependencies, they all land in wave 1 and run in parallel even when they shouldn't.

Standard plan creation goes through dependency analysis → wave computation. Gap closure plans bypass this entirely.

## Impact

- `phase-plan-index` in gsd-tools reads `wave` from frontmatter (`parseInt(fm.wave, 10) || 1`) and groups plans by wave number
- Execute-phase runs all wave-1 plans in parallel
- Gap closure plans with actual dependencies between them can fail or produce incorrect results when run simultaneously

## Recommended Fix

Remove the hardcoded `wave: 1` from the gap closure template and route gap closure plans through the same wave assignment logic used for standard plans. The planner already knows how to compute waves from `depends_on` and `files_modified` — gap closure mode just needs to use it instead of bypassing it.

Alternatively, add a note in the gap closure instructions that explicitly tells the planner: "After creating gap closure plans, compute wave numbers using the same dependency analysis as standard planning. Do not default all plans to wave 1."

Tibsfox:
## Summary

- Gap closure plans in `agents/gsd-planner.md` hardcode `wave: 1` and `depends_on: []` in the plan template
- When multiple gap closure plans have dependencies between them, they all land in wave 1 and execute in parallel — ignoring dependency ordering
- Adds an explicit wave computation step using the same `assign_waves` algorithm the planner already uses for standard plans

## Root Cause

The `<gap_closure_mode>` section (line ~862) has a hardcoded template:

```yaml
wave: 1               # Gap closures typically single wave
depends_on: []
```

Standard plan creation goes through `<step name="assign_waves">` (line 1103) which computes waves from `depends_on`. Gap closure mode bypasses this entirely.

## Fix

- Replace step 7 (write PLAN.md) with step 7 (assign waves) + step 8 (write PLAN.md)
- The new step 7 instructs the planner to compute waves using the same dependency analysis as standard planning
- The template now shows `wave: N` (computed) and `depends_on: [...]` (populated) instead of hardcoded values

## Test Plan

- [ ] Create a phase with verification gaps that have dependencies (e.g., gap A fixes a stub, gap B wires into that stub)
- [ ] Run `/gsd:plan-phase --gaps`
- [ ] Verify gap closure plans have correct wave numbers (dependent plan in wave 2+, not all wave 1)
- [ ] Verify `execute-phase` runs them in correct order

Closes #856

### DIFF
--- a/agents/gsd-planner.md
+++ b/agents/gsd-planner.md
@@ -852,15 +852,20 @@ grep -l "status: diagnosed" "$phase_dir"/*-UAT.md 2>/dev/null
 </task>
 ```
 
-**7. Write PLAN.md files:**
+**7. Assign waves using standard dependency analysis** (same as `assign_waves` step):
+- Plans with no dependencies → wave 1
+- Plans that depend on other gap closure plans → max(dependency waves) + 1
+- Also consider dependencies on existing (non-gap) plans in the phase
+
+**8. Write PLAN.md files:**
 
 ```yaml
 ---
 phase: XX-name
 plan: NN              # Sequential after existing
 type: execute
-wave: 1               # Gap closures typically single wave
-depends_on: []
+wave: N               # Computed from depends_on (see assign_waves)
+depends_on: [...]     # Other plans this depends on (gap or existing)
 files_modified: [...]
 autonomous: true
 gap_closure: true     # Flag for tracking


glittercowboy:
Cherry-picked to main — thanks for the fix! 🎯

The unique commit from this PR has been cherry-picked directly to main along with a batch of other fixes from your `dev-bugfix` branch.

Note: the shared base commits (agent frontmatter, spawn type standardization, anti-heredoc, tests) were also cherry-picked separately from the first PR in the series.

glittercowboy:
Closed — changes cherry-picked to main.

PR #892, opened 2026-03-03
fix(hooks): rename .js to .cjs for ES module compatibility

### LINKED ISSUE #125: ES module compatibility error in startup hook

Tibsfox:
## Summary

- Projects with `"type": "module"` in `package.json` treat `.js` files as ESM, causing `ReferenceError: require is not defined` in all 3 GSD hooks
- Renames `gsd-check-update.js`, `gsd-context-monitor.js`, and `gsd-statusline.js` to `.cjs`
- The `.cjs` extension explicitly forces CommonJS regardless of the project's module type

## Changes

**Hook files:**
- `hooks/gsd-check-update.js` → `hooks/gsd-check-update.cjs`
- `hooks/gsd-context-monitor.js` → `hooks/gsd-context-monitor.cjs`
- `hooks/gsd-statusline.js` → `hooks/gsd-statusline.cjs`

**`bin/install.js`:**
- Updated `buildHookCommand()` calls and local-install fallback paths
- Added old `.js` filenames to `cleanupOrphanedFiles()` so they're removed on update
- Added old `.js` hook patterns to `cleanupOrphanedHooks()` so settings.json entries are migrated
- Added statusline `.js` → `.cjs` migration for existing installs
- Updated uninstall hook cleanup list to cover both old and new filenames

**`scripts/build-hooks.js`:**
- Updated `HOOKS_TO_COPY` array

## Migration

Existing installs are handled automatically on next `/gsd:update`:
1. `cleanupOrphanedFiles()` removes old `.js` hook files from disk
2. `cleanupOrphanedHooks()` removes old `.js` hook registrations from settings.json
3. New `.cjs` files are installed and registered

## Test Plan

- [ ] All 523 existing tests pass
- [ ] Fresh install in an ESM project (`"type": "module"`) — no `require is not defined` errors
- [ ] Update from previous version — old `.js` hooks removed, new `.cjs` hooks installed
- [ ] Verify statusline, update check, and context monitor all work after migration

Closes #125


glittercowboy:
The shared base commits have been cherry-picked to main.

**Not picked:** The .js → .cjs rename (`02101eb4`) is a structural change that affects the installer, build scripts, and all hook files. Worth doing but needs its own focused PR against current main (which now has the statusline fix, stdin timeout, CLAUDE_CONFIG_DIR, and context monitor changes applied to the .js files). A clean rebase should be straightforward.

glittercowboy:
Closed — shared base on main, rename needs fresh PR against current main.

PR #921, opened 2026-03-03
fix(lifecycle): extract one-liner from summary body when not in frontmatter

j2h4u:
## What

Adds `extractOneLinerFromBody()` helper to core.cjs, used as fallback in `cmdSummaryExtract` and `cmdMilestoneComplete` when the frontmatter `one-liner` field is empty.

## Why

The summary template writes the one-liner as a `**bold**` markdown line after the `# Phase N` heading, but the code only checked frontmatter's `one-liner` field — which is often empty. This caused `summary-extract` to return `null` and `milestone complete` to miss accomplishments.

## Testing
- [x] Tested on Linux
- [ ] Tested on macOS
- [ ] Tested on Windows

## Checklist
- [x] Works on Windows (no path-sensitive changes)

## Breaking Changes
None

### DIFF
--- a/get-shit-done/bin/lib/commands.cjs
+++ b/get-shit-done/bin/lib/commands.cjs
@@ -4,7 +4,7 @@
 const fs = require('fs');
 const path = require('path');
 const { execSync } = require('child_process');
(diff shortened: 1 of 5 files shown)


trek-e:

PR #924, opened 2026-03-03
fix(state): support plain format STATE.md fields in phase/milestone complete

j2h4u:
## What

`cmdPhaseComplete` and `cmdMilestoneComplete` now use `stateReplaceField`/`stateExtractField` from `state.cjs` with fallback field names to update STATE.md, handling both `**Bold:**` and plain `Field:` formats.

## Why

The STATE.md template generates plain-format fields (`Phase: 1 of 3`, `Status: In progress`), but the code only matched bold-format fields (`**Current Phase:**`). This caused `phase complete` and `milestone complete` to silently skip all STATE.md updates on projects initialized with the current template.

Also preserves the compound `Phase: X of Y (Name)` format by parsing the existing value structure before writing.

## Testing
- [x] Tested on Linux
- [ ] Tested on macOS
- [ ] Tested on Windows

## Checklist
- [x] Works on Windows (no path-sensitive changes)

## Breaking Changes
None

### DIFF
--- a/get-shit-done/bin/lib/milestone.cjs
+++ b/get-shit-done/bin/lib/milestone.cjs
@@ -6,7 +6,7 @@ const fs = require('fs');
 const path = require('path');
 const { escapeRegex, getMilestonePhaseFilter, extractOneLinerFromBody, normalizeMd, planningPaths, output, error } = require('./core.cjs');
 const { extractFrontmatter } = require('./frontmatter.cjs');
-const { writeStateMd } = require('./state.cjs');
+const { writeStateMd, stateReplaceField } = require('./state.cjs');
 
 function cmdRequirementsMarkComplete(cwd, reqIdsRaw, raw) {
   if (!reqIdsRaw || reqIdsRaw.length === 0) {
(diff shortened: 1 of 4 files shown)


tylyp:
This is a critical fix. On my project, STATE.md uses the plain `Field:` format (generated by the template), but `extractField()` in gsd-tools expects `**Bold:**` format. The result: `phase complete` silently fails to update STATE.md fields — `completed_phases` never increments, `percent` never recalculates, and `status` stays stale.

I only caught this because STATE.md claimed 100% completion when Phase 18 hadn't started. The format mismatch causes silent failures that are very hard to trace. Documented in #956.

j2h4u:
Closing — upstream's #1234 (refactor: consolidate STATE.md field helpers) implemented the same fix more cleanly.

Our PR added inline `replaceWithFallback` helpers in each consumer. Upstream extracted this into a shared `stateReplaceFieldWithFallback()` in state.cjs — single source, no duplication. Same result (plain format support), better architecture.

Thanks for merging the broader refactor @trek-e!

PR #925, opened 2026-03-03
feat: add /gsd:review cross-AI peer review command

ashanuoc:
## Summary
- Add `/gsd:review` command that invokes external AI CLIs (Gemini, Claude, Codex) to independently review phase plans
- Produces structured REVIEWS.md with per-reviewer feedback sections (YAML frontmatter + Summary/Strengths/Concerns/Suggestions/Risk Assessment)
- Add `--reviews` flag to `/gsd:plan-phase` to incorporate review feedback into planning
- 11 new tests, all 549 tests passing

## Test plan
- [x] `npm test` — all 549 tests pass (0 failures)
- [x] `review check-cli` returns JSON with CLI availability
- [x] `review build-prompt --phase N` creates prompt file
- [x] `review write-reviews --phase N` writes REVIEWS.md
- [x] `init review N` returns phase info with `cli_available`
- [x] `init plan-phase N` includes `has_reviews` when REVIEWS.md exists

### DIFF
--- a/commands/gsd/plan-phase.md
+++ b/commands/gsd/plan-phase.md
@@ -1,7 +1,7 @@
 ---
 name: gsd:plan-phase
 description: Create detailed phase plan (PLAN.md) with verification loop
-argument-hint: "[phase] [--auto] [--research] [--skip-research] [--gaps] [--skip-verify] [--prd <file>]"
+argument-hint: "[phase] [--auto] [--research] [--skip-research] [--gaps] [--skip-verify] [--prd <file>] [--reviews]"
 agent: gsd-planner
 allowed-tools:
   - Read
@@ -35,6 +35,7 @@ Phase number: $ARGUMENTS (optional — auto-detects next unplanned phase if omit
 - `--gaps` — Gap closure mode (reads VERIFICATION.md, skips research)
 - `--skip-verify` — Skip verification loop
 - `--prd <file>` — Use a PRD/acceptance criteria file instead of discuss-phase. Parses requirements into CONTEXT.md automatically. Skips discuss-phase entirely.
+- `--reviews` — Incorporate cross-AI review feedback from REVIEWS.md into planning (run /gsd:review first)
 
 Normalize phase input in step 2 before any directory lookups.
 </context>
--- a/commands/gsd/review.md
+++ b/commands/gsd/review.md
@@ -1,0 +1,36 @@
+---
+name: gsd:review
+description: Request cross-AI peer review of phase plans from external AI CLIs
+argument-hint: "--phase N [--gemini] [--claude] [--codex] [--all]"
+agent: gsd-planner
+allowed-tools:
+  - Read
+  - Write
+  - Bash
+  - Glob
+  - Grep
+---
+<objective>
+Invoke external AI CLIs (Gemini, Claude, Codex) to independently review phase plans. Produces a REVIEWS.md document with structured feedback from each reviewer that can be fed back into planning.
+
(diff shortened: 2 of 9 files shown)


azeitler-dotcom:
I like this idea. In fact, I've been doing something manually that I figured I should try to automate (or try to see if I could get added as a feature directly to gsd). My approach (which is not elegant for sure) is to copy my codebase right after I built my CONTEXT.MD file for a phase. I then have a separate AI (like in Codex or using GLM through Claude Code) to create the RESEARCH.MD and PLAN.MD file. I then have the original directory with Claude Code + Anthropic models also do an independent planning and research. I then have copied the GLM and/or Codex PLAN.md (sometimes I've done more than 1 alternative) into an "/externals" folder I have in my codebase. I tell my original Claude Code to compare the plan and research files and see if there is anything it can use that is done better in that version. It usually finds at least one thing.

I've been doing this for a few phases and it helps fill things out more. Your approach is way more streamlined. Part of what I was trying to achieve was independent producers of the files, sort of like having two people look at the same directions and see if they come to different conclusions.

Oh, and I also more recently even tried doing the split again, where I took the merged PLAN.md files (with input from all runs merged) and then let two independent directories execute on the same plan. Then I made a custom skill that produces a diff. I then feed the diff to my original instance of Claude Code so it can compare the changes it made to the one the other model made when executing the same PLAN.md. This also allowed for some better execution ideas to flow back into the original instance.

cavanaug:
With copilot you could use different models to evaluate, such as gpt-5.3-codex-xhigh vs opus vs gemini.     Would you see that as having value or does the entirely different agentic loop of calling a separate cli tool generate more value that just changing the model?

azeitler-dotcom:
> With copilot you could use different models to evaluate, such as gpt-5.3-codex-xhigh vs opus vs gemini. Would you see that as having value or does the entirely different agentic loop of calling a separate cli tool generate more value that just changing the model?

The workflow I was trying to capture is (after discussion produces the Context.md file):
Context.md -> Model A -> Plan A 
Context.md -> Model B -> Plan B
Model A -review Plan B -> Plan A+B

Right now GSD is set up to only have one plan. If you have more than one model write the plan it will overwrite the plan. I wanted to see what two separate models (given the exact same input) would produce. I carried this forward then by taking the Plan A+B and letting it execute on that plan. Then gave the diff from Model B to Model A and said, hey, here is how another model implemented the plan, is there anything good to take from this?

You could do what I do above by producing separate Plan files, maybe in different directories. And you could do the code thing by branching and doing a diff on the branch. But I haven't gone through the trouble of doing that yet.

cavanaug:
This is sort of a type of adversarial review which is becoming more common for code reviews, but this also makes great sense for plans, because a lot of work happens downstream from plans.   If your plans are bad, your implementation wont likely give you what you really want...

Multi-Agent Debate (GitHub - alecnielsen/adversarial-review): Two AI agents (e.g., Claude and GPT Codex) independently review code, then critique each other’s findings through iterative debate to reduce false positives and build consensus.

glittercowboy:
This is a solid contribution — cross-AI peer review automates a workflow users are already doing manually. Implementation follows existing GSD patterns well and the core internal touches are minimal and appropriate.

One minor note: `cmdReviewCheckCli` in `commands.cjs` duplicates the CLI check logic already in `cmdInitReview` in `init.cjs`. Would be good to deduplicate, but not a blocker.

This has merge conflicts with main. Could you rebase onto `main` and force-push? Happy to merge once it's clean.

trek-e:

PR #571, opened 2026-02-13
feat(tooling): add cross-plan conflict detection to plan-checker

davesienkowski:
## What
Extends plan-checker with new verification dimensions:
- Cross-plan file conflict detection — scans all plans in the same wave for overlapping file modifications
- `verify cross-plan-conflicts` CLI command with file reference extraction
- Enhanced plan-checker agent with Dimension 8 (cross-plan conflicts) and Step 9

## Why
Without cross-plan conflict detection, parallel wave execution can silently create conflicting changes that only surface during git merge. This catches the problem at plan time, not merge time.

## Testing
- 82/82 tests pass (7 new cross-plan conflict tests)

## Breaking Changes
None — purely additive.

PR #965, opened 2026-03-07
fix(install): strip 'skills' from Gemini/OpenCode agents to prevent validation errors

stevie86:
Gemini CLI and OpenCode runtimes fail with a validation error when subagent frontmatter contains unknown keys like 'skills:'. 

This PR updates the install script to dynamically strip the 'skills:' key and its list items when converting agents for Gemini and OpenCode. 

- Keeps source agents forward-compatible for Claude Code.
- Prevents startup crashes on Gemini and OpenCode.
- Handles both tools: and skills: array stripping correctly.

Closes #571

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -769,6 +769,7 @@ function convertClaudeToGeminiAgent(content) {
   const lines = frontmatter.split('\n');
   const newLines = [];
   let inAllowedTools = false;
+  let inSkills = false;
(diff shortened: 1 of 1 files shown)


glittercowboy:
Superseded by #1045 which removes `skills:` from all agent source files and adds stripping in the Gemini converter.

Issue #987, opened 2026-03-09
.planning files added to repository leak PII


nacklequack:
### GSD Version

1.22.4

### Runtime

Claude Code

### What happened?

I don't like the fact that my username is being leaked in the `execution_context` blocks of planning files, which then get committed to my repository:

```
<execution_context>
--
@C:/Users/<username>/.claude/get-shit-done/workflows/execute-plan.md
@C:/Users/<username>claude/get-shit-done/templates/summary.md
</execution_context>
```

### What did you expect?

I know that this is for context discovery, but is there any way to anonymize or disable this?

### Steps to reproduce

1. Initialize a project with gsd
2. Create planning for a project
3. Commit and push .planning to git repository
4. Dox your local username

### Relevant logs or error messages

```shell

```

allcounter:
Just went through this exact problem. I made my first GSD project public (a macOS dashboard app) without thinking about PII — and the `.planning/` directory turned out to be a privacy minefield.

Beyond the `execution_context` paths you mention, I found:

- **`.continue-here.md` session handoff files** — full context dumps with personal workflow notes, tool references, and decisions
- **Personal tool/system references** scattered across research files, context docs, and summaries
- **Decision rationale in `PROJECT.md`** — personal preferences baked into Key Decisions
- **Cross-references to other private projects** by name in planning files
- **Home directory paths** (`/Users/myname/...`) in code examples within research docs
- **Git history** — even after cleaning all current files, `git log -p` still exposed every old version with the original content

I ended up squashing 322 commits into one, deleting and recreating the GitHub repo, and re-cloning locally to fully clean it up.

**Suggestions for GSD:**

1. **Add `.continue-here.md` to the default `.gitignore` recommendation** — these are session-specific and should never be committed
2. **Anonymize or make `execution_context` paths relative** (e.g., `@<GSD_HOME>/workflows/execute-plan.md` instead of the full absolute path)
3. **Add a `gsd:sanitize` or `gsd:pre-publish` command** that scans `.planning/` for PII patterns (usernames, home paths, personal references) before going public
4. **Document the privacy implications** of committing `.planning/` to public repos — most users won't think about this until it's too late

The `@~/.claude/get-shit-done/...` boilerplate in every PLAN.md is the most visible leak since it's in every single plan file, but the research/context/summary files can contain much more revealing personal details depending on your setup.

rpgdev:
>but is there any way to anonymize or disable this?

[...]

PR #989, opened 2026-03-09
fix: keep Claude prompt paths home-relative

tsubasakong:
## What

Keep Claude global-install prompt file references home-relative (`~` / `$HOME`) instead of expanding them to absolute home paths, and add a focused regression test for Windows-style home paths.

## Why

Generated `.planning` files copy `execution_context` references from installed prompts today, which can bake local usernames into committed artifacts. This keeps those references portable and fixes the PII leak reported in #987.

## Testing

- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Checklist

## Breaking Changes

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


glittercowboy:
Superseded by #1047 which uses tilde paths in pathPrefix to prevent username leak in .planning/ files.

PR #999, opened 2026-03-09
fix(gemini): strip skills key and hooks comments from agent frontmatter

gridworks:
Hey TÂCHES,

First — GSD is genuinely impressive work. The context engineering approach is exactly what the AI dev tooling space needed, and building and maintaining it solo is no small feat.

I'm submitting a fix for two compounding bugs that cause all 12 GSD agents to fail loading in Gemini CLI on every launch. I discovered these through hands-on debugging and verified the fix brings the agent load count from 0/12 to 12/12 on macOS with Gemini CLI v0.32.x.

---

THE TWO BUGS

Bug 1 — `skills:` frontmatter key (affects all 12 agents)

  Validation failed: Agent Definition:
  : Unrecognized key(s) in object: 'skills'

Gemini CLI's agent schema rejects the `skills:` key entirely. Every agent file fails to load.

Bug 2 — Commented `# hooks:` block inside frontmatter (affects 8 agents)

The installer correctly comments out the hooks: block for Gemini, but leaves the commented lines INSIDE the YAML frontmatter (between the --- delimiters). Gemini CLI's YAML parser rejects comment lines inside frontmatter, causing silent loading failures — no error message, the agent just doesn't appear.

Together these two bugs mean zero GSD agents load on a fresh Gemini install.

---

THE FIX

Two filter passes added to `convertClaudeToGeminiAgent()` in bin/install.js, right before `newFrontmatter` is assembled from `newLines`:

  // Bug 1: Strip skills: key
  const filteredLines = newLines.filter(line => {
    const trimmed = line.trim();
    return !trimmed.startsWith('skills:') && !trimmed.match(/^-\s+gsd-.*-workflow$/);
  });

  // Bug 2: Strip # hooks: comment block
  const withoutHooks = filteredLines.filter(line => !line.match(/^#/));

  const newFrontmatter = withoutHooks.join('\n').trim();

That's it — 8 lines replacing 1. No other changes.

Note: replace, web_fetch, and google_web_search are all valid Gemini CLI tool names and are correctly preserved by the existing convertGeminiToolName() mapping. No tool name changes needed.

---

VERIFIED

Tested on macOS Sonoma, Gemini CLI v0.32.x, GSD v1.22.4 local install.
Result: 0/12 agents loading → 12/12 agents loading, zero errors on launch.

---

I know PR #965 is already in review for a related fix — happy for you to fold this into that PR rather than merge separately if that's cleaner.

Thanks for everything you've built.

Best,
Ali

---

ADDITIONAL NOTE — Claude references in agent prompt bodies

The agent .md files contain references to "Claude" throughout the prompt body
text (e.g. "Claude = builder", "Claude's Discretion", "Claude execution time").
These are not bugs — Gemini correctly interprets these as self-referential
instructions and executes them as intended.

What would be an actual problem (and is not present):
  - model: claude-* in frontmatter — would force wrong model
  - Claude-specific API calls hardcoded in prompts

This is an accepted limitation of GSD's write-once architecture. A future
improvement could be a runtime token substitution pass that replaces "Claude"
with the target runtime name (e.g. "Gemini") in agent body text during install,
similar to how tool names are already mapped via claudeToGeminiTools.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -823,7 +823,17 @@ function convertClaudeToGeminiAgent(content) {
     }
   }
 
-  const newFrontmatter = newLines.join('\n').trim();
+  // Bug 1: Strip skills: key — Claude Code-specific, not supported by Gemini CLI schema
+  const filteredLines = newLines.filter(line => {
+    const trimmed = line.trim();
+    return !trimmed.startsWith('skills:') && !trimmed.match(/^-\s+gsd-.*-workflow$/);
+  });
+
+  // Bug 2: Strip # hooks: comment block — Gemini CLI YAML parser rejects
+  // comments inside frontmatter even when commented out
+  const withoutHooks = filteredLines.filter(line => !line.match(/^#/));
+
+  const newFrontmatter = withoutHooks.join('\n').trim();
 
   // Escape ${VAR} patterns in agent body for Gemini CLI compatibility.
   // Gemini's templateString() treats all ${word} patterns as template variables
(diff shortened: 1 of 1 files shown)


glittercowboy:
Superseded by #1045 which removes `skills:` from all agent source files and adds stripping in the Gemini converter.

PR #1008, opened 2026-03-10
fix: emit valid Codex agent TOML

artsparkAI:
# Human Summary
Codex was unable to run gsd-planner subagent with error `agent type is currently not available`.
I was surprised to not see an issue open for this because it completely broke GSD usage within codex for me. Its possible that this may have been a quirk of just my system. 

# Agent Summary
## What happened
Generated Codex agent TOML files used TOML basic multiline strings for `developer_instructions`. Several GSD agents include raw backslashes in regexes, shell snippets, and fenced examples. In TOML basic strings those backslashes are escape sequences, so the generated `.toml` files became invalid and Codex left those agents unavailable at runtime.

## What this changes
- switch generated Codex agent `developer_instructions` to TOML literal multiline strings
- preserve agent prompt content verbatim instead of forcing TOML escape parsing
- update the Codex config test to assert the new delimiter

## Verification
- `node --test tests/codex-config.test.cjs`
- generated and parsed all `gsd-*.toml` files successfully with `tomllib` after the change

## Impact
This makes Codex installs generate valid agent configs for planner, verifier, researcher, mapper, and integration-checker prompts that contain raw backslashes.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -563,9 +563,11 @@ function generateCodexAgentToml(agentName, agentContent) {
 
   const lines = [
     `sandbox_mode = "${sandboxMode}"`,
-    `developer_instructions = """`,
+    // Agent prompts contain raw backslashes in regexes and shell snippets.
+    // TOML literal multiline strings preserve them without escape parsing.
+    `developer_instructions = '''`,
     instructions,
-    `"""`,
+    `'''`,
   ];
   return lines.join('\n') + '\n';
 }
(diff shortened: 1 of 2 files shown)


otavio (approved):

ottenhoff:
Thank you. Codex broke for me after latest update. Codex -> Settings -> Configuration clued me into the TOML validation errors preventing proper Codex startup.

onursahindur:
PS: This was merged but not tagged as a version. 
So when installing with `npx get-shit-done-cc@latest`, it does not include this fix.

janprill:
for reference in current versions: https://github.com/gsd-build/get-shit-done/issues/1018#issuecomment-4060666542

PR #1114, opened 2026-03-16
fix(build): add syntax validation to hook build script

### LINKED ISSUE #1113: v1.25.1 ships stale hooks/dist with duplicate const cwd in context-monitor
## Bug

The `gsd-context-monitor.js` hook in the published npm package (v1.25.1) has a duplicate `const cwd` declaration that crashes on every `PostToolUse` invocation, causing persistent `PostToolUse hook error` messages in Claude Code.

## Root Cause

`hooks/gsd-context-monitor.js` on `main` was fixed (the `copilot/fix-duplicate-const-cwd` branch removed the duplicate), but `hooks/dist/gsd-context-monitor.js` in the npm tarball still contains the stale pre-fix version with both declarations:

```
$ npm pack get-shit-done-cc@1.25.1
$ grep -n "const cwd" hooks/dist/gsd-context-monitor.js
47:    const cwd = data.cwd || process.cwd();
118:    const cwd = data.cwd || process.cwd();
```

Line 47 was added when the `context_warnings` config check was introduced. Line 118 was the original declaration. Both are in the same `try` block scope, so Node throws `SyntaxError: Identifier 'cwd' has already been declared` on every hook invocation.

The `prepublishOnly` script (`npm run build:hooks`) should have rebuilt `hooks/dist/` from the fixed source, but it appears v1.25.1 was published before the fix landed on `main`.

## Impact

Every PostToolUse call triggers the hook error. The hook fails silently (context monitor warnings never fire), but the error message is visible to users on every tool call.

## Fix

The source is already fixed on `main`. A patch release (v1.25.2) with `npm publish` would pick up the fix via `prepublishOnly`.

## Workaround

Edit `~/.claude/hooks/gsd-context-monitor.js` line 118 — remove the duplicate `const cwd = data.cwd || process.cwd();` line (the `cwd` variable from line 47 is already in scope).

trek-e:
## Summary

Fixes #1113 — v1.25.1 shipped `hooks/dist/` with the stale pre-fix `gsd-context-monitor.js` containing the duplicate `const cwd` declaration.

## Root Cause

The `prepublishOnly` script (`build-hooks.js`) copies hooks from `hooks/` to `hooks/dist/`, but had no validation. v1.25.1 was published before the `const cwd` fix landed on `main`, so the npm tarball contained the broken hook.

## Fix

Enhanced `scripts/build-hooks.js` to validate each hook with `node --check` after copying. If any hook has a syntax error, the build fails with exit code 1, **preventing `npm publish` from shipping broken hooks**.

Verified that `node --check` correctly catches the duplicate `const cwd` pattern:

```
SyntaxError: Identifier 'cwd' has already been declared
```

## What This Means for Release

The source `hooks/gsd-context-monitor.js` is already fixed on `main` (PR #1105). The next `npm publish` will run `prepublishOnly → build-hooks.js`, which will:
1. Copy the fixed source to `hooks/dist/`
2. Validate all hooks pass `node --check`
3. Only then allow publish to proceed

All 755 tests pass.

### DIFF
--- a/scripts/build-hooks.js
+++ b/scripts/build-hooks.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 /**
  * Copy GSD hooks to dist for installation.
+ * Runs via prepublishOnly to ensure hooks/dist/ is always fresh before npm publish.
  */
 
 const fs = require('fs');
@@ -22,6 +23,8 @@ function build() {
     fs.mkdirSync(DIST_DIR, { recursive: true });
   }
 
+  let errors = 0;
+
   // Copy hooks to dist
   for (const hook of HOOKS_TO_COPY) {
     const src = path.join(HOOKS_DIR, hook);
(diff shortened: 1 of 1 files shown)


trek-e:
This PR is superseded by PR #1165 which was already merged to main. Both implement the same syntax validation for the hook build script. This PR can be closed.

trek-e:
Closing: superseded by PR #1165 which was merged to main with the same fix (syntax validation in hook build script).

PR #1143, opened 2026-03-17
fix(codex): include required agent metadata in TOML

MiaoDX:
## Summary
- include `name` and `description` in generated Codex agent TOML files
- fall back to safe non-empty metadata when frontmatter fields are missing
- extend Codex config tests to cover the stricter agent-role schema

## Why
Recent Codex builds validate agent role TOML more strictly and warn or break when `~/.codex/agents/*.toml` omits required metadata. Current GSD installs still generate agent TOML with only `sandbox_mode` and `developer_instructions`, which leaves files like `gsd-user-profiler.toml` malformed for Codex.

This keeps the existing installer flow, but makes the generated agent files match the required schema.

Related upstream context:
- openai/codex#14451

## Verification
- `node --test tests/codex-config.test.cjs`
- smoke-ran `node bin/install.js --codex --global --config-dir "$(mktemp -d)"` and verified generated agent TOML includes `name`, `description`, `sandbox_mode`, and `developer_instructions`

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -819,14 +819,22 @@ purpose: ${toSingleLine(description)}
 
 /**
  * Generate a per-agent .toml config file for Codex.
- * Sets sandbox_mode and developer_instructions from the agent markdown body.
+ * Sets required agent metadata, sandbox_mode, and developer_instructions
+ * from the agent markdown content.
  */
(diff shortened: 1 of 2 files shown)


MiaoDX:
Superseded by a PR from the requested org fork: ChaptersOfFloatingLife/get-shit-done.

PR #1147, opened 2026-03-17
fix(quick): use lighter planner defaults for trivial quick tasks

### LINKED ISSUE #609: /gsd:quick can be quite time (and token) consuming for trivial quick tasks
### Problem or motivation

I occasionally have simple things to execute, maybe I had to close claude in the middle, or it forgot to commit some stuff. I still want GSD to steer the process rather than interact directly with claude, or claude itself might decide to use `gsd:quick` on its own. The problem is that it starts a whole planning exercise (using Opus), for something that's pretty trivial... It ends up taking time and using a lot more tokens than feels necessary.

### Proposed solution

I'm not sure what's the best approach, but a couple of options off the top of my head:

* use `quick` only for really quick stuff that shouldn't require much thinking, and default to cheaper models, then maybe add a `gsd:task` or `gsd:adhoc` for something that's more involved?
* ask which model to use for this quick task?

### Alternatives considered

_No response_

### Additional context

_No response_

Solvely-Colin:
Closes #609

## Summary
- add a dedicated `gsd-quick-planner` profile so standard `/gsd:quick` runs use a lighter planner default
- keep `--full` on the existing `gsd-planner` profile so stricter quick runs still get heavyweight planning
- document the split in quick command docs and model-profile references
- add tests for quick planner resolution, overrides, and init quick output

## Why
Issue #609 called out that trivial quick tasks still paid the full planning cost in balanced mode because `/gsd:quick` always resolved the standard planner profile, which defaults to Opus. This keeps milestone-scale planning untouched while making the default quick path cheaper and faster.

## Testing
- `npm test -- --test-name-pattern="resolveModelInternal|cmdInitQuick"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ Then `/gsd:new-milestone` starts the next version — same flow as `new-project`
 
 Quick mode gives you GSD guarantees (atomic commits, state tracking) with a faster path:
 
-- **Same agents** — Planner + executor, same quality
+- **Lighter default planner** — Standard `/gsd:quick` uses the quick-planner profile to keep trivial tasks moving
 - **Skips optional steps** — No research, no plan checker, no verifier by default
 - **Separate tracking** — Lives in `.planning/quick/`, not phases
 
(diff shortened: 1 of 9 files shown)


trek-e:

PR #1150, opened 2026-03-17
feat(execute-phase): support wave-specific execution

### LINKED ISSUE #662: Execute a specific wave in a phase
### Problem or motivation

I've not found a way to do this. Some of us are limited by usage rate caps that vary by plan. Right now GSD runs all waves in a phase automatically, which makes me hit those limits very quickly. It would be helpful to have the option to execute specific waves individually to pace my work and stay within the usage limits more easily.

### Proposed solution

Add a parameter to the current execute-phase command to specify the wave(s) to execute

### Alternatives considered

_No response_

### Additional context

_No response_

Solvely-Colin:
Closes #662

## Summary
- add `--wave N` to `/gsd:execute-phase` so users can run a specific wave instead of the whole phase
- document and test the safety rule that later waves cannot run while lower matching waves are still incomplete
- keep phase verification/completion gated on all phase plans being finished, even when execution is wave-filtered

## Why
Some users need to pace phase execution to stay inside usage caps. Running a single wave at a time solves that, but only if partial execution does not accidentally mark the full phase complete.

## Testing
- `npm test -- --test-name-pattern="execute-phase.*wave|wave filter|wave-specific"`

### DIFF
--- a/commands/gsd/execute-phase.md
+++ b/commands/gsd/execute-phase.md
@@ -1,7 +1,7 @@
 ---
 name: gsd:execute-phase
 description: Execute all plans in a phase with wave-based parallelization
-argument-hint: "<phase-number> [--gaps-only]"
+argument-hint: "<phase-number> [--wave N] [--gaps-only]"
 allowed-tools:
   - Read
   - Write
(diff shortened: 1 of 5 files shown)


trek-e:

PR #1181, opened 2026-03-18
feat: materialize full config on new-project initialization

diegomarino:
## What

Add a `config-new-project` CLI command that writes a fully-materialized `.planning/config.json` with all keys and sane defaults when a new project is initialized.

## Why

The previous implementation only wrote the 6-7 keys the user explicitly chose during the wizard, leaving everything else silently resolved by `loadConfig()` at read time — making the effective config non-discoverable and hard to audit.

Closes #

## How

- `buildNewProjectConfig(userChoices)` — new single source of truth for all project defaults; performs a three-level merge: hardcoded ← `~/.gsd/defaults.json` ← user choices.
- `ensureConfigFile()` refactored to delegate to `buildNewProjectConfig({})`.
- `new-project.md` Steps 2a and 5 updated to call `config-new-project` with user-chosen keys as the only argument — remaining keys are filled from defaults automatically.
- `VALID_CONFIG_KEYS` extended with four keys that had hardcoded defaults but were previously not settable via `config-set`: `workflow.auto_advance`, `workflow.node_repair`, `workflow.node_repair_budget`, `hooks.context_warnings`.

### Platforms tested

- [x] macOS
- [x] Windows (including backslash path handling)
- [x] Linux

### Runtimes tested

### Test details

- Automated tests extended: `config.cjs` statement coverage 78.96% → 93.81%, function coverage 78.26% → 100%.
- Added full test suite for `config-set-model-profile`, `config-new-project`, and uncovered edge cases in `config-get`/`config-set`.
- Verified `ensureConfigFile()` now writes the same nested-key format (`git.branching_strategy`) as the new command; updated 2 test assertions accordingly.

## Checklist

## Breaking Changes

None

## Future Work

Existing projects with partial configs are not backfilled — `ensureConfigFile()` is idempotent and skips files that already exist. A follow-up `config-migrate` command (or a `--merge` flag) could read the current config, merge in any missing keys from `buildNewProjectConfig({})` without overwriting existing values, and write back only if anything changed.

### DIFF
--- a/.github/workflows/auto-label-issues.yml
+++ b/.github/workflows/auto-label-issues.yml
@@ -10,7 +10,7 @@ jobs:
     permissions:
       issues: write
     steps:
-      - uses: actions/github-script@v7
+      - uses: actions/github-script@v8
         with:
           script: |
             await github.rest.issues.addLabels({
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,10 +25,10 @@ jobs:
         node-version: [20, 22, 24]
 
     steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
       - name: Set up Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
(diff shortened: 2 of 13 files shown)


diegomarino:
Merged upstream's new `firecrawl` and `exa_search` API key detection into `buildNewProjectConfig()`.
 
Both follow the same pattern as `brave_search`: auto-detected from env vars or `~/.gsd/` key files and written explicitly to the materialized config. Also added `USERPROFILE` override alongside `HOME` in sandboxed tests to fix Windows CI failures.

PR #1223, opened 2026-03-19
fix(workflows): text_mode config for Claude Code remote session compatibility

### LINKED ISSUE #1214: Claude Remote (App) Support
### GSD Version

1.26.0

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

v24.14.0

### Shell

/bin/bash

### Installation Method

npm install -g get-shit-done-cc

### What happened?

Maybe it's a Claude App bug and not a GSD bug.. but here's the story

The Claude Code TUI has a /rc mode that enables the Claude App to connect to that box but with GSD it bugs on the menu-questions: Selecting and submitting an answer through the App doesn't actually submit it. The App UI shows the selected answer but it doesn't proceed until I get back to the TUI and make the selection.

### What did you expect?

Chill at the pool and order from the menu on the phone

### Steps to reproduce

Run claude
Enable remote control with `/rc`
**go to the pool**
**open the Claude app and connect to the remote session**
`/gsd:next`
**wait for a menu and get stuck**
**walk back to the box and unstuck it**
**grab a beer**
**rinse and repeat**

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

Minor — Cosmetic or edge case

### Workaround (if any)

_No response_

### Additional context

_No response_

### Privacy Checklist

trek-e:
Closes #1214

## What Changed

Claude Code's `/rc` remote mode doesn't forward TUI menu selections from the Claude App back to the host. GSD uses `AskUserQuestion` throughout its workflows for interactive menus, which breaks in remote sessions.

1. **New `workflow.text_mode` config** — When enabled, all `AskUserQuestion` calls are replaced with plain-text numbered lists. The user types a number to select.

2. **`--text` flag** — Can be passed to any command (e.g., `/gsd:discuss-phase --text`) for per-session text mode.

3. **Updated answer_validation** — The existing fallback (retry once on empty response, then switch to text) is documented alongside text_mode as the recommended approach for remote sessions.

### How to use

```bash
# Per-project (persistent)
gsd-tools config-set workflow.text_mode true

# Per-session (flag)
/gsd:discuss-phase --text
```

## Testing
- 801/801 tests passing
- New test: config-set workflow.text_mode

### DIFF
--- a/get-shit-done/bin/lib/config.cjs
+++ b/get-shit-done/bin/lib/config.cjs
@@ -16,6 +16,7 @@ const VALID_CONFIG_KEYS = new Set([
   'search_gitignored', 'brave_search',
   'workflow.research', 'workflow.plan_check', 'workflow.verifier',
   'workflow.nyquist_validation', 'workflow.ui_phase', 'workflow.ui_safety_gate',
+  'workflow.text_mode',
   'workflow._auto_chain_active',
   'git.branching_strategy', 'git.phase_branch_template', 'git.milestone_branch_template',
   'planning.commit_docs', 'planning.search_gitignored',
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -64,6 +64,7 @@ function loadConfig(cwd) {
     nyquist_validation: true,
     parallelization: true,
     brave_search: false,
+    text_mode: false, // when true, use plain-text numbered lists instead of AskUserQuestion menus
     resolve_model_ids: false, // when true, resolve aliases (opus/sonnet/haiku) to full model IDs
     context_window: 200000, // default 200k; set to 1000000 for Opus/Sonnet 4.6 1M models
   };
@@ -108,6 +109,7 @@ function loadConfig(cwd) {
       nyquist_validation: get('nyquist_validation', { section: 'workflow', field: 'nyquist_validation' }) ?? defaults.nyquist_validation,
       parallelization,
       brave_search: get('brave_search') ?? defaults.brave_search,
+      text_mode: get('text_mode', { section: 'workflow', field: 'text_mode' }) ?? defaults.text_mode,
       resolve_model_ids: get('resolve_model_ids') ?? defaults.resolve_model_ids,
       context_window: get('context_window') ?? defaults.context_window,
       model_overrides: parsed.model_overrides || null,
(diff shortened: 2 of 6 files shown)


MarlonAEC:
Hey @trek-e, great work on this! I was looking into #1214 as well and had a similar approach in mind. The `workflow.text_mode` config + `--text` flag is a clean solution.

I had a couple of questions while reading through the diff — mostly trying to understand the design choices, since I'm still learning the codebase:

### 1. Init passthrough for `text_mode`

I noticed `text_mode` isn't added to the `cmdInit*` result objects in `init.cjs`. Since workflows read config flags from the init JSON (e.g., `commit_docs`, `parallelization`, etc.), would the workflows actually be able to see the `text_mode` flag at runtime? Or is there another mechanism I'm not aware of where workflows read config directly?

### 2. Coverage across other workflows

The updated `discuss-phase.md` mentions "Text mode applies to ALL workflows in the session, not just discuss-phase" — but I only see the instructions added to `discuss-phase.md` and `settings.md`. The other ~20 workflows that use `AskUserQuestion` (like `new-project.md`, `plan-phase.md`, `execute-phase.md`, etc.) don't seem to have any text mode instructions yet.

Was this intentional — maybe relying on Claude picking up the pattern from `discuss-phase.md` since it's loaded via `required_reading`? Or would it make sense to add a shared reference file (something like `references/text-menus.md`) that could be wired into all the relevant command `<execution_context>` blocks?

Happy to help with either of these if you'd like — was already working on a similar implementation locally. Thanks for getting this started! 🙏

trek-e:
@MarlonAEC Both points were valid gaps — fixed in the latest push:

### 1. Init passthrough
`text_mode` is now included in all `cmdInit*` result objects in `init.cjs` (same pattern as `commit_docs`, `parallelization`). Workflows reading the init JSON will see `text_mode: true/false`.

### 2. Coverage across workflows
Created `references/text-mode.md` — a shared protocol document with the activation check order, presentation format, and response parsing rules. Wired it into all 25 commands that list `AskUserQuestion` in their `allowed-tools` via `<execution_context>` references.

Added a regression test that catches any future command using `AskUserQuestion` without the text-mode reference — so this can't silently regress.

Thanks for the review.

PR #1227, opened 2026-03-19
fix(security): prevent path traversal in summary-extract command

tonymfer:
## What
Prevent path traversal in the `summary-extract` command.

## Why
The `summaryPath` parameter is joined with `cwd` using `path.join()` without validating the resolved path stays within the project directory. A path containing `../` sequences could read arbitrary files outside `.planning/`.

## How
Added path normalization and boundary check after `path.join()`. The resolved path must start with the project root to proceed.

## Testing
- [x] macOS
- [ ] Windows
- [ ] Linux

**Runtimes tested:**
- [x] Claude Code

## Checklist
- [x] Follows GSD style — no filler
- [x] No unnecessary dependencies
- [x] Windows-compatible (uses `path.sep` not hardcoded `/`)
- [x] Existing tests pass

## Breaking Changes
Commands that previously used `../` in summary paths will now be rejected. This is intentional — such paths should not have been allowed.

### DIFF
--- a/get-shit-done/bin/lib/commands.cjs
+++ b/get-shit-done/bin/lib/commands.cjs
@@ -269,6 +269,11 @@ function cmdSummaryExtract(cwd, summaryPath, fields, raw) {
   }
 
   const fullPath = path.join(cwd, summaryPath);
+  const normalized = path.resolve(fullPath);
+  const projectRoot = path.resolve(cwd);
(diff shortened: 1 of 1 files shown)


tonymfer:
Closing — this was speculative, not based on a real issue encountered in practice. Apologies for the noise.

PR #1232, opened 2026-03-19
feat: add block-no-verify PreToolUse hook for Claude Code contributors

### LINKED ISSUE #1231: Add block-no-verify to prevent AI agents from skipping git hooks
GSD installs hooks into Claude Code settings.json. There is a gap: when Claude Code runs Bash commands, it can use the hook-bypass flag on git commit, skipping pre-commit hooks, linters, and tests. In autonomous multi-phase GSD workflows this is a real risk. Suggest adding block-no-verify@1.1.2 as a PreToolUse Bash hook alongside existing hooks. See https://github.com/tupe12334/block-no-verify. Disclosure: I am the author and maintainer of block-no-verify.

tupe12334:
Closes #1231. Add .claude/settings.json with a block-no-verify PreToolUse hook. When contributors develop GSD using Claude Code, the agent can bypass pre-commit hooks with hook-bypass flags. This adds a Bash PreToolUse hook running npx --yes block-no-verify@1.1.2 to intercept and block such commands. See https://github.com/tupe12334/block-no-verify. Disclosure: I am the author and maintainer of block-no-verify.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,15 @@
+{
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "npx --yes block-no-verify@1.1.2"
+          }
+        ]
+      }
+    ]
+  }
+}


trek-e:
## Adversarial Review — FAIL ❌

### 🔴 Supply Chain Risk: Self-Promoted External Dependency via `npx --yes`

**PR #1232** adds a `.claude/settings.json` that runs `npx --yes block-no-verify@1.1.2` as a PreToolUse hook on every Bash invocation. This is a **supply chain risk** and must not be merged.

**Findings:**

1. **Self-promotion spam pattern.** The author (`tupe12334`) has filed identical issues across dozens of unrelated projects (twenty, documenso, drizzle-orm, anthropic cookbooks, and more) all promoting the same package. This is a coordinated campaign to get `block-no-verify` embedded in as many `.claude/settings.json` files as possible. Merging this grants the author a persistent execution hook in every GSD contributor's development environment.

2. **`npx --yes` auto-installs without consent.** The `--yes` flag means npm will silently download and execute whatever is published at `block-no-verify@1.1.2` with zero user interaction. If the author's npm account is compromised, or the author publishes a malicious update to the pinned version (npm allows re-publishing yanked versions under the same semver in some edge cases), arbitrary code runs on every contributor's machine.

3. **Runs on EVERY Bash command.** The hook matcher is `"Bash"`, meaning this package is invoked for every single Bash tool call — not just git commits. This is a performance tax on every operation and a persistent arbitrary code execution vector.

4. **Version pinning is insufficient mitigation.** While `@1.1.2` pins the version, `npx --yes` still downloads it on every invocation (or uses a cache that could be poisoned). There is no integrity hash verification. A compromised npm registry or account gives the attacker code execution.

5. **GSD already handles `--no-verify` correctly.** GSD's own commit protocol in `gsd-tools.cjs` does not use `--no-verify`. If specific git hook enforcement is needed, it should be implemented in-repo (e.g., a local script checked into the project) — not via a third-party npm package executed on every Bash call.

6. **The `.claude/` directory is `.gitignore`'d in this project.** Adding a file to `.claude/settings.json` would be overwritten by GSD's own installer. This file would conflict with any existing settings.json configuration.

**Verdict:** Do not merge. This PR introduces an external supply chain dependency with persistent execution privileges, authored by someone running a mass-promotion campaign across GitHub. The stated problem (Claude using `--no-verify`) should be solved with in-repo tooling, not a third-party `npx` package.

@glittercowboy ^^

PR #1311, opened 2026-03-22
docs: clarify model profiles for non-Claude runtimes

### LINKED ISSUE #1310: Clarify model profiles for Codex and other non-Claude runtimes
### Problem or motivation

I mainly use Codex as the main implementer thanks to its current usage limits.

I recently started using gsd on my own open source project, and while I was tweaking the model setup, I realized I was not fully sure what the intended non-Claude path is supposed to be.

From what I can tell, most of the pieces already exist:
- `model_profile`
- `inherit`
- `model_overrides`
- the newer non-Claude default-model behavior

So I do not think this is a missing feature exactly. It feels more like the non-Claude story is there, but still not very clear.

On Claude, the profile system is easy to understand. On Codex / OpenAI-style runtimes, I can mostly infer what to do from code, docs, and older issues, but I do not think users should have to figure it out that way.

### Proposed solution

The two directions that seem lowest-risk to me are option 1 and option 2.

Option 1 would be the simplest: keep the current behavior and just explain it better. I think that would already help a lot if the docs were clearer about when `inherit` is the right choice, when `model_overrides` is the better fit, and what a couple of non-Claude examples look like. The upside is that this is small and unlikely to break anything. The downside is that users still have to come up with their own model mapping.

Option 2 would be a bit more opinionated, but still fairly safe: keep the current profile system and document one official non-Claude starting point, not as a hardcoded default, just as a recommended setup. For example, stronger models for planner / verifier / debugger and cheaper ones for executor / mapper / lower-risk agents. The upside is that it gives people something concrete to start from without changing the abstraction. The downside is that it gets provider-specific more quickly, and model names can age.

My guess is that option 1 is the safest, while option 2 is probably more useful if maintainers want to give users a clearer starting point.

Documentation

### Applicable runtimes

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Codex
- [ ] Copilot
- [ ] Antigravity
- [x] All runtimes

### Alternatives considered

Option 3 would be to make the profile system less Claude-shaped over time, with a clearer separation between profile intent and runtime or model family. That does sound cleaner in the long run, but it also feels like a much bigger change, so I do not think it is the right first step unless there is already interest in reworking this area more broadly.

### Additional context

If a small docs-first pass would help, I’d be happy to contribute that

trek-e:
## Summary

- Documents `resolve_model_ids: "omit"` -- the mechanism the installer sets for non-Claude runtimes (Codex, OpenCode, Gemini CLI) that was previously undocumented in user-facing docs
- Adds concrete `model_overrides` examples using non-Claude model IDs (e.g., `o3`, `o4-mini`)
- Adds a decision table explaining when to use `inherit` vs `"omit"` vs `model_overrides`
- Separates "non-Claude runtimes" (Codex, OpenCode) from "Claude Code with non-Anthropic providers" (OpenRouter, local) since they use different mechanisms

Closes #1310

## Test plan

- [x] Full test suite passes (1166 tests, 0 failures)
- [ ] Verify anchor links resolve correctly in rendered markdown
- [ ] Spot-check that Codex users following the new docs section land on a working config

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -267,7 +267,44 @@ Override specific agents without changing the entire profile:
 }
 ```
 
-Valid override values: `opus`, `sonnet`, `haiku`, `inherit`
+Valid override values: `opus`, `sonnet`, `haiku`, `inherit`, or any fully-qualified model ID (e.g., `"openai/o3"`, `"google/gemini-2.5-pro"`).
+
+### Non-Claude Runtimes (Codex, OpenCode, Gemini CLI)
+
+When GSD is installed for a non-Claude runtime, the installer automatically sets `resolve_model_ids: "omit"` in `~/.gsd/defaults.json`. This causes GSD to return an empty model parameter for all agents, so each agent uses whatever model the runtime is configured with. No additional setup is needed for the default case.
+
+If you want different agents to use different models, use `model_overrides` with fully-qualified model IDs that your runtime recognizes:
+
+```json
+{
+  "resolve_model_ids": "omit",
+  "model_overrides": {
+    "gsd-planner": "o3",
+    "gsd-executor": "o4-mini",
+    "gsd-debugger": "o3",
+    "gsd-codebase-mapper": "o4-mini"
+  }
+}
+```
+
+The intent is the same as the Claude profile tiers -- use a stronger model for planning and debugging (where reasoning quality matters most), and a cheaper model for execution and mapping (where the plan already contains the reasoning).
+
+**When to use which approach:**
+
+| Scenario | Setting | Effect |
+|----------|---------|--------|
+| Non-Claude runtime, single model | `resolve_model_ids: "omit"` (installer default) | All agents use the runtime's default model |
+| Non-Claude runtime, tiered models | `resolve_model_ids: "omit"` + `model_overrides` | Named agents use specific models, others use runtime default |
+| Claude Code with OpenRouter/local provider | `model_profile: "inherit"` | All agents follow the session model |
+| Claude Code with OpenRouter, tiered | `model_profile: "inherit"` + `model_overrides` | Named agents use specific models, others inherit |
+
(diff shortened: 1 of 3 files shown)


trek-e:
## Adversarial Review -- PR #1311

Reviewed docs against actual code behavior in `core.cjs` and `install.js`.

**Accuracy -- PASS**

All documented behaviors match the code:
- `resolve_model_ids: "omit"` returns empty string (`resolveModelInternal` line 898-899) -- correctly documented
- `model_overrides` takes precedence over `omit` (`resolveModelInternal` line 888-893) -- correctly documented
- Installer writes `resolve_model_ids: "omit"` to `~/.gsd/defaults.json` for non-Claude runtimes (`install.js` line 4178-4188) -- correctly documented
- Fully-qualified model IDs pass through as-is -- correctly documented

**Anchor links -- PASS**

- `USER-GUIDE.md` internal link `#using-non-claude-runtimes-codex-opencode-gemini-cli` matches the new heading
- `USER-GUIDE.md` cross-doc link `CONFIGURATION.md#non-claude-runtimes-codex-opencode-gemini-cli` matches the new CONFIGURATION.md heading
- No existing links to the old `#using-non-anthropic-models` anchor exist elsewhere in the repo, so the rename breaks nothing

**Test coverage -- PASS**

Existing tests in `core.test.cjs` cover all documented scenarios: `omit` behavior, `model_overrides` precedence over `omit`, fully-qualified model IDs in overrides.

**Decision table -- PASS**

The three-row table (single model / tiered models / Claude Code with non-Anthropic provider) accurately reflects the three distinct code paths.

**Localization note (non-blocking):** The zh-CN translation of `model-profiles.md` does not have the "Non-Anthropic Models" section at all, so the rename does not create drift -- it was already missing.

**Security / prompt injection / supply chain -- PASS**

No concerns. Clean docs-only PR that accurately reflects existing code behavior.

Issue #1392, opened 2026-03-26
gsd-tools.cjs generate-claude-md fails with ENOENT in node v25


scibuff:
### GSD Version

1.29.0

### Runtime

Windsurf

### Operating System

macOS

### Node.js Version

v25.7.0

### Shell

/bin/zsh

### Installation Method

npx get-shit-done-cc@latest (fresh run)

### What happened?

node gsd-tools.cjs generate-claude-md fails with ENOENT (or EISDIR if the directory already exists) when writing to the default output path .windsurf/rules/.

```
Error: ENOENT: no such file or directory, open '/Users/.../.windsurf/rules/'
    at Object.writeFileSync (node:fs:2398:20)
    at Object.cmdGenerateClaudeMd (.../profile-output.cjs:872:8)
```

### What did you expect?

The command should create/write the file .windsurf/rules (no trailing slash) — the same behavior as when using the --output workaround:

```
node gsd-tools.cjs generate-claude-md --output .windsurf/rules  # ← works fine
```

### Steps to reproduce

1. Use Node.js v25.7.0 (key detail — path.join preserves trailing slashes in this version)
2. Run node gsd-tools.cjs generate-claude-md (no --output flag)
3. Command fails at profile-output.cjs:872 (fs.writeFileSync)

Root cause: Line 851 of profile-output.cjs sets the default output path with a trailing slash:
```
outputPath = path.join(cwd, '.windsurf/rules/');
```

In Node.js v25, path.join preserves the trailing /, producing a path like /foo/.windsurf/rules/ which writeFileSync rejects (interprets as directory). Same issue on lines 775 and 779.

Suggested Fix: Remove the trailing slash from the string literal:
```
outputPath = path.join(cwd, '.windsurf/rules');
```

### Error output / logs

```shell
node “/Users/user/code/my-project/.windsurf/get-shit-done/bin/gsd-tools.cjs" generate-claude-md
node:fs:2398
    return binding.writeFileUtf8(
                   ^

Error: ENOENT: no such file or directory, open '/Users/user/code/my-project/.windsurf/rules/'
    at Object.writeFileSync (node:fs:2398:20)
    at Object.cmdGenerateClaudeMd (/Users/user/code/my-project/.windsurf/get-shit-done/bin/lib/profile-output.cjs:872:8)
    at runCommand (/Users/user/code/my-project/.windsurf/get-shit-done/bin/gsd-tools.cjs:882:21)
[...]

Issue #1391, opened 2026-03-26
\n in folder names


abrauchli:
### GSD Version

1.28.0

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

v24.14.0

### Shell

/bin/bash

### Installation Method

npx get-shit-done-cc@latest (fresh run)

### What happened?

It doesn't break anything on my system so technically not a bug but this should definitely not happen as it's not guaranteed to work on all filesystems etc. I was confused by the autocompletion at first, then I saw what happened when I used `ls` to list the folder.

bash autocompletion (without escaping):
```
.planning$ vi phases/
}                                                     04-visualization-creation-and-enqueue/                  "slug": "execution-flow-graph-view"
01-core-infrastructure/                               999.1-{                                                 "slug": "milestone-plan-editor-with-launch"
02-dashboard-and-navigation/                          999.2-{                                               
03-live-state-and-detail-view/                        999.3-queue-editor-and-reorder/
```

ls (with escaping)
```
.planning$ ls phases/
 01-core-infrastructure          04-visualization-creation-and-enqueue                                         999.3-queue-editor-and-reorder
 02-dashboard-and-navigation    '999.1-{'$'\n''  "slug": "execution-flow-graph-view"'$'\n''}'
 03-live-state-and-detail-view  '999.2-{'$'\n''  "slug": "milestone-plan-editor-with-launch"'$'\n''}'
```

### What did you expect?

Plain filenames. Doesn't have to be 8.3 oldschool DOS but without newlines would be better ;)

### Steps to reproduce

Not sure.. I probably created them with `/gsd:add-backlog [long description]`

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

### Runtime settings.json (if relevant)

```json

```

### How often does this happen?

Every time (100% reproducible)

### Impact

Blocker — Cannot use GSD at all

### Privacy Checklist

- [x] I have reviewed all pasted output for PII (usernames, paths, API keys) and redacted where necessary

abrauchli:
Thought I pasted the same text twice but I edited to fix it and it's still the same so it looks like I can't paste the newlines directly as they get replaced with `\n` by github.. escaping or not, you get the idea.

Issue #1389, opened 2026-03-25
Commands registered twice with double namespace prefix (gsd:gsd:*)


SalesTeamToolbox:
## Summary

All GSD slash commands appear twice in Claude Code's skill list — once with the correct namespace (`gsd:progress`, `gsd:plan-phase`, etc.) and again with a double prefix (`gsd:gsd:progress`, `gsd:gsd:plan-phase`, etc.).

## Cause

Commands are defined in `~/.claude/commands/gsd/*.md`. Claude Code discovers these and registers them as `gsd:*` (correct).

However, the GSD installation directory (`~/.claude/get-shit-done/`) also contains a `commands/gsd/` subdirectory (currently with `workstreams.md`). When Claude Code scans plugin directories for commands, it discovers the `~/.claude/commands/gsd/` tree a second time through the plugin namespace, producing `gsd:gsd:*` duplicates.

## Impact

- Cosmetic: the skill list is doubled in size (~60 duplicate entries)
- Both versions work identically — no functional issue
- Clutters the system prompt with redundant skill descriptions, wasting context tokens

## Reproduction

1. Install GSD normally
2. Start a Claude Code session
3. Check available skills — every `gsd:*` command has a `gsd:gsd:*` duplicate

## Suggested Fix

Either:
1. Remove or empty the `commands/` directory in the `get-shit-done/` plugin root (keep commands only in `~/.claude/commands/gsd/`)
2. Add deduplication logic so the plugin loader skips commands whose fully-qualified name matches an already-registered command

## Environment

- Claude Code (VS Code extension)
- GSD v1.27.0
- Windows 11

Tibsfox:
Fix submitted in PR #1394. The duplicate `get-shit-done/commands/gsd/workstreams.md` file has been removed — it was an identical copy of `commands/gsd/workstreams.md` that caused Claude Code to discover and register every command twice.

PR #1394, opened 2026-03-26
fix: trailing slash, slug newlines, and duplicate command registration

Tibsfox:
## Summary

- **#1392** — Remove trailing `/` from `.windsurf/rules/` path in install.js (Node v25 compat)
- **#1391** — Add missing `--raw` flag to `generate-slug` callers; cap slugs at 60 chars
- **#1389** — Remove duplicate `get-shit-done/commands/gsd/workstreams.md` causing `gsd:gsd:*` double registration

### #1392 — Windsurf ENOENT on Node v25
`bin/install.js` replaced `CLAUDE.md` with `.windsurf/rules/` (trailing slash) in 5 locations. Node v25 preserves trailing slashes in `path.join`, causing `writeFileSync` to fail with ENOENT when the path ends in `/`. Fix: remove trailing slash from all replacement strings.

### #1391 — Newlines in folder names
`commands/gsd/add-backlog.md` and `commands/gsd/thread.md` called `generate-slug` without `--raw`, capturing JSON output (with newlines) as directory names. Additionally, `generateSlugInternal` had no length cap, allowing absurdly long slugs. Fix: add `--raw` flag to both callers; cap slugs at 60 characters in both `core.cjs` and `commands.cjs`.

### #1389 — Double namespace `gsd:gsd:*`
`get-shit-done/commands/gsd/workstreams.md` was an identical copy of `commands/gsd/workstreams.md`. Claude Code discovered both paths, registering every command twice. Fix: remove the duplicate from the plugin directory.

## Test plan

- [x] `tests/windsurf-conversion.test.cjs` — updated test verifies no trailing slash in converted paths
- [x] `tests/core.test.cjs` — 2 new tests for newline stripping and 60-char truncation in `generateSlugInternal`
- [x] `tests/commands.test.cjs` — existing `generate-slug` tests pass with length cap
- [x] Full suite: **1506 tests, 0 failures**

```
# tests 1506
# suites 290
# pass 1506
# fail 0
```


trek-e (approved):
## Review: fix: trailing slash, slug newlines, and duplicate command registration

**Summary:** Omnibus fix addressing 3 distinct bugs bundled into one PR:

1. **Trailing slash (#1392):** Removes trailing `/` from `.windsurf/rules/` replacement strings in `bin/install.js`. Node v25 preserves trailing slashes in `path.join`, causing ENOENT failures.

2. **Slug newlines (#1391):** Adds `--raw` flag to `generate-slug` callers in `add-backlog.md` and `thread.md` so they get plain text instead of JSON. Also caps slugs at 60 characters in both `commands.cjs` and `core.cjs`.

3. **Duplicate registration (#1389):** Removes `get-shit-done/commands/gsd/workstreams.md` — an identical copy of `commands/gsd/workstreams.md` that caused double-registration of `gsd:gsd:*` commands.

**Correctness:**
- Trailing slash fix is correct. All 5 replacement sites are updated consistently.
- Slug `--raw` flag fix is correct. Without it, the command outputs `{"slug":"my-slug"}` which gets embedded in directory names.
- The 60-char cap via `.substring(0, 60)` is applied in both the command handler and the internal function, which is correct since they can be called independently.
- Duplicate file removal is correct — the file was genuinely duplicated.

**Test coverage:** Good across all three fixes:
- Windsurf test updated to assert no trailing slash
- Core tests added for newline stripping and 60-char truncation
- Existing command tests cover the slug generation path

**Concern — PR scope:** This bundles 3 independent fixes into one PR. While each fix is correct and well-tested, the "one concern per PR" principle makes bisection and revert easier. If any single fix needs to be reverted, all three get reverted together. Not a blocker, but worth noting for future contributions.

**Overlap note:** This PR modifies `commands.cjs` at line 16 (slug generation). PR #1408 modifies `commands.cjs` at line 257 (phase regex). Different functions, no conflict expected.

**Verdict:** APPROVE. All three fixes are correct with good test coverage. The bundling is a style concern, not a correctness one.

PR #1409, opened 2026-03-27
feat(debug): add --diagnose flag for diagnosis-only mode

### LINKED ISSUE #1396: feat: /gsd:debug should diagnose first, with optional separate fix agent
## Problem

Currently, `/gsd:debug` diagnoses and implements a fix in a single pass. For simple, obvious bugs this works well — but for complex or multi-layered problems, the agent may misidentify the root cause and apply the wrong fix, wasting time and potentially introducing new bugs.

The core issue: **diagnosis and treatment are coupled by default**, removing the developer's ability to validate the root cause before any code changes are made.

## Proposed Change

Split `/gsd:debug` into two distinct phases:

### Phase 1 — Diagnosis only (default behavior)

```
/gsd:debug "description of the problem"
```

The agent should:
1. Analyze the codebase to identify the root cause
2. Output a structured **Root Cause Report**:
   - What is broken and why
   - Which files/functions are involved
   - Confidence level in the diagnosis
   - Possible fix strategies (without implementing any)
3. **Stop** — no code changes made

### Phase 2 — Fix agent (explicit opt-in)

```
/gsd:debug "description" --fix
# or, after reading the diagnosis:
/gsd:debug:fix
```

Invokes a **separate fix agent** that receives the root cause report as input and proceeds to implement the solution.

## Why This Is Better

| | Current | Proposed |
|---|---|---|
| Developer visibility | Fix happens without validation | Root cause reviewed before any change |
| Risk | Fixes symptoms, not causes | Developer confirms diagnosis first |
| Post-mortems / docs | No structured diagnostic output | Root cause report is reusable artifact |
| Simple bugs | ✅ works | ✅ works (use  flag) |
| Complex bugs | ⚠️ risky | ✅ safer |

## Alignment with GSD Principles

This aligns with GSD's existing separation-of-concerns model (Architect → Developer → QA). Diagnosis is a **read-only, analytical** step — it should be separated from **write** operations (the fix), the same way planning is separated from execution.

## Backward Compatibility

Adding a `--fix` flag to current `/gsd:debug` preserves existing behavior for users who want a single-pass workflow.

## Example Output (Phase 1)

```
🔍 Root Cause Report
─────────────────────────────────
Problem: Stripe webhook events not being processed after deploy

Root Cause: Missing STRIPE_WEBHOOK_SECRET env var in production Lambda.
The handler at src/handlers/stripe.py:42 calls stripe.Webhook.construct_event()
which raises a ValueError when the secret is None — silently caught by the
broad except clause at line 58, returning 200 OK with no processing.

Confidence: HIGH

Files involved:
- src/handlers/stripe.py (lines 40-60)
- infra/terraform/lambda.tf (missing var binding)
- .env.example (STRIPE_WEBHOOK_SECRET not documented)

Suggested fix strategies:
1. Add STRIPE_WEBHOOK_SECRET to Lambda env via Terraform variable
2. Add explicit None-check with proper error logging before construct_event()
3. Update .env.example documentation

Run /gsd:debug:fix to proceed with implementation.
```

---

Happy to contribute a PR if there's interest. Thanks for the great tool 🙏

Tibsfox:
## Summary

- Add `--diagnose` flag to `/gsd:debug` that uses `goal: find_root_cause_only` instead of `find_and_fix`
- The `gsd-debugger` agent already fully supports this mode (used internally by `diagnose-issues` workflow) — this just exposes it to users
- After diagnosis, users can choose "Fix now" (spawns continuation agent), "Plan fix" (`/gsd:plan-phase --gaps`), or "Manual fix"

## Why

Per #1396: for complex bugs, coupling diagnosis and fix in a single pass risks misidentifying root cause and applying the wrong fix. Diagnosis is a read-only analytical step that should be separable from write operations, consistent with GSD's existing plan/execute separation.

## What Changed

`commands/gsd/debug.md` only — no agent or tooling changes needed:
- Added `--diagnose` flag documentation in `<objective>`
- Flag parsing in `<context>` section
- Goal selection: `find_root_cause_only` when `--diagnose`, `find_and_fix` otherwise
- Updated step 4 to handle both `## ROOT CAUSE FOUND` (diagnose-only) and `## DEBUG COMPLETE` (find_and_fix) returns
- Updated step 5 to support "Fix now" continuation from diagnose-only results

## Backward Compatibility

Default behavior unchanged — without `--diagnose`, the command behaves exactly as before (`find_and_fix`).

## Test plan

- [x] Full suite passes: **1506 tests, 0 failures**
- [ ] Manual: `/gsd:debug --diagnose "issue"` returns root cause report without fixing
- [ ] Manual: `/gsd:debug "issue"` still runs full find-and-fix cycle

Closes #1396

### DIFF
--- a/commands/gsd/debug.md
+++ b/commands/gsd/debug.md
@@ -1,7 +1,7 @@
 ---
 name: gsd:debug
 description: Systematic debugging with persistent state across context resets
-argument-hint: [issue description]
+argument-hint: [--diagnose] [issue description]
 allowed-tools:
   - Read
   - Bash
@@ -15,6 +15,9 @@ Debug issues using scientific method with subagent isolation.
 **Orchestrator role:** Gather symptoms, spawn gsd-debugger agent, handle checkpoints, spawn continuations.
 
 **Why subagent:** Investigation burns context fast (reading files, forming hypotheses, testing). Fresh 200k context per investigation. Main context stays lean for user interaction.
+
+**Flags:**
+- `--diagnose` — Diagnose only. Find root cause without applying a fix. Returns a structured Root Cause Report. Use when you want to validate the diagnosis before committing to a fix.
 </objective>
 
 <available_agent_types>
@@ -24,6 +27,10 @@ Valid GSD subagent types (use exact names — do not fall back to 'general-purpo
 
 <context>
 User's issue: $ARGUMENTS
+
+Parse flags from $ARGUMENTS:
+- If `--diagnose` is present, set `diagnose_only=true` and remove the flag from the issue description.
+- Otherwise, `diagnose_only=false`.
 
 Check for active sessions:
 ```bash
@@ -87,7 +94,7 @@ timeline: {timeline}
 
 <mode>
 symptoms_prefilled: true
-goal: find_and_fix
+goal: {if diagnose_only: "find_root_cause_only", else: "find_and_fix"}
 </mode>
 
 <debug_file>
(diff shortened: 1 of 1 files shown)


Tibsfox:
Superseded by a clean v2 PR — the original carried 140+ files of stale shared fixes that are now upstream. Recreated with only the --diagnose feature changes.

PR #1410, opened 2026-03-27
feat(manager): add per-step passthrough flags via config

### LINKED ISSUE #1400: feat: /gsd:manager passthrough flags per step
### Problem or motivation

The /gsd:manager is great, but it would be nice to be able to have it pass flags to the different steps it runs (ex: pass --analyze to discuss) possibly via some settings, or prompting the user

### Proposed solution

Add options to the prompt /gsd:manager spits out or a second layer of prompting for the flags to pass

Core workflow (init, plan, build, verify)

### Applicable runtimes

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Codex
- [ ] Copilot
- [ ] Antigravity
- [ ] Cursor
- [ ] Windsurf
- [x] All runtimes

### Alternatives considered

The other options are all viable but more manual:
- run the command manually with the flags then come back to /gsd:manager. The major downside here is you either lose the parallelization that /gsd:manager gives you or you have to run multiple terminals and manually do that parallelization
- specify the flags to use in the freeform text of the prompt, not sure how consistent this will work though

### Additional context

_No response_

Tibsfox:
## Summary

- Add `manager.flags.{discuss,plan,execute}` config keys for passing flags through to each step when `/gsd:manager` dispatches work
- Flags are read from `.planning/config.json`, included in `init manager` JSON output, and applied at each dispatch point in the manager workflow
- Example: `gsd-tools config-set manager.flags.discuss "--auto --analyze"` makes every discuss from the manager include those flags

## What Changed

| File | Change |
|------|--------|
| `config.cjs` | Added `manager.flags.*` to `VALID_CONFIG_KEYS` |
| `core.cjs` | Added `manager` section passthrough in `loadConfig()` |
| `init.cjs` | Read `manager.flags` from config, include in `init manager` JSON output |
| `workflows/manager.md` | Apply `manager_flags.*` to Skill args and Task prompts at each dispatch point |
| `references/planning-config.md` | Documented new config keys |
| `tests/init-manager.test.cjs` | 2 new tests: default empty flags + configured flags passthrough |

## Usage

```bash
# Configure once — persists across sessions
gsd-tools config-set manager.flags.discuss "--auto --analyze"
gsd-tools config-set manager.flags.plan "--skip-research"
gsd-tools config-set manager.flags.execute "--interactive"

# Flags are applied automatically when /gsd:manager dispatches each step
```

## Test plan

- [x] `init manager` returns `manager_flags` with defaults (empty strings) when not configured
- [x] `init manager` returns configured flags when `manager.flags.*` is set in config
- [x] Full suite: **1508 tests, 0 failures**

Closes #1400

### DIFF
--- a/get-shit-done/bin/lib/init.cjs
+++ b/get-shit-done/bin/lib/init.cjs
@@ -1028,10 +1028,23 @@ function cmdInitManager(cwd, raw) {
   const completedCount = phases.filter(p => p.disk_status === 'complete').length;
 
   // Read manager flags from config (passthrough flags for each step)
+  // Validate: flags must be CLI-safe (only --flags, alphanumeric, hyphens, spaces)
+  const sanitizeFlags = (raw) => {
+    const val = typeof raw === 'string' ? raw : '';
+    if (!val) return '';
+    // Allow only --flag patterns with alphanumeric/hyphen values separated by spaces
+    const tokens = val.split(/\s+/).filter(Boolean);
+    const safe = tokens.every(t => /^--[a-zA-Z0-9][-a-zA-Z0-9]*$/.test(t) || /^[a-zA-Z0-9][-a-zA-Z0-9_.]*$/.test(t));
+    if (!safe) {
+      process.stderr.write(`gsd-tools: warning: manager.flags contains invalid tokens, ignoring: ${val}\n`);
+      return '';
+    }
+    return val;
+  };
   const managerFlags = {
-    discuss: (config.manager && config.manager.flags && config.manager.flags.discuss) || '',
-    plan: (config.manager && config.manager.flags && config.manager.flags.plan) || '',
-    execute: (config.manager && config.manager.flags && config.manager.flags.execute) || '',
+    discuss: sanitizeFlags(config.manager && config.manager.flags && config.manager.flags.discuss),
+    plan: sanitizeFlags(config.manager && config.manager.flags && config.manager.flags.plan),
+    execute: sanitizeFlags(config.manager && config.manager.flags && config.manager.flags.execute),
   };
 
   const result = {
--- a/tests/init-manager.test.cjs
+++ b/tests/init-manager.test.cjs
@@ -453,4 +453,30 @@ describe('init manager', () => {
     assert.strictEqual(output.manager_flags.plan, '--skip-research');
     assert.strictEqual(output.manager_flags.execute, '--interactive');
   });
+
+  test('sanitizes invalid manager_flags to prevent injection (#1410)', () => {
+    writeState(tmpDir);
+    writeRoadmap(tmpDir, [{ number: '1', name: 'Test' }]);
+
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({
+        manager: {
+          flags: {
+            discuss: '; rm -rf /',
+            plan: '--valid-flag',
+            execute: '$(whoami)',
+          }
+        }
+      })
+    );
+
+    const result = runGsdTools('init manager', tmpDir);
+    const output = JSON.parse(result.output);
+
+    // Invalid flags should be sanitized to empty string
+    assert.strictEqual(output.manager_flags.discuss, '', 'injection attempt should be sanitized');
+    assert.strictEqual(output.manager_flags.plan, '--valid-flag', 'valid flag should pass through');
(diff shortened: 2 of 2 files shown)


Tibsfox:
### All review feedback addressed

**1. Bundled unrelated changes** — Resolved. This PR was rebased onto current main; the shared fixes (windsurf, slug, workstreams.md) all merged upstream via earlier PRs. Current diff is 15 files, feature-specific only.

**2. Command injection risk** — Pushed commit `f3bdf31` adding `sanitizeFlags()` validation. Flag tokens must match `--flag-name` pattern or `alphanumeric-value` pattern. Invalid tokens (like `; rm -rf /` or `$(whoami)`) are dropped with a stderr warning. Added test verifying injection attempts are sanitized.

**3. Plan/execute passthrough mechanism** — Already resolved in the rebase. The manager workflow now uses Skill() args for all three steps:
- Discuss: `Skill(skill="gsd:discuss-phase", args="{PHASE_NUM}")`
- Plan: `Skill(skill="gsd:plan-phase", args="{N} --auto {manager_flags.plan}")`
- Execute: `Skill(skill="gsd:execute-phase", args="{N} {manager_flags.execute}")`

The vague "pass them" step-8 instruction from the original version is gone — replaced with concrete Skill() arg interpolation.

Ready for re-review!

trek-e:
## Re-Review: feat(manager): add per-step passthrough flags via config

### Previous blockers — verification

**Bundled unrelated changes (VERIFIED):** The diff is 15 files, all manager-flags-specific. No windsurf, slug, or workstreams.md changes present. Blocker resolved.

**Command injection via `sanitizeFlags` (PARTIALLY ADDRESSED):**

The `sanitizeFlags` implementation:
```js
const tokens = val.split(/\s+/).filter(Boolean);
const safe = tokens.every(t => /^--[a-zA-Z0-9][-a-zA-Z0-9]*$/.test(t) || /^[a-zA-Z0-9][-a-zA-Z0-9_.]*$/.test(t));
```

The all-or-nothing rejection (any invalid token rejects the entire string) is correct and conservative.

However, the second regex (`/^[a-zA-Z0-9][-a-zA-Z0-9_.]*$/`) permits bare words like `auto`, `skip-research`, `interactive` as positional-style values. This means `"--flag value"` passes — `--flag` matches the first regex, `value` matches the second. If the downstream `Skill()` arg string is `"{N} --auto {manager_flags.discuss}"` and `manager_flags.discuss` is `"--auto somevalue"`, the agent receives `"{N} --auto --auto somevalue"`. The `somevalue` token is passed as a positional arg. Whether this causes harm depends on the phase skill's arg parsing. For the current use case (flags like `--auto --analyze --skip-research`), bare words are not needed — the second regex branch is unnecessary and expands the attack surface. Tightening to require all tokens to match `^--[a-zA-Z0-9][-a-zA-Z0-9]*$` (flags only, no values) would be safer and still covers the documented use cases.

This is not a full reversal of the injection fix — the sanitizer does block the original examples (`; rm -rf /`, `$(whoami)`). But the permissiveness of the second regex is worth tightening.

**Plan/execute passthrough mechanism (VERIFIED):** The diff confirms concrete `Skill()` arg interpolation for all three steps. The vague "pass them" step-8 instruction is gone. Discuss uses `Skill(skill="gsd:discuss-phase", args="{PHASE_NUM} {manager_flags.discuss}")`, plan and execute similarly. Blocker resolved.

### Fresh adversarial review

**`manager_flags` with empty strings.** When flags are empty strings, the Skill args become `"{PHASE_NUM} "` (trailing space). Whether this causes issues depends on the skill's argument parsing. Most argument parsers handle trailing whitespace gracefully. Low severity.

**Task prompt includes `Manager flags: {manager_flags.plan}` as plain text.** This is a natural-language instruction to the subagent, not a structured argument. An agent reading the prompt will try to interpret this line. If `manager_flags.plan` is an empty string, the line reads "Manager flags: " — slightly confusing but harmless. Consider omitting the `Manager flags:` line entirely when the value is empty.

**No prompt injection patterns in the PR itself.**

**Mergeability:** MERGEABLE.

### Verdict: APPROVED WITH NOTE

All three original blockers are resolved. The sanitizer closes the injection door for the reported attack patterns. The secondary regex permissiveness is a follow-up hardening opportunity, not a blocker — the sanitizer is strictly better than the previous unsanitized state. Applying approved label.

trek-e:
Hey @Tibsfox do you want to look at that second regex and see if you can tighten it up any?

PR #1429, opened 2026-03-27
fix(hooks): use shared cache dir and correct stale hooks path

### LINKED ISSUE #1421: Update check writes cache to wrong directory when multiple AI tool configs exist
## Bug

`gsd-check-update.js` uses `detectConfigDir()` which checks config directories in this order: `.config/opencode`, `.opencode`, `.gemini`, `.claude`. If a user has GSD installed in multiple AI tools (e.g. Claude Code + Gemini + OpenCode), the update check hook writes the cache file to whichever config dir matches first — but the **statusline hook** (`gsd-statusline.js`) reads from a potentially different location (the one that matches for the statusline's own `detectConfigDir` call, which runs in a different context).

## Reproduction

1. Have GSD installed in both `~/.claude/` and `~/.gemini/` (or `~/.config/opencode/`)
2. Start a Claude Code session
3. `gsd-check-update.js` runs, resolves `globalConfigDir` to `~/.gemini/` (wins the race)
4. Cache written to `~/.gemini/cache/gsd-update-check.json`
5. `gsd-statusline.js` looks for cache in `~/.claude/cache/gsd-update-check.json` — finds nothing
6. Update notification never appears in statusline

## Root Cause

Both hooks independently run `detectConfigDir()`, but they can resolve to different directories depending on which stale GSD installs exist on the filesystem. The first match in the priority list wins, regardless of which AI tool is actually running.

## Suggested Fix

The update check hook (and statusline) should either:
- **Detect which AI tool is running** and use that tool's config dir specifically (e.g. check for `CLAUDE_CONFIG_DIR` env var first, or detect the parent process)
- **Always write/read the cache relative to the config dir of the tool that invoked the hook**, rather than doing independent filesystem discovery
- **Or use a shared, tool-agnostic cache location** (e.g. `~/.cache/gsd/`)

## Secondary Issue

The stale hooks detection on line 71 looks for hooks in `path.join(configDir, 'get-shit-done', 'hooks')` but hooks actually live in `path.join(configDir, 'hooks')` — so stale hook warnings can never trigger.

## Environment

- GSD v1.30.0
- Node v25.6.1
- Linux (LXC on Proxmox)
- GSD installed in: `~/.claude/`, `~/.gemini/`, `~/.config/opencode/`

Tibsfox:
## Summary

Two fixes for multi-runtime installations:

1. **Update cache directory mismatch** — `gsd-check-update.js` and `gsd-statusline.js` could resolve to different runtime config dirs when multiple runtimes are installed (e.g. Claude + Gemini). Cache is now written to `~/.cache/gsd/` (tool-agnostic), with statusline reading from shared location first and falling back to legacy path.

2. **Stale hooks path wrong** — Stale hook detection checked `configDir/get-shit-done/hooks/` but hooks are installed at `configDir/hooks/`. The check could never find stale hooks.

## Root Cause

Per #1421: `detectConfigDir()` iterates `.config/opencode`, `.opencode`, `.gemini`, `.claude` and returns the first match. When multiple runtimes exist, the check-update hook and statusline hook can independently resolve to different directories.

## Files Changed

| File | Change |
|------|--------|
| `hooks/gsd-check-update.js` | Write cache to `~/.cache/gsd/`; fix hooks path to `configDir/hooks/` |
| `hooks/gsd-statusline.js` | Read from shared cache first, fall back to legacy `~/.claude/cache/` |
| `tests/core.test.cjs` | Updated stale hook path test to match corrected path |

## Test plan

- [x] Full suite: **1504 tests, 0 failures**
- [ ] Manual: install in both Claude + Gemini, verify update notification appears in statusline

Closes #1421

### DIFF
--- a/tests/core.test.cjs
+++ b/tests/core.test.cjs
@@ -1009,6 +1009,45 @@ describe('stale hook path', () => {
   });
 });
 
+// ─── shared cache directory regression (#1421) ─────────────────────────────────
+
+describe('shared cache directory (#1421)', () => {
+  test('gsd-check-update.js writes cache to shared ~/.cache/gsd/ directory', () => {
+    const content = fs.readFileSync(
+      path.join(__dirname, '..', 'hooks', 'gsd-check-update.js'), 'utf-8'
+    );
+    // Cache must use a tool-agnostic path so statusline can find it
+    // regardless of which runtime (Claude, Gemini, OpenCode) ran the check
+    assert.ok(
+      content.includes("path.join(homeDir, '.cache', 'gsd')"),
+      'check-update must write cache to ~/.cache/gsd/ (shared, tool-agnostic)'
+    );
+  });
+
+  test('gsd-statusline.js checks shared cache first, falls back to legacy (#1421)', () => {
+    const content = fs.readFileSync(
+      path.join(__dirname, '..', 'hooks', 'gsd-statusline.js'), 'utf-8'
+    );
+    // Statusline must check the shared cache path first
+    assert.ok(
+      content.includes("path.join(homeDir, '.cache', 'gsd', 'gsd-update-check.json')"),
+      'statusline must check shared cache at ~/.cache/gsd/gsd-update-check.json'
+    );
+    // Must fall back to legacy runtime-specific cache for backward compat
+    assert.ok(
+      content.includes("path.join(claudeDir, 'cache', 'gsd-update-check.json')"),
+      'statusline must fall back to legacy cache at claudeDir/cache/gsd-update-check.json'
+    );
+    // Shared cache must be checked before legacy (existsSync order matters)
+    const sharedIdx = content.indexOf('sharedCacheFile');
+    const legacyIdx = content.indexOf('legacyCacheFile');
+    assert.ok(
+      sharedIdx < legacyIdx,
+      'shared cache must be defined and checked before legacy cache'
+    );
+  });
+});
+
 // ─── resolveWorktreeRoot ─────────────────────────────────────────────────────
 
 describe('resolveWorktreeRoot', () => {


Tibsfox:
### Manual verification — partial (Claude-only, multi-runtime needs separate validation)

Verified the cache resolution logic and end-to-end flow on a Claude-only installation. Full multi-runtime testing (Claude + Gemini simultaneously) would need someone with both runtimes installed — we could only validate the single-runtime path and the code logic.

**Cache path resolution:**

| Check | Result |
|-------|--------|
| `gsd-check-update.js` writes to `~/.cache/gsd/` (shared, tool-agnostic) | PASS |
| `gsd-statusline.js` reads `~/.cache/gsd/` first | PASS |
| `gsd-statusline.js` falls back to `~/.claude/cache/` (legacy) | PASS |
| `fs.existsSync(sharedCacheFile) ? shared : legacy` selection logic | PASS |

**End-to-end cache flow:**

| Check | Result |
|-------|--------|
| Write test cache to `~/.cache/gsd/gsd-update-check.json` | PASS |
| Statusline resolves to shared cache file | PASS |
| Update notification parsed correctly (`update_available: true`) | PASS |
| Statusline would display "GSD 0.9.99-test available" | PASS |

**Legacy fallback:**

| Check | Result |
|-------|--------|
| Shared cache removed → statusline falls back to `~/.claude/cache/` | PASS |
| Neither cache exists → no update shown (graceful) | PASS |

**Stale hooks path:**

| Check | Result |
|-------|--------|
| Before: `configDir/get-shit-done/hooks/` (wrong — never finds hooks) | Confirmed broken |
| After: `configDir/hooks/` (correct — where hooks are installed) | PASS |

**What still needs multi-runtime validation:**

The core fix is that `detectConfigDir()` can resolve to different directories when multiple runtimes are installed (e.g. `~/.gemini` for check-update, `~/.claude` for statusline). The shared `~/.cache/gsd/` path eliminates this mismatch. The code logic is sound, but an actual dual-install test would confirm no runtime-specific edge cases exist (e.g. different `homeDir` resolution, symlink quirks).

If anyone has both Claude and Gemini (or OpenCode) installed, validating that update notifications appear in statusline across runtimes would complete this verification.

PR #1437, opened 2026-03-28
feat(researcher): add claim provenance tagging and assumptions log

### LINKED ISSUE #1431: GSD agents present unvalidated assumptions as decisions
## Problem

GSD agents (particularly `gsd-phase-researcher` and `gsd-planner`) present assumptions as validated decisions without citing sources or surfacing them as questions for the user.

This is particularly dangerous because users rely on these agents for expert guidance — when an agent confidently states "audit logs should be permanent" or "industry standard is X", users may not challenge it.

### Real-world example

During a security hardening phase, the research and planning agents designed an audit log system with **permanent retention** — no TTL, no cleanup, no archival. The assumption cascaded unchallenged through research → planning → execution → verification.

When challenged, the agent cited "standard practice for security audit trails" but could not name a specific regulation. In reality, retention requirements vary dramatically:
- PCI DSS: 1 year minimum
- SOX: 7 years
- GDPR: as short as reasonably necessary
- Most SaaS: company-defined policy (1-3 years)

### Scope

This isn't limited to compliance. The pattern appears whenever multiple valid approaches exist:
- Architecture decisions presented as universal ("use microservices")
- Security posture assumed without threat model context
- Performance targets picked without user research
- Auth requirements assumed without knowing the user base
- Accessibility standards assumed without knowing jurisdiction

### Add a validation rule to gsd-phase-researcher and gsd-planner agent definitions

When research or plans touch areas where requirements vary by jurisdiction, industry, or company policy (data retention, auth requirements, accessibility level, encryption standards, privacy regulations), agents should flag these as decision points rather than defaulting to an assumed standard.

Suggested addition to agent definitions:

```markdown
<validation_rule>
When your research or plan touches areas where requirements vary by jurisdiction,
industry, or company policy (data retention, auth requirements, accessibility level,
encryption standards, privacy regulations), flag these as decision points rather than
defaulting to an assumed standard. Include the range of options with their trade-offs.
"Industry standard" and "best practice" are not sources — cite the specific regulation,
study, or data point with version and date, or surface it as a question for the user.
</validation_rule>
```

This should also apply to `gsd-plan-checker` — it should flag plans that contain unvalidated assumptions about compliance, regulatory, or context-dependent requirements.

## Why This Matters

GSD agents are designed to be trusted advisors that users rely on for research and planning. An unvalidated assumption from these agents is more dangerous than one from a junior developer — it won't be questioned the same way. The fix is simple: when you don't have a specific citation, say so and ask.

## Workaround

For individual projects, adding a hard rule to `CLAUDE.md` catches GSD agents (since they read it before executing). But this requires each project to independently discover the problem. A fix in the agent definitions would make this the default behavior for all GSD users.

Tibsfox:
## Summary

- Research agents must now tag every factual claim with `[VERIFIED]`, `[CITED: url]`, or `[ASSUMED]`
- New `## Assumptions Log` section in RESEARCH.md template collects all `[ASSUMED]` claims in a table
- Downstream agents (planner, discuss-phase) can use this to surface unverified decisions for user confirmation

## Problem

Per #1431: research agents present assumptions as validated decisions without citing sources. This is dangerous because users rely on these agents for expert guidance — when an agent confidently states "audit logs should be permanent" or "industry standard is X", users may not challenge it. Assumptions then propagate unchallenged through research → planning → execution → verification.

## Changes

| File | Change |
|------|--------|
| `agents/gsd-phase-researcher.md` | Added claim provenance requirement to core responsibilities; added `## Assumptions Log` section to RESEARCH.md template |

## How It Works

Research agents tag claims inline:
- `[VERIFIED: npm registry]` — confirmed via tool during this session
- `[CITED: docs.example.com/page]` — referenced from official documentation  
- `[ASSUMED]` — based on training knowledge, not verified

The `## Assumptions Log` table at the end of RESEARCH.md collects all `[ASSUMED]` items with their section and risk-if-wrong assessment. Planners and discuss-phase can surface these for user review.

## Test plan

- [x] Full suite: **1504 tests, 0 failures**
- [x] Manual: run `/gsd:plan-phase` and verify RESEARCH.md includes provenance tags and assumptions log

Closes #1431

### DIFF
--- a/agents/gsd-phase-researcher.md
+++ b/agents/gsd-phase-researcher.md
@@ -25,6 +25,13 @@ If the prompt contains a `<files_to_read>` block, you MUST use the `Read` tool t
 - Document findings with confidence levels (HIGH/MEDIUM/LOW)
 - Write RESEARCH.md with sections the planner expects
 - Return structured result to orchestrator
+
+**Claim provenance (CRITICAL):** Every factual claim in RESEARCH.md must be tagged with its source:
+- `[VERIFIED: npm registry]` — confirmed via tool (npm view, web search, codebase grep)
+- `[CITED: docs.example.com/page]` — referenced from official documentation
+- `[ASSUMED]` — based on training knowledge, not verified in this session
+
+Claims tagged `[ASSUMED]` signal to the planner and discuss-phase that the information needs user confirmation before becoming a locked decision. Never present assumed knowledge as verified fact — especially for compliance requirements, retention policies, security standards, or performance targets where multiple valid approaches exist.
 </role>
 
 <project_context>
@@ -342,6 +349,17 @@ Verified patterns from official sources:
 
 **Deprecated/outdated:**
 - [Thing]: [why, what replaced it]
+
+## Assumptions Log
+
+> List all claims tagged `[ASSUMED]` in this research. The planner and discuss-phase use this
+> section to identify decisions that need user confirmation before execution.
+
+| # | Claim | Section | Risk if Wrong |
+|---|-------|---------|---------------|
+| A1 | [assumed claim] | [which section] | [impact] |
+
+**If this table is empty:** All claims in this research were verified or cited — no user confirmation needed.
 
 ## Open Questions
 


Tibsfox:
### Manual verification complete

Verified the researcher agent's provenance tagging and assumptions log through structural analysis of the agent prompt and template, plus downstream consumption path validation:

**Provenance tagging in agent role:**

| Check | Result |
|-------|--------|
| `Claim provenance (CRITICAL)` instruction in `<role>` section | PASS |
| `[VERIFIED: source]` tag type documented | PASS |
| `[CITED: url]` tag type documented | PASS |
| `[ASSUMED]` tag type documented | PASS |
| Safety domains called out (compliance, security, performance) | PASS |
| "Never present assumed knowledge as verified fact" guardrail | PASS |

**Assumptions Log template in RESEARCH.md:**

| Check | Result |
|-------|--------|
| `## Assumptions Log` section present in template | PASS |
| Table with columns: #, Claim, Section, Risk if Wrong | PASS |
| Empty table guidance ("All claims verified or cited") | PASS |
| Section ordering: Assumptions Log before Open Questions | PASS (line 353 < 364) |
| Downstream consumers documented (planner, discuss-phase) | PASS |

**Downstream consumption path:**

| Check | Result |
|-------|--------|
| `gsd-planner.md` reads `*-RESEARCH.md` (line 1126) | PASS |
| Planner uses research when `has_research=true` (line 1132) | PASS |
| `plan-phase.md` passes `research_path` and `has_research` from init | PASS |
| Full pipeline: researcher writes → plan-phase passes → planner reads | PASS |

The assumptions log flows correctly through the research → planning pipeline. When a researcher tags claims with `[ASSUMED]`, they'll appear in the Assumptions Log table, which the planner reads when `has_research=true`. This gives downstream agents (and users during discuss-phase) visibility into which research claims need confirmation before becoming locked decisions.

Issue #1390, opened 2026-03-25
Add Qwencode support


Azyrn:
### Problem or motivation

add QwenCode support please https://qwen.ai/qwencode

### Proposed solution

Add Qwencode support

### Which area does this affect?

Other

### Applicable runtimes

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Codex
- [ ] Copilot
- [ ] Antigravity
- [ ] Cursor
- [ ] Windsurf
- [ ] All runtimes

Tibsfox:
Same situation as #1401 — GSD is a plugin layer, not a model invoker. It works on top of host CLIs that handle the model communication.

Looking at [QwenCode](https://qwen.ai/qwencode), it appears to be a code-focused AI tool. If it offers a CLI interface with tool-use/function-calling support (similar to Claude Code or Gemini CLI), adding GSD support would involve creating an install path for it — which would be a community contribution.

Could you share more details on what QwenCode's CLI interface looks like? Specifically:
- Does it have a CLI tool-use / function-calling mode?
- Does it support custom skills/commands/plugins?
- Is there a plugin/extension API?

That would help assess feasibility.

grepsuzette:
`qwen-code` is a fork of `gemini-cli`.
It's mostly the same thing, but uses `.qwen/` dirs instead of `.gemini/`.
The system of commands FWIW is exactly the same as gemini-cli.

trek-e:
PR #1450 is open for this (Qwen-Code runtime support). Currently tagged `review: needs discussion` — an agent installation failure report from a commenter is unresolved, and menu numbering conflicts with PR #1524 (Augment Code).

---
_Generated by [Claude Code](https://claude.ai/code)_

trek-e:
Being addressed by PR #1567.

trek-e:
Thanks for the request. To move forward we need more detail:

1. **What runtime is QwenCode?** (A CLI tool? An API? A VS Code extension?)
2. **Which runtimes does it replace/extend?** (Is it compatible with OpenCode? Codex-style?)
3. **What specific GSD changes are needed?** (Install step, agent config, settings.json format?)
4. **Is there public documentation?** (A link to the QwenCode runtime spec/docs)

"Add support" alone doesn't give us enough to scope or implement the change. Please reopen or file a new issue with installation steps and a runtime spec link.

PR #1450, opened 2026-03-28
feat: add Qwen-Code runtime support

grepsuzette:
In answer to #1390.

Qwen-Code is a fork of Gemini CLI that uses the same TOML format for commands and agents.

This adds:
- `--qwen` flag for installer
- `QWEN_CONFIG_DIR` env var support (`~/.qwen` default)
- Reuses Gemini's TOML conversion for commands/agents
- Added to `--all` and interactive menu (option 9)
- Test coverage for QWEN.md instruction file

Qwen-Code: https://github.com/QwenLM/Qwen-Code


grepsuzette:
qwen-code is a fork of gemini-cli.
It's mostly the same thing, but uses .qwen/ dirs instead of .gemini/.
The system of commands FWIW is exactly the same as gemini-cli.

The implementation is therefore extremely simple.

Tested as shown below:

SS 2026-03-28 at 7 25 24 PM

ngagne:
According to the Qwen Code docs, [TOML support is deprecated](https://qwenlm.github.io/qwen-code-docs/en/users/features/commands/#toml-file-format-deprecated) in favor of Markdown. Using the Copilot runtime, and renaming all `.github` references to `.qwen` works well for me.

ngagne (changes requested):
According to the Qwen Code docs, [TOML support is deprecated](https://qwenlm.github.io/qwen-code-docs/en/users/features/commands/#toml-file-format-deprecated) in favor of Markdown. There are warnings displayed and Qwen asks to migrate the TOML files when you run.

Can we use the Copilot runtime format instead with the ".qwen" naming?

grepsuzette:
Yes toml is deprecated, the new version is now using  markdown  like Claude Code.

Reopened for review! Thanks.

ShoyebOP:
For my case the agents didn't install for some reason, no idea if I did something wrong or not

trek-e:
Would appreciate an issue being opened first, qwen-cdoe being a fork of gemini CLI is a bit too edge case for me, open code is already supported and works with the models from qwen just fine.

ShoyebOP:
@trek-e didn't know open-code had a plugin to use qwen code oauth, and the problem with the agents not installing is i think because this pr was using claude code files, so installing gsd in gemini cli and copying the agent files from there worked. still this is a very wonky solution and as i am not the creator of this pr and not have much knowledge i can't say much. 

hope @grepsuzette will add more on this

PR #1454, opened 2026-03-28
fix(sdk): skip advance step when verification finds gaps

odmrs:
## Problem

The `advance` step in `PhaseRunner.run()` executes unconditionally after `verify`, marking phases as `[x]` (complete) in ROADMAP.md even when verification returns `gaps_found`. On subsequent `auto` runs, the SDK sees `[x]` and skips the phase entirely.

**Impact observed:** In a 6-phase project, phases 2 and 3 were marked "Complete" with zero lines of code written. ~$15 spent on research/plan/verify cycles that produced nothing because the SDK kept advancing past unfinished phases.

## Fix

`phase-runner.ts` now checks if all verify steps passed before running advance:

```typescript
const verifyPassed = steps.every(s => s.step !== PhaseStepType.Verify || s.success);
if (!halted && verifyPassed) {
  const advanceResult = await this.runAdvanceStep(...);
  steps.push(advanceResult);
} else if (!halted && !verifyPassed) {
  this.logger?.warn(`Skipping advance for phase ${phaseNumber}: verification found gaps`);
}
```

When verification fails, the phase remains `[ ]` so the next `auto` run re-attempts it.

## Open question

**Should the milestone runner continue to the next phase when a phase fails verification?**

Currently it stops on first `success: false` (`index.ts` line ~197). Possible behaviors:

1. **Stop (current)** — safest, user re-runs `auto` after fixing
2. **Retry then stop** — re-attempt the phase (re-execute → re-verify) N times
3. **Config-driven** — `workflow.on_verify_fail: "stop" | "retry"`

Happy to implement whichever approach you prefer.

## Test plan

- [x] `npm run test:unit` — 675 tests passing, zero breakage
- [x] `npm run build` — zero errors
- [x] Live test: phases with `gaps_found` stay `[ ]` in ROADMAP

### DIFF
--- a/sdk/src/phase-runner.test.ts
+++ b/sdk/src/phase-runner.test.ts
@@ -577,9 +577,9 @@ describe('PhaseRunner', () => {
 
       // 1 initial + 1 retry = 2 calls (not 3)
       expect(verifyCallCount).toBe(2);
-      // Verify step still succeeds (gap closure exhausted → proceed)
+      // Verify step fails when gaps persist after exhausting retries
       const verifyStep = result.steps.find(s => s.step === PhaseStepType.Verify);
-      expect(verifyStep!.success).toBe(true);
+      expect(verifyStep!.success).toBe(false);
     });
 
     it('gaps_found triggers plan → execute → re-verify cycle', async () => {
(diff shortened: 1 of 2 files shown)


jeremymcs:
@odmrs if you can update your PR with the findings. I"ll merge it.

odmrs:
Thanks for the thorough review! Updated the PR with your findings — verify now properly returns success:false on exhausted gaps, and added the 3 regression tests you suggested. Will be more careful with these edge cases on future PRs. Mind taking another look?

PR #1462, opened 2026-03-29
fix: add environment-based runtime detection for /gsd-review

monokoo:
## Problem

The `/gsd-review` workflow relies on AI self-identification to decide which review CLI to skip. This fails when the backend model does not match the client — for example, running a Gemini model inside the Antigravity client (a Claude Code fork). The AI might incorrectly identify as Claude and skip the Claude CLI review unnecessarily.

## Solution

Replace the AI self-identification approach with **environment variable-based detection**:

| Environment | Detection | Skip Strategy |
|---|---|---|
| **Antigravity** | `ANTIGRAVITY_AGENT=1` | Skip **none** — all CLIs are external |
| **Claude Code CLI** | `CLAUDE_CODE_ENTRYPOINT` set | Skip `claude` |
| **Other** (Gemini CLI, Codex, etc.) | Fallback | AI self-identification (model matches CLI) |

### Key insight

Antigravity is a standalone client that can run *any* backend model. Since it is not any of the review CLIs (gemini/claude/codex), all three CLIs are "external" and should all be invoked for independent review.

## Verification

Tested in Antigravity environment:
```
$ env | grep ANTIGRAVITY_AGENT
ANTIGRAVITY_AGENT=1
```

Also verified via source code that Claude Code sets `CLAUDE_CODE_ENTRYPOINT` in its process environment.

### DIFF
--- a/get-shit-done/workflows/review.md
+++ b/get-shit-done/workflows/review.md
@@ -22,7 +22,7 @@ command -v codex >/dev/null 2>&1 && echo "codex:available" || echo "codex:missin
 
 Parse flags from `$ARGUMENTS`:
 - `--gemini` → include Gemini
-- `--claude` → include Claude
+- `--claude` → include the agent
 - `--codex` → include Codex
 - `--all` → include all available
 - No flags → include all available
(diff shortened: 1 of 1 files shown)


monokoo:
Closing: PR contained unintended changes beyond the target diff. Will resubmit with clean diff.

PR #1464, opened 2026-03-29
fix settings.json getting overwritten when contains comments (jsonc)

### LINKED ISSUE #1461: settings.json overwritten (lost) when settings.json contains comments
## Summary

When a `settings.json` is actually written in jsonc (with comments, which many CLI allow), GSD's installer, it fails to parse it and **silently overwrites the entire file with an empty object**, causing complete data loss.

## Severity

**Critical** - Complete data loss of user configuration

## Steps to Reproduce

1. Have a `settings.json` file in the runtime config directory (e.g., `~/.qwen/settings.json`) that contains comments:
   ```json
   {
     // My configuration
     "hooks": {
       "SessionStart": [...]
     }
   }
   ```
2. Run GSD installer: `node bin/install.js --qwen`
3. Observe that `settings.json` is replaced with `{}\`

## Root Cause

In `bin/install.js`, the `readSettings()` function (lines 384-393):

```javascript
function readSettings(settingsPath) {
  if (fs.existsSync(settingsPath)) {
    try {
      return JSON.parse(fs.readFileSync(settingsPath, 'utf8'));
    } catch (e) {
      return {};  // ← BUG: Returns empty object on parse error
    }
  }
  return {};
}
```

When `JSON.parse()` fails on JSONC content, it returns an empty object. The installer then calls `writeSettings(settingsPath, settings)` which writes `{}` to the file, **completely replacing all user configuration**.

## Recommended Fix

GSD already has a `parseJsonc()` function (lines 3724-3784) that handles JSONC. Update `readSettings()` to use it:

```javascript
function readSettings(settingsPath) {
  if (fs.existsSync(settingsPath)) {
    try {
      const content = fs.readFileSync(settingsPath, 'utf8');
      return parseJsonc(content);  // Use JSONC-aware parser
    } catch (e) {
      console.warn(`Failed to parse ${settingsPath}: ${e.message}`);
      return {};
    }
  }
  return {};
}
```

grepsuzette:
When a settings.json contains a comment (e.g. `// some comment`) GSD's installer failed to parse it and silently overwrote the entire file with an empty object to which it would add the triggers, causing loss of the original configuration.

Fixes #1461

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -369,7 +369,8 @@ function resolveOpencodeConfigPath(configDir) {
 function readSettings(settingsPath) {
   if (fs.existsSync(settingsPath)) {
     try {
-      return JSON.parse(fs.readFileSync(settingsPath, 'utf8'));
+      const content = fs.readFileSync(settingsPath, 'utf8');
+      return parseJsonc(content);
     } catch (e) {
       return {};
     }


trek-e:
superceeded by 1500, we went with that.

PR #1456, opened 2026-03-28
feat(config): add workflow.use_worktrees toggle to disable worktree isolation

Tibsfox:
## Summary

Adds `workflow.use_worktrees` config option (default: `true`) that lets users disable git worktree isolation for executor agents. This provides the escape hatch that multiple users have requested for solo development workflows where worktree merge ordering causes data loss.

```bash
# Disable worktrees
gsd-tools config-set workflow.use_worktrees false

# Or via settings UI
/gsd:settings
```

## Problem

When GSD spawns parallel executor agents with `isolation="worktree"`, each agent works in its own git worktree. The merge step back to the main branch is where things break:

- **Merge ordering is a coin flip** — when two worktrees modify overlapping files, which changes survive depends on merge order
- **Orphaned worktrees pile up** — crashed or interrupted agents leave worktrees that eat disk and confuse future operations
- **No cleanup protocol** — users report needing extra CLAUDE.md instructions to force cleanup after phase completion

For solo developers working through phases sequentially, worktree isolation is pure overhead with no upside.

## What changed

| File | Change |
|------|--------|
| `get-shit-done/bin/lib/config.cjs` | Add `workflow.use_worktrees` to valid config keys |
| `get-shit-done/references/planning-config.md` | Document the option with description |
| `get-shit-done/workflows/execute-phase.md` | Read config at init; conditional `isolation="worktree"`; sequential mode docs when disabled |
| `get-shit-done/workflows/execute-plan.md` | Conditional worktree in Pattern A |
| `get-shit-done/workflows/quick.md` | Read config; conditional worktree for quick executor |
| `get-shit-done/workflows/diagnose-issues.md` | Read config; conditional worktree for debug agents |
| `tests/config.test.cjs` | Test that `workflow.use_worktrees` can be set to `false` |
| `tests/execute-phase-wave.test.cjs` | Structural test: workflow references `USE_WORKTREES` and config read |

## Behavior when `use_worktrees: false`

- Executor agents spawn **without** `isolation="worktree"` — they run on the main working tree
- Plans within a wave execute **sequentially** (not parallel) to avoid conflicts on the same tree
- Normal git hooks run — no `--no-verify` needed since there's no hook contention
- No worktrees to merge or clean up

## Design decisions

- **Default `true`** — preserves existing behavior for users who haven't hit issues. Worktrees do provide real isolation benefits for parallel multi-plan waves.
[...]

PR #1412, opened 2026-03-27
feat(i18n): add response_language config for cross-phase language consistency

Tibsfox:
## Summary

- Add `response_language` config key (e.g. `"Portuguese"`, `"Japanese"`) that persists across all phases and subagents
- Injected into every init output via `withProjectRoot()` — no per-workflow plumbing needed
- Key workflows (discuss-phase, plan-phase, execute-phase) read the field and instruct agents to present user-facing content in the configured language

## Problem

Per #1399: language preference resets at every phase/subagent boundary. Users configure their preferred language in Claude settings or CLAUDE.md, but each spawned subagent gets a fresh context without those preferences. Correcting in-session only lasts until the next phase transition.

## Solution

A project-level `response_language` config field that:
1. Is set once: `gsd-tools config-set response_language "Portuguese"`
2. Loads via `loadConfig()` in `core.cjs`
3. Propagates to every init output via `withProjectRoot()` in `init.cjs`
4. Workflows read it and apply language directives to user-facing output and subagent prompts

**What gets translated:** AskUserQuestion labels, option text, gray area descriptions, discussion summaries, status messages.
**What stays in English:** Code, file paths, technical terms, internal agent prompts, CONTEXT.md content (consumed by downstream agents).

## Files Changed

| File | Change |
|------|--------|
| `config.cjs` | Added `response_language` to `VALID_CONFIG_KEYS` |
| `core.cjs` | Added `response_language` to `loadConfig()` return |
| `init.cjs` | Inject `response_language` in `withProjectRoot()` for all init outputs |
| `references/planning-config.md` | Documented the new config key |
| `workflows/discuss-phase.md` | Read and apply `response_language` to user-facing questions |
| `workflows/plan-phase.md` | Propagate `response_language` to spawned subagents |
| `workflows/execute-phase.md` | Propagate `response_language` to spawned subagents |
| `tests/core.test.cjs` | 2 new tests for config loading |
| `tests/init-manager.test.cjs` | 2 new tests for init output propagation |

## Test plan

- [x] `loadConfig` reads `response_language` when set, returns `null` when not
- [x] `init manager` output includes `response_language` when configured
- [x] `init manager` output omits `response_language` when not configured
- [x] Full suite: **1510 tests, 0 failures**

Closes #1399

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1475, opened 2026-03-29
feat(config): add workflow.subagent_timeout to replace hardcoded 5-minute limit

### LINKED ISSUE #1472: feat(config): too-short subagent wait / idle timeout for long-running tasks
### Problem or motivation

On large codebases, or when using slower models(GPT) that may need 10–20+ minutes before producing visible output(PLAN.md, etc.), the parent workflow can check too early, see no result yet, and stop waiting even though the subagent is still working.

There seem to be two places where this happens:

1. In `$gsd-map-codebase`, the workflow md file explicitly defines a 5-minute timeout. I have actually seen cases where the 4 subagents had just finished the research phase and were about to generate their .md outputs, but then the main agent suddenly told them to shut down and forced the workflow back to sequential execution.
- Image

2. This is not limited to map-codebase. I also see it during ordinary subagent spawning: after waiting only a few minutes, the parent loses patience and terminates the subagent immediately. 

I have not been able to find where this second behavior is defined. I did notice a 5-minute check in init.cjs -> cmdInitManager(), but I am not sure whether that is related.

### Proposed solution

Make the timeout a parameter instead of a hard-coded value.

Runtime integration (hooks, statusline, settings)

### Applicable runtimes

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Codex
- [ ] Copilot
- [ ] Antigravity
- [ ] Cursor
- [ ] Windsurf
- [ ] All runtimes

### Alternatives considered

_No response_

### Additional context

_No response_

Tibsfox:
## Summary

Addresses #1472 — the `map-codebase` workflow had a hardcoded 5-minute (300000ms) timeout for parallel subagent tasks. On large codebases or with slower models (e.g. GPT via Codex), subagents can need 10-20+ minutes before producing output, causing the parent workflow to kill still-working agents and fall back to sequential execution.

### What changed

- **New config key: `workflow.subagent_timeout`** — registered in `VALID_CONFIG_KEYS` (config.cjs), with a default of 300000ms (5 minutes) in `loadConfig()` (core.cjs). Users can increase this via `/gsd:settings workflow.subagent_timeout 900000` (15 minutes) or by editing `.planning/config.json` directly.
- **Init context propagation** — `cmdInitMapCodebase()` in init.cjs now emits `subagent_timeout` in its JSON output, making the value available to the workflow at runtime.
- **Workflow updated** — `map-codebase.md` now references the config value from init context instead of hardcoding `timeout: 300000`. Includes an inline note documenting the config key.
- **Reference docs** — `planning-config.md` updated with the new config entry, default value, and description.

### About the second timeout

The reporter mentioned a second 5-minute check in `cmdInitManager()` at init.cjs:884. After investigation, this is actually an **activity detection** threshold (`isActive = (now - newestMtime) < 300000`) that checks if any phase files were modified in the last 5 minutes — it's used for the manager dashboard display, not for subagent timeouts. It doesn't kill any processes. We've left it unchanged since it serves a different purpose, but it could be made configurable in a follow-up if needed.

### Example usage

```json
// .planning/config.json
{
  "workflow": {
    "subagent_timeout": 900000
  }
}
```

Or via CLI:
```
/gsd:settings workflow.subagent_timeout 900000
```

## Test plan

- [x] 7 new tests in `tests/subagent-timeout.test.cjs`:
  - Config key accepted by `config-set` command
  - Invalid keys still rejected (validates key registration works)
  - `subagent_timeout` appears in `init map-codebase` output with configured value
  - Default value (300000) emitted when not configured
  - Workflow file references `subagent_timeout` from init context
  - Workflow file no longer has hardcoded `timeout: 300000`
  - Reference docs include the new config entry
- [x] Full test suite passes (1511 tests, 0 failures)

Closes #1472

### DIFF
--- a/tests/subagent-timeout.test.cjs
+++ b/tests/subagent-timeout.test.cjs
@@ -123,3 +123,132 @@ describe('planning-config.md documents subagent_timeout (#1472)', () => {
     );
   });
 });
+
+// ─── init execute-phase includes context_window ─────────────────────────────
+
+describe('init execute-phase context_window (#1472)', () => {
+  let tmpDir;
+
+  beforeEach(() => {
+    tmpDir = createTempProject();
+  });
+
+  afterEach(() => {
+    cleanup(tmpDir);
+  });
+
+  test('init execute-phase output includes context_window from config', () => {
+    // Write config with a custom context_window value (1M for Opus/Sonnet 4.6)
+    const configPath = path.join(tmpDir, '.planning', 'config.json');
+    fs.writeFileSync(configPath, JSON.stringify({
+      context_window: 1000000,
+    }, null, 2));
+
+    // Create a phase directory with a plan so init execute-phase succeeds
+    const phaseDir = path.join(tmpDir, '.planning', 'phases', '01-setup');
+    fs.mkdirSync(phaseDir, { recursive: true });
+    fs.writeFileSync(path.join(phaseDir, '01-01-PLAN.md'), '# Plan');
+
+    const result = runGsdTools('init execute-phase 1', tmpDir, { HOME: tmpDir });
+    assert.ok(result.success, `Command failed: ${result.error}`);
+
+    const output = JSON.parse(result.output);
+    assert.strictEqual(output.context_window, 1000000, 'context_window should reflect configured value');
+  });
+
+  test('init execute-phase uses default context_window when not configured', () => {
+    // Write minimal config without context_window
+    const configPath = path.join(tmpDir, '.planning', 'config.json');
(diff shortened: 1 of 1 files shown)


Tibsfox:
**Heads up — minor merge overlap** 🔀

This PR adds `workflow.subagent_timeout` to the config table in `planning-config.md`. Two other open PRs (#1412 and #1456) also add entries to the same table, so whichever lands second/third will see a small conflict in that file (just adjacent table rows — trivial to resolve).

The PRs are otherwise fully independent and don't interact functionally.

PR #1484, opened 2026-03-30
feat: GSD_PROJECT env var for multi-project workspaces

NedMalki-Chief:
## Summary

- Adds `GSD_PROJECT` environment variable to `planningDir()` for project-scoped planning directory resolution
- When set, routes to `.planning/{project}/` instead of `.planning/`
- Updates `loadConfig()` to read from the project-scoped directory
- Composable with existing `GSD_WORKSTREAM`: `.planning/{project}/workstreams/{ws}/`

## Use Case

Shared workspaces where multiple independent GSD projects coexist under one `.planning/` root. Example: an Obsidian vault used as a command center for managing several code projects simultaneously.

```
.planning/
├── project-a/          ← GSD_PROJECT=project-a
│   ├── config.json
│   ├── ROADMAP.md
│   ├── STATE.md
│   └── phases/
├── project-b/          ← GSD_PROJECT=project-b
│   ├── config.json
│   ├── ROADMAP.md
│   └── phases/
└── config.json         ← root workspace config (shared)
```

Each project maintains its own `config.json`, `ROADMAP.md`, `STATE.md`, and `phases/` directory. Without `GSD_PROJECT`, all GSD commands resolve to `.planning/` root, causing cross-project collisions when multiple projects share the same workspace.

## Implementation

Follows the existing `GSD_WORKSTREAM` pattern exactly:
- New optional `project` parameter on `planningDir(cwd, ws, project)`
- Falls back to `GSD_PROJECT` env var when not passed explicitly
- Zero behavior change when `GSD_PROJECT` is unset (backwards compatible)

## Test plan

- [ ] Verify `planningDir()` returns `.planning/` when `GSD_PROJECT` is unset (no regression)
- [ ] Verify `planningDir()` returns `.planning/myproject/` when `GSD_PROJECT=myproject`
- [ ] Verify `loadConfig()` reads from `.planning/myproject/config.json` when `GSD_PROJECT=myproject`
- [ ] Verify `GSD_PROJECT` + `GSD_WORKSTREAM` compose correctly: `.planning/myproject/workstreams/ws/`
- [ ] Run existing test suite to confirm no regressions

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -195,7 +195,7 @@ function safeReadFile(filePath) {
 }
 
 function loadConfig(cwd) {
-  const configPath = path.join(cwd, '.planning', 'config.json');
+  const configPath = path.join(planningDir(cwd), 'config.json');
   const defaults = {
     model_profile: 'balanced',
     commit_docs: true,
@@ -540,20 +540,34 @@ function withPlanningLock(cwd, fn) {
 }
 
 /**
- * Get the .planning directory path, workstream-aware.
- * When a workstream is active (via explicit ws arg or GSD_WORKSTREAM env var),
- * returns `.planning/workstreams/{ws}/`. Otherwise returns `.planning/`.
+ * Get the .planning directory path, project- and workstream-aware.
+ *
+ * Resolution order:
+ * 1. If GSD_PROJECT is set (env var or explicit `project` arg), routes to
+ *    `.planning/{project}/` — supports multi-project workspaces where several
+ *    independent projects share a single `.planning/` root directory (e.g.,
+ *    an Obsidian vault or monorepo knowledge base used as a command center).
+ * 2. If GSD_WORKSTREAM is set, routes to `.planning/workstreams/{ws}/`.
+ * 3. Otherwise returns `.planning/`.
+ *
+ * GSD_PROJECT and GSD_WORKSTREAM can be combined:
+ *   `.planning/{project}/workstreams/{ws}/`
  *
  * @param {string} cwd - project root
  * @param {string} [ws] - explicit workstream name; if omitted, checks GSD_WORKSTREAM env var
+ * @param {string} [project] - explicit project name; if omitted, checks GSD_PROJECT env var
  */
-function planningDir(cwd, ws) {
+function planningDir(cwd, ws, project) {
+  if (project === undefined) project = process.env.GSD_PROJECT || null;
   if (ws === undefined) ws = process.env.GSD_WORKSTREAM || null;
-  if (!ws) return path.join(cwd, '.planning');
-  return path.join(cwd, '.planning', 'workstreams', ws);
+
+  let base = path.join(cwd, '.planning');
+  if (project) base = path.join(base, project);
+  if (ws) base = path.join(base, 'workstreams', ws);
+  return base;
 }
 
-/** Always returns the root .planning/ path, ignoring workstreams. For shared resources. */
+/** Always returns the root .planning/ path, ignoring workstreams and projects. For shared resources. */
 function planningRoot(cwd) {
   return path.join(cwd, '.planning');
 }
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -561,6 +561,15 @@ function planningDir(cwd, ws, project) {
   if (project === undefined) project = process.env.GSD_PROJECT || null;
   if (ws === undefined) ws = process.env.GSD_WORKSTREAM || null;
 
+  // Reject path separators and traversal components in project/workstream names
+  const BAD_SEGMENT = /[/\\]|\.\./;
+  if (project && BAD_SEGMENT.test(project)) {
+    throw new Error(`GSD_PROJECT contains invalid path characters: ${project}`);
+  }
+  if (ws && BAD_SEGMENT.test(ws)) {
+    throw new Error(`GSD_WORKSTREAM contains invalid path characters: ${ws}`);
+  }
+
   let base = path.join(cwd, '.planning');
   if (project) base = path.join(base, project);
   if (ws) base = path.join(base, 'workstreams', ws);
--- a/tests/core.test.cjs
+++ b/tests/core.test.cjs
@@ -30,6 +30,7 @@ const {
   findPhaseInternal,
   findProjectRoot,
   detectSubRepos,
+  planningDir,
 } = require('../get-shit-done/bin/lib/core.cjs');
 
 // ─── loadConfig ────────────────────────────────────────────────────────────────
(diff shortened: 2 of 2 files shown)


abrauchli:
Very interesting work. In your use case or the targeted application for the PR, are those independent projects all on the same codebase (sharing the master branch for releases) or living individually on diverging branches with optional cherry-picking in between?

NedMalki-Chief:
Thanks. They're fully independent projects, separate repos, separate branches, no shared codebase. The use case is an Obsidian vault that serves as a local database for all project context, with a `.planning/{project}/` subdirectory for each active project.

In practice it looks like:

```
obsidian-vault/
  .planning/
    project-alpha/    → code at github.com/org/project-alpha
    project-beta/     → code at github.com/org/project-beta
    project-gamma/    → code at github.com/org/project-gamma
```

Each project has its own full GSD lifecycle (PROJECT.md, ROADMAP.md, STATE.md, phases/). The vault provides business context (client notes, proposals, research) while each code repo holds its own implementation. `GSD_PROJECT=project-alpha` tells GSD which planning subdirectory to operate in.

I've also experimented with keeping `.planning/` in each code repo directly (avoids cross-project collisions), but the env var is still useful when you want a single terminal window managing multiple projects, switching between them by changing the env var rather than cd-ing between repos.

abrauchli:
That's awesome. I haven't used obsidian but I'm very interested to see this since I'm currently working on a gsd meta manager (TUI) to get the status of a all my independent GSD projects. Mine is AI-disabled, so purely a view of the relics left by GSD (inferred state, phases, git log viewer, ...) - I'll open it when I'm ready for a public preview. This is probably the AI-enabled counterpart - I see value in both approaches and as stated would love to see this progress.

NedMalki-Chief:
That's really cool. Would love to see it when it's ready for preview.

I think you're right that they're natural counterparts. Mine is AI-driven (Claude actively manages the planning state), yours is human-driven (read-only view of the artifacts). Different workflows, same underlying files.

One thought: if we both converge on `.planning/{project}/` as the multi-project directory convention, our tools would compose naturally without either needing to know about the other. Your TUI could scan that structure the same way the env var targets it. Might be worth keeping in sync on that as both progress.

Happy to share more about how the directory layout works in practice if that's useful for your TUI design.

trek-e (approved):
## Adversarial Review -- PR #1484

### Summary
Adds `GSD_PROJECT` environment variable to `planningDir()` for project-scoped planning directory resolution. When set, routes to `.planning/{project}/` instead of `.planning/`. Composable with `GSD_WORKSTREAM`. 22 additions, 8 deletions in a single file (`core.cjs`).

### Vision Alignment Analysis

**Verdict: Well-aligned.**

1. **Follows the existing pattern exactly.** `GSD_PROJECT` is implemented identically to `GSD_WORKSTREAM` -- optional env var, falls back to null, zero behavior change when unset. The author correctly identified the pattern and extended it.

2. **Solo developer focused.** The use case (Obsidian vault as a command center for multiple independent projects) is a real solo developer workflow. One terminal, multiple projects, switch via env var.

3. **File-based state.** No database, no server. Just `.planning/{project}/` directories.

4. **Absent = enabled default.** When `GSD_PROJECT` is unset, `planningDir()` returns `.planning/` exactly as before. Zero regression.

5. **No external dependencies.**

6. **Thin change.** 22 additions to one function in one file. `loadConfig()` updated to use `planningDir()` instead of hardcoded path -- this was a latent bug where config loading ignored workstreams, now fixed.

### Correctness Assessment

- `planningDir(cwd, ws, project)` resolution order: project first, then workstream. Composing both produces `.planning/{project}/workstreams/{ws}/`. Correct.
- Backward compatible: no change when both env vars are unset.
- `planningRoot()` JSDoc updated to note it ignores both projects and workstreams. Correct -- shared resources should always be at the root.
- `loadConfig()` now calls `planningDir(cwd)` instead of `path.join(cwd, '.planning', 'config.json')`. This means config loading is now project-aware and workstream-aware. Correct fix for what was previously a bug (config was always read from root even when a workstream was active).

### Bugs / Edge Cases

- **Path traversal:** If `GSD_PROJECT=../../etc`, `planningDir()` would return `.planning/../../etc/`. The `path.join` call resolves this to a path outside `.planning/`. There is no validation that the project name is a safe directory segment. Should validate that project name contains no path separators or `..` components. Low risk since the env var is set by the user themselves, but defense in depth suggests sanitizing.
- **Missing tests.** The PR description includes a test plan but the diff shows no test file changes. The existing test suite presumably still passes (backward compatible), but there are no tests verifying the new `GSD_PROJECT` behavior. Given the change is 14 lines of logic in a single function, manual testing is acceptable, but adding 3-4 unit tests for `planningDir()` with project/workstream/both/neither would be prudent.

### Security

- The path traversal concern above is the only surface. Since the env var is user-controlled (not from untrusted input), this is low severity.

### Verdict: APPROVE

Minimal, well-patterned change that extends an existing mechanism. The `loadConfig()` fix is a bonus -- it corrects a latent bug where workstream-scoped configs were ignored. Consider adding path segment validation and a few unit tests in a follow-up.

trek-e:
Please address the path traversal  and missing tests edge cases.

NedMalki-Chief:
Thanks for the thorough review. Both addressed:

**Path traversal:** Added validation in `planningDir()` that rejects project and workstream names containing path separators (`/`, `\`) or traversal components (`..`). Throws a clear error message. Applied the same check to `GSD_WORKSTREAM` as well since the same vector exists there.

**Tests:** Added 9 unit tests covering the full resolution matrix: project only, workstream only, both, neither, env var fallback, and 4 traversal rejection cases (forward slash, backslash, `..` in project, `..` in workstream).

All 136 tests pass.

trek-e:
## Re-Review — PR #1484

@NedMalki-Chief — all three previously requested changes are confirmed addressed in the diff:

**Path traversal:** The \`BAD_SEGMENT = /[/\\]|\.\./\` check is present in \`planningDir()\` and throws a clear error for both \`GSD_PROJECT\` and \`GSD_WORKSTREAM\`. The validation is applied to both env vars symmetrically — correct.

**Tests:** 9 unit tests added to \`tests/core.test.cjs\` covering the full resolution matrix (project-only, workstream-only, both, neither, env var fallback) plus 4 traversal rejection cases (forward slash, backslash, \`..)\` in project, \`..)\` in workstream). Coverage is complete.

**\`loadConfig()\` fix:** Confirmed — now calls \`planningDir(cwd)\` instead of the hardcoded \`.planning\` path, making config loading project- and workstream-aware.

CI has not run yet (pending maintainer approval for the fork). The diff itself is clean — the logic is correct, the tests are thorough, and the change is minimal and well-patterned.

Removing the "changes requested" label and approving.

abrauchli:
@NedMalki-Chief https://github.com/bitcreed/gsd-meta-manager

NedMalki-Chief:
Nice, you shipped it. The 9-tab detail view and the HANDOFF parser are exactly what I was hoping a read-only counterpart would look like. I'll register a couple of my projects against it this week and send notes if anything in the `.planning/{project}/` layout trips it up. Good to see us converging on the same directory convention without either tool needing to know about the other.

PR #1492, opened 2026-03-30
feat: detect and prevent silent scope reduction in planner

grgbrasil:
## Problem

When phases have many user decisions (from `/gsd:discuss-phase`), the planner sometimes silently simplifies them to fit within the plan budget. For example:

- **User decided (D-26):** "Config displays cost references calculated in impulses from the pricing table"
- **Planner delivered:** "D-26 cost references (v1 — static labels). NOT wired to billing. Dynamic pricing is a future enhancement."

The planner invented a "v1/v2" versioning that doesn't exist in the user's decision. The executor then faithfully implemented the simplified plan — building something the user never asked for.

This is the most insidious failure mode because:
- Plans reference D-XX (passes Dimension 7 context compliance check)
- Tasks mention the decision ID (looks traceable)
- But the implementation is a shadow of the actual requirement

## Root Cause

The planner's instinct when facing a complex phase is to **simplify individual decisions** ("make it static for v1"). The correct behavior should be to **split the phase** so every decision is implemented at full fidelity, just across smaller phases.

### 1. `agents/gsd-planner.md` — `<scope_reduction_prohibition>`

- Explicitly prohibits language like "v1", "static for now", "hardcoded", "future enhancement", "placeholder"
- Requires a **decision coverage matrix** mapping every D-XX to plan/task with Full/Partial status
- If any decision would be Partial: return `## PHASE SPLIT RECOMMENDED` instead of simplifying
- The user spent time making decisions in discuss-phase — silently reducing them wastes that time

### 2. `agents/gsd-plan-checker.md` — Dimension 7b: Scope Reduction Detection

- Scans all task actions for scope reduction language patterns
- Cross-references each match with the CONTEXT.md decision it claims to implement
- Compares: does the task deliver what D-XX **actually says**, or a reduced version?
- **Always BLOCKER severity** — scope reduction is never acceptable as a warning
- Includes a real-world example from the production incident that motivated this change

### 3. `get-shit-done/workflows/plan-phase.md` — Step 9b: Phase Split Handling

New orchestrator flow when planner returns `## PHASE SPLIT RECOMMENDED`:

- **Option 1: Split** — Create sub-phases via `/gsd:insert-phase`, each with a subset of D-XX decisions
- **Option 2: Proceed** — Attempt all decisions (accepting more plans/tasks, potential quality degradation)
- **Option 3: Prioritize** — User picks which D-XX are "now" vs "later"

## Real-world incident

Phase 17 (Media Processing) had 30 decisions including detailed billing requirements (D-20 to D-27). The planner:
1. Reduced D-26 ("calculated costs in impulses") to "static labels /min"
2. Reduced D-23 ("everything well explained for admin") to a single generic sentence
3. The executor then reimplemented an entire extraction service from scratch instead of reusing the existing one (D-03)

Result: 4 waves of execution building the wrong thing. The RESEARCH.md even had the correct pattern (dynamic costs from billing API) but the planner ignored it.

With these changes, the planner would have been blocked at plan creation, and the checker would have caught it even if the planner slipped through.

## Test plan

- [ ] Run `/gsd:plan-phase` on a phase with 15+ decisions — verify decision coverage matrix is produced
- [ ] Artificially add "v1 static" language to a plan — verify Dimension 7b flags BLOCKER
- [ ] Test PHASE SPLIT RECOMMENDED flow — verify 3 options presented to user
- [ ] Verify no regression on phases with few decisions (normal flow unchanged)

### DIFF
--- a/agents/gsd-plan-checker.md
+++ b/agents/gsd-plan-checker.md
@@ -314,6 +314,49 @@ issue:
   fix_hint: "Remove search task - belongs in future phase per user decision"
 ```
 
+## Dimension 7b: Scope Reduction Detection
+
+**Question:** Did the planner silently simplify user decisions instead of delivering them fully?
+
+**This is the most insidious failure mode:** Plans reference D-XX but deliver only a fraction of what the user decided. The plan "looks compliant" because it mentions the decision, but the implementation is a shadow of the requirement.
+
+**Process:**
+1. For each task action in all plans, scan for scope reduction language:
+   - `"v1"`, `"v2"`, `"simplified"`, `"static for now"`, `"hardcoded"`
+   - `"future enhancement"`, `"placeholder"`, `"basic version"`, `"minimal"`
+   - `"will be wired later"`, `"dynamic in future"`, `"skip for now"`
+   - `"not wired to"`, `"not connected to"`, `"stub"`
+2. For each match, cross-reference with the CONTEXT.md decision it claims to implement
+3. Compare: does the task deliver what D-XX actually says, or a reduced version?
+4. If reduced: BLOCKER — the planner must either deliver fully or propose phase split
+
+**Red flags (from real incident):**
+- CONTEXT.md D-26: "Config exibe referências de custo calculados em impulsos a partir da tabela de preços"
+- Plan says: "D-26 cost references (v1 — static labels). NOT wired to billingPrecosOriginaisModel — dynamic pricing display is a future enhancement"
+- This is a BLOCKER: the planner invented "v1/v2" versioning that doesn't exist in the user's decision
+
+**Severity:** ALWAYS BLOCKER. Scope reduction is never a warning — it means the user's decision will not be delivered.
+
+**Example:**
+```yaml
+issue:
+  dimension: scope_reduction
+  severity: blocker
+  description: "Plan reduces D-26 from 'calculated costs in impulses' to 'static hardcoded labels'"
+  plan: "03"
+  task: 1
+  decision: "D-26: Config exibe referências de custo calculados em impulsos"
+  plan_action: "static labels v1 — NOT wired to billing"
+  fix_hint: "Either implement D-26 fully (fetch from billingPrecosOriginaisModel) or return PHASE SPLIT RECOMMENDED"
+```
+
+**Fix path:** When scope reduction is detected, the checker returns ISSUES FOUND with recommendation:
+```
+Plans reduce {N} user decisions. Options:
+1. Revise plans to deliver decisions fully (may increase plan count)
+2. Split phase: [suggested grouping of D-XX into sub-phases]
+```
+
 ## Dimension 8: Nyquist Compliance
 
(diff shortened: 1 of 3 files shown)


trek-e (approved):
## Adversarial Review -- PR #1492

### Summary
Adds scope reduction detection to the planner and plan-checker. Three coordinated changes: (1) `<scope_reduction_prohibition>` in gsd-planner.md with decision coverage matrix requirement, (2) Dimension 7b in gsd-plan-checker.md for scope reduction scanning, (3) Phase split handling in plan-phase.md. 115 additions, 0 deletions across 3 files.

### Vision Alignment Analysis

**Verdict: Strongly aligned.**

1. **Defense in depth.** This is textbook GSD philosophy -- the planner has a prohibition, the checker has a detection dimension, and the orchestrator has a recovery path. Three layers for one failure mode.

2. **Complexity in the system, not the workflow.** The user sees one new interaction: a phase split prompt with 3 options (split, proceed, prioritize). This surfaces only when the planner detects it cannot deliver all decisions at full fidelity -- a situation where user input is genuinely needed.

3. **No external dependencies.** Pure prompt engineering changes to agent instructions and one new orchestrator step.

4. **Protects user time.** The PR description documents a real production incident where a planner silently reduced decisions and 4 waves built the wrong thing. This feature prevents that waste.

5. **Solo developer focused.** A solo developer's discuss-phase decisions are the spec. Silently reducing them is the worst kind of failure because it looks correct. This catches it.

### Correctness Assessment

- The scope reduction language patterns are well-chosen: "v1", "static for now", "hardcoded", "future enhancement", "placeholder", "basic version", "minimal", etc. These are the exact phrases LLMs use when they want to simplify.
- The decision coverage matrix (D-XX -> Plan -> Task -> Full/Partial) is a forcing function that makes scope reduction visible during planning.
- Dimension 7b in the checker cross-references task actions against CONTEXT.md decisions. Always BLOCKER severity -- no warnings for scope reduction. Correct.
- The phase split handler presents three options via AskUserQuestion: Split (via insert-phase), Proceed (attempt all), Prioritize (user picks). All three are legitimate recovery paths.
- The planner returns `## PHASE SPLIT RECOMMENDED` -- this integrates cleanly with the existing return value parsing in plan-phase.md step 9.

### Bugs / Edge Cases

- **False positives:** The word "minimal" in a task like "Add minimal error handling for edge case X" could trigger a false positive. However, since Dimension 7b cross-references against the actual CONTEXT.md decision, and the decision presumably does not say "minimal error handling," the cross-reference step should filter it out.
- **Non-English decisions:** The real-world example uses Portuguese ("Config exibe referencias de custo calculados em impulsos"). The scope reduction patterns are English-only. If a planner generates task actions in the same language as CONTEXT.md decisions, the English patterns may not trigger. Low risk since planners typically write actions in English.
- **No tests.** This is prompt-engineering (Markdown workflow files), not code. The test plan is manual. Acceptable for workflow-level changes.

### Security

- No executable code changes. The scope reduction patterns are string matching performed by the plan-checker agent, not by runtime code. No injection surface.

### Verdict: APPROVE

This addresses a real, documented failure mode (silent scope reduction) with a well-designed three-layer defense. The implementation is purely additive, the user interaction is minimal and genuinely useful, and the recovery paths are complete. The decision coverage matrix is a particularly good forcing function.

PR #1500, opened 2026-03-30
fix(install): handle JSONC (comments) in settings.json without data loss

### LINKED ISSUE #1461: settings.json overwritten (lost) when settings.json contains comments

Tibsfox:
## Summary

**Critical data loss bug** — when `settings.json` contains comments (`//` or `/* */`), which many CLI tools allow (Qwen, VS Code, etc.), `JSON.parse()` fails and `readSettings()` silently returned `{}`. This empty object was then written back by `writeSettings()`, destroying the user's entire configuration.

### Root cause
```javascript
// Before: silent data loss
function readSettings(settingsPath) {
  try {
    return JSON.parse(fs.readFileSync(settingsPath, 'utf8'));
  } catch (e) {
    return {};  // ← comments cause parse failure, returns empty, gets written back
  }
}
```

### Fix
- Added `stripJsonComments()` that handles line comments (`//`), block comments (`/* */`), trailing commas, and preserves comments inside string values
- `readSettings()` tries standard `JSON.parse` first (fast path), falls back to JSONC stripping
- On truly malformed files (even JSONC stripping fails), returns `null` with a warning instead of `{}` — callers check for null and skip settings modification rather than overwriting
- All callers (`getCommitAttribution`, uninstall cleanup, hook configuration) now guard against null return

### What this handles
- Line comments: `// My configuration`
- Block comments: `/* match all */`
- Inline comments: `"timeout": 5000, // milliseconds`
- Trailing commas: `{ "a": 1, "b": 2, }`
- Comments inside strings preserved: `"url": "https://example.com"` (not stripped)

## Test plan
- [x] 12 new tests in `tests/settings-jsonc.test.cjs`:
  - 9 `stripJsonComments` unit tests (line, block, multi-line, inline, string preservation, trailing commas, empty, standard JSON, real-world settings)
  - 3 integration tests (install.js uses stripJsonComments, returns null not {}, callers guard null)
- [x] Full test suite passes

Closes #1461

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -364,14 +364,78 @@ function resolveOpencodeConfigPath(configDir) {
 }
 
 /**
- * Read and parse settings.json, returning empty object if it doesn't exist
+ * Strip JSONC comments (// and /* *​/) from a string to produce valid JSON.
+ * Handles comments inside strings correctly (does not strip them).
+ */
+function stripJsonComments(text) {
+  let result = '';
+  let i = 0;
+  let inString = false;
+  let stringChar = '';
+  while (i < text.length) {
+    // Handle string literals — don't strip comments inside strings
+    if (inString) {
+      if (text[i] === '\\') {
+        result += text[i] + (text[i + 1] || '');
+        i += 2;
+        continue;
+      }
+      if (text[i] === stringChar) {
+        inString = false;
+      }
+      result += text[i];
+      i++;
+      continue;
+    }
+    // Start of string
+    if (text[i] === '"' || text[i] === "'") {
+      inString = true;
+      stringChar = text[i];
+      result += text[i];
+      i++;
+      continue;
+    }
+    // Line comment
+    if (text[i] === '/' && text[i + 1] === '/') {
+      // Skip to end of line
+      while (i < text.length && text[i] !== '\n') i++;
+      continue;
+    }
+    // Block comment
+    if (text[i] === '/' && text[i + 1] === '*') {
+      i += 2;
+      while (i < text.length && !(text[i] === '*' && text[i + 1] === '/')) i++;
+      i += 2; // skip closing */
+      continue;
+    }
+    result += text[i];
+    i++;
+  }
+  // Remove trailing commas before } or ] (common in JSONC)
+  return result.replace(/,\s*([}\]])/g, '$1');
+}
+
+/**
+ * Read and parse settings.json, returning empty object if it doesn't exist.
+ * Supports JSONC (JSON with comments) — many CLI tools allow comments in
+ * their settings files, so we strip them before parsing to avoid silent
+ * data loss from JSON.parse failures.
  */
 function readSettings(settingsPath) {
   if (fs.existsSync(settingsPath)) {
     try {
-      return JSON.parse(fs.readFileSync(settingsPath, 'utf8'));
+      const raw = fs.readFileSync(settingsPath, 'utf8');
+      // Try standard JSON first (fast path)
+      try {
+        return JSON.parse(raw);
+      } catch {
+        // Fall back to JSONC stripping
+        return JSON.parse(stripJsonComments(raw));
+      }
     } catch (e) {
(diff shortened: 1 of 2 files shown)


trek-e (approved):
## Adversarial Review: #1500 fix(install): handle JSONC (comments) in settings.json without data loss

**Summary:** Fixes a critical data loss bug where `readSettings()` silently returned `{}` when settings.json contained JSONC comments, causing `writeSettings()` to overwrite the entire config. Adds `stripJsonComments()` with a fast-path fallback and null-return on truly malformed files. All callers now guard against null. Fixes #1461.

**Correctness:** The `stripJsonComments()` implementation is a well-structured character-by-character parser that correctly handles:
- String context tracking (does not strip comments inside string values)
- Escape character handling (`\\` inside strings)
- Both line (`//`) and block (`/* */`) comments
- Trailing comma removal via regex
- Single-quote strings (defensive, though not standard JSON)

The fast-path approach (try `JSON.parse` first, fall back to JSONC stripping) avoids performance regression for valid JSON files.

**Critical improvement:** Returning `null` instead of `{}` on parse failure is the correct fix for the data loss vector. All three call sites (`getCommitAttribution` for opencode/gemini/claude, `uninstall`, and `install`) now guard against null.

**Edge cases verified:**
- Unterminated block comment: the `while` loop exits when `i >= text.length`, `i += 2` could overshoot past end of buffer but `text[i]` would return `undefined` and the while condition handles it. No crash.
- Empty file: returns `{}` via `JSON.parse(stripJsonComments(''))` which would fail, falling to null. Actually -- empty file would return `''` from readFileSync, `JSON.parse('')` throws, `stripJsonComments('')` returns `''`, second `JSON.parse('')` also throws, returns null. This is correct behavior (empty file is not valid JSON).
- File that exists but is just whitespace/comments: strips to empty string, `JSON.parse` fails, returns null. Correct -- preserves the file rather than overwriting.

**Tests:** 12 tests covering unit and integration scenarios. The test file duplicates the `stripJsonComments` implementation rather than importing it, which is a reasonable pattern for testing install.js (which doesn't export the function).

**Security:** No concerns. The JSONC parser only strips content, never adds or transforms.

**Verdict: APPROVE** -- high-quality fix for a genuine data loss bug with appropriate defensive handling.

PR #1560, opened 2026-04-02
[codex] isolate active workstream state per session

LakshmanTurlapati:
## What changed

This fixes workstream routing so concurrent Claude/Codex sessions can work on different milestones without silently repointing each other's planning context.

The core change is to stop relying on a single shared `.planning/active-workstream` file whenever the runtime exposes a stable session identity. Instead, GSD now stores the active workstream in a session-scoped pointer under temp storage, keyed from runtime/session signals such as `GSD_SESSION_KEY`, `CODEX_THREAD_ID`, Claude session identifiers, terminal session identifiers, or the controlling TTY.

## Why this changed

The existing shared pointer was last-writer-wins. If one AI session switched to another workstream, another active session on the same repo could suddenly start reading and writing the wrong `STATE.md`, `ROADMAP.md`, and phase paths.

That made parallel milestone work unsafe even though workstreams are meant to support it.

## Impact

Concurrent sessions now keep their active workstream isolated from each other by default when the runtime provides a stable session key.

Legacy behavior is preserved as a fallback for environments that do not expose any session identity: in that case GSD still uses `.planning/active-workstream`.

The change also documents the new resolution order and adds regression coverage for concurrent-session behavior.

## Root cause

A single shared `.planning/active-workstream` pointer was being used as implicit global state across multiple terminals and AI runtimes in the same repository.

## Validation

- `node --test tests/workstream.test.cjs`
- `npm test`

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -22,9 +22,10 @@ const WORKSTREAM_SESSION_ENV_KEYS = [
   'WT_SESSION',
   'TMUX_PANE',
   'ZELLIJ_SESSION_NAME',
-  'TTY',
-  'SSH_TTY',
 ];
+
+let cachedControllingTtyToken = null;
+let didProbeControllingTtyToken = false;
 
 // ─── Path helpers ────────────────────────────────────────────────────────────
 
@@ -637,10 +638,14 @@ function sanitizeWorkstreamSessionToken(value) {
   return token ? token.slice(0, 160) : null;
 }
 
-function getControllingTtyToken() {
-  for (const envKey of ['TTY', 'SSH_TTY']) {
-    const token = sanitizeWorkstreamSessionToken(process.env[envKey]);
-    if (token) return `tty-${token.replace(/^dev_/, '')}`;
+function probeControllingTtyToken() {
+  if (didProbeControllingTtyToken) return cachedControllingTtyToken;
+  didProbeControllingTtyToken = true;
+
+  // `tty` reads stdin. When stdin is already non-interactive, spawning it only
+  // adds avoidable failures on the routing hot path and cannot reveal a stable token.
+  if (!(process.stdin && process.stdin.isTTY)) {
+    return cachedControllingTtyToken;
   }
 
   try {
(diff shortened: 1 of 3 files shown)


LakshmanTurlapati:
Addressed the requested changes in `caec78e`.

Response to the review points:

- Rebase overlap with `#1561`: this branch has been rebased on top of the merged `main` changes and force-pushed, so the PR head now includes the concurrency-safety work.
- `execFileSync('tty')` hot-path concern: `TTY` / `SSH_TTY` are now handled in the dedicated controlling-terminal path instead of the generic session-env list, the direct `tty` probe is cached, and it is skipped entirely when `stdin` is non-interactive.
- Tmpdir cleanup behavior: cleanup is now centralized through `clearActiveWorkstreamPointer()`, and the shared per-project temp directory is only removed when it is empty. I also documented the best-effort cleanup behavior in `get-shit-done/references/workstream-flag.md`.
- Deleted workstream pointers: stale pointers now self-clean through the same helper path, and the code/docs call out that this resolves to `null` rather than leaving unexplained stale state behind.
- TTY fallback coverage / commentary: added black-box regressions for headless no-session-key fallback, explicit-session-vs-TTY precedence, direct `TTY` env handling, sibling pointer preservation, stale-pointer cleanup, and last-pointer temp-dir removal. I also added helper-level comments describing the fallback order and cleanup semantics.

Validation run on the updated branch:

- `node --test tests/workstream.test.cjs tests/concurrency-safety.test.cjs`

That completed green locally (`105/105`).

trek-e:
## ⚠️ Path Canonicalization Bug Origin

This PR introduced a latent Windows path canonicalization bug in `getSessionScopedWorkstreamFile` (commit `5ce8183`).

**Root cause:** `path.resolve()` preserves caller-supplied path casing, while `fs.realpathSync.native()` returns the OS-canonical path. On Windows, these differ for the same physical path:
- `path.resolve('/users/TREKKIE/.planning')` → `C:\Users\TREKKIE\.planning`  
- `fs.realpathSync.native(...)` → `C:\Users\trekkie\.planning`

The SHA-1 hash of the planning directory path was computed with `path.resolve` in the implementation, but `fs.realpathSync.native` in the test helper — producing different hashes and different session tmpdir slots. Tests checked a directory the implementation never created.

**Fixed by:** PR #1601 — changed `path.resolve` to `fs.realpathSync.native` with `path.resolve` fallback.

**Pattern to avoid:** When hashing a filesystem path as a stable identifier, always use `fs.realpathSync.native()` (with `path.resolve` fallback for paths that don't exist yet). Never mix `path.resolve` in production code with `fs.realpathSync.native` in tests for the same path — they produce different hashes on Windows.

PR #1574, opened 2026-04-03
fix: add AskUserQuestion to review-backlog allowed-tools

xiaolai:
## What

Add `AskUserQuestion` to the `allowed-tools` list in `commands/gsd/review-backlog.md`.

## Why

Step 3 of the review-backlog command instructs the agent to present each backlog item to the user via `AskUserQuestion`, but the tool is not listed in `allowed-tools`. This causes a runtime failure — the command cannot prompt the user for decisions on individual backlog items.

Found during an NLPM audit of all 80 NL artifacts in the repo.

## How

Added `AskUserQuestion` to the existing `allowed-tools` YAML list, matching the pattern used by other interactive commands like `debug.md`.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Codex
- [ ] Copilot

### Test details

Confirmed that step 3 references `AskUserQuestion` and that other commands using the same tool (e.g., `debug.md`) include it in `allowed-tools`.

## Checklist

## Breaking Changes

None

### DIFF
--- a/commands/gsd/review-backlog.md
+++ b/commands/gsd/review-backlog.md
@@ -5,6 +5,7 @@ allowed-tools:
   - Read
   - Write
   - Bash
+  - AskUserQuestion
 ---
 
 <objective>


trek-e:
Thanks for the audit and the fix — the diagnosis is correct. However, commit `08f59fa` in PR #1590 landed all five NLPM audit fixes simultaneously before this PR could be merged, making this a no-op against current `main`. Closing to keep the queue clean. The underlying issue (#1578) is resolved.

xiaolai:
Thanks for the review and for getting this merged via `08f59fa`. Happy to close — the fix is already live on `main`.

PR #1576, opened 2026-04-03
fix: add missing allowed-tools to 4 commands

xiaolai:
## What

Add `allowed-tools` field to 4 command files that were missing it entirely.

## Why

Commands without `allowed-tools` have no tool access constraints — the runtime grants unrestricted tool access. This violates least-privilege. All other commands in `commands/gsd/` declare explicit tool sets.

Found during an NLPM audit of all 80 NL artifacts in the repo.

## How

Added `allowed-tools` with the minimum tool set each command actually needs:

| Command | Tools | Rationale |
|---------|-------|-----------|
| `cleanup.md` | Read, Write, Bash, Glob | Archives directories, reads/writes files |
| `help.md` | `[]` (empty) | Pure-output command, displays static reference |
| `join-discord.md` | `[]` (empty) | Pure-output command, displays invite link |
| `workstreams.md` | Read, Bash | Reads state files, runs `node gsd-tools.cjs` |

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Codex
- [ ] Copilot

### Test details

Audited all 59 command files. These 4 were the only ones missing `allowed-tools`. Tool sets were determined by reading each command's process steps and identifying which tools are actually invoked.

## Checklist

## Breaking Changes

None — commands that previously had unrestricted access now have explicit constraints. If a command needs a tool not listed, it can be added later.

### DIFF
--- a/commands/gsd/cleanup.md
+++ b/commands/gsd/cleanup.md
@@ -1,6 +1,11 @@
 ---
 name: gsd:cleanup
 description: Archive accumulated phase directories from completed milestones
+allowed-tools:
+  - Read
+  - Write
+  - Bash
+  - Glob
 ---
 <objective>
 Archive phase directories from completed milestones into `.planning/milestones/v{X.Y}-phases/`.
--- a/commands/gsd/help.md
+++ b/commands/gsd/help.md
@@ -1,6 +1,7 @@
 ---
 name: gsd:help
 description: Show available GSD commands and usage guide
+allowed-tools: []
 ---
 <objective>
 Display the complete GSD command reference.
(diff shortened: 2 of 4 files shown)


trek-e:
Thanks for the audit and the fix — the diagnosis is correct. However, commit `08f59fa` in PR #1590 landed all five NLPM audit fixes simultaneously before this PR could be merged, making this a no-op against current `main`. Closing to keep the queue clean. The underlying issue (#1578) is resolved.

xiaolai:
Thanks for the detailed review. Good catch that the three other files are already fixed on `main`.

Per your recommendation, I've opened #1638 to tighten `workstreams.md` allowed-tools from `[Read, Write, Bash]` to `[Read, Bash]` — since the command only delegates to `node "$GSD_TOOLS"` via Bash and never writes files. That should bring it in line with least-privilege.

PR #1595, opened 2026-04-03
fix: only remove session tmp dir when last pointer file is cleared (#1593)

### LINKED ISSUE #1593: bug: pointer lifecycle hardening tests fail on Windows (session tmp dir removed prematurely)
## Description

Three tests in `pointer lifecycle hardening` (added in #1560) fail on Windows/Node 22 while passing on Ubuntu and macOS:

1. `clearing one session pointer leaves sibling session pointers intact`
2. `stale pointers self-clean without deleting sibling session pointers`
3. `clearing the last session pointer removes the empty session tmp directory`

All three assert on the existence/non-existence of the session tmp directory under `os.tmpdir()/gsd-workstream-sessions/<projectId>/`.

## Observed behavior

On Windows, when clearing one session's pointer file, the session tmp directory is removed even though a sibling pointer file still exists. The `assert.ok(fs.existsSync(sessionDir))` assertion fails (expected true, actual false).

## Likely root cause

The workstream `--clear` command's cleanup logic (introduced in #1560) deletes the `<projectId>` directory after removing the pointer file, without checking whether other pointer files remain in that directory. On Linux/macOS this may work due to filesystem ordering differences, but on Windows the deletion is unconditional.

## Steps to reproduce

Run `node --test tests/workstream.test.cjs` on Windows with Node 22.

## Impact

Blocks CI for any PR rebased on the commit that introduced this (post #1560 merge). PR #1591 is currently blocked by this pre-existing failure.

## Location

`tests/workstream.test.cjs:252-317`
Implementation: likely `get-shit-done/bin/lib/workstream.cjs` or equivalent session pointer cleanup code.

trek-e:
Closes #1593

## What
After removing a session pointer file, only remove the parent session tmp directory if it is empty (no sibling pointer files remain).

## Why
The previous code relied on `rmdirSync` throwing `ENOTEMPTY` to prevent deletion of a non-empty directory, but on Windows that error is not reliably raised, causing the shared session tmp directory to be deleted while sibling session pointers still exist.

## How
In `clearActiveWorkstreamPointer`, call `fs.readdirSync` on the directory first and only call `rmdirSync` when the result is empty.

## Testing
- [ ] macOS
- [ ] Linux

Test details: All three `pointer lifecycle hardening` tests now pass on local run; root cause (unconditional directory deletion) fixed.

## Checklist
- [x] Issue linked above (`Closes #1593`)
- [x] Follows GSD style

## Breaking Changes
None

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -1128,9 +1128,15 @@ function getRoadmapPhaseInternal(cwd, phaseNum) {
  * gsd-tools.cjs lives at <configDir>/get-shit-done/bin/gsd-tools.cjs,
  * so agents/ is at <configDir>/agents/.
  *
+ * GSD_AGENTS_DIR env var overrides the default path. Used in tests and for
+ * installs where the agents directory is not co-located with gsd-tools.cjs.
+ *
  * @returns {string} Absolute path to the agents directory
  */
 function getAgentsDir() {
+  if (process.env.GSD_AGENTS_DIR) {
+    return process.env.GSD_AGENTS_DIR;
+  }
   // __dirname is get-shit-done/bin/lib/ → go up 3 levels to configDir
   return path.join(__dirname, '..', '..', '..', 'agents');
 }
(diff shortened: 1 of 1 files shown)


trek-e:
Closing — this branch accidentally included changes belonging to #1594 and #1596 while missing the actual pointer lifecycle fix. Replacing with a clean PR.

Issue #1656, opened 2026-04-04
bug: 3 bash hooks referenced in settings.json but never installed


trek-e:
## Summary

The GSD installer adds hook entries to `~/.claude/settings.json` referencing three bash scripts that are never actually installed to `~/.claude/hooks/`:

- `gsd-session-state.sh` (SessionStart)
- `gsd-validate-commit.sh` (PreToolUse on Bash)
- `gsd-phase-boundary.sh` (PostToolUse on Write|Edit)

## Evidence

After a clean install of GSD 1.32.0, `~/.claude/hooks/` contains:

```
autoresearch.js
gsd-check-update.js
gsd-context-monitor.js
gsd-prompt-guard.js
gsd-read-guard.js
gsd-research-cache.js
gsd-statusline.js
gsd-workflow-guard.js
rtk-rewrite.sh          (from RTK, not GSD)
```

No `.sh` files from GSD are present. Meanwhile `settings.json` hooks reference:

```json
{"command": "bash /Users/.../.claude/hooks/gsd-session-state.sh"}
{"command": "bash /Users/.../.claude/hooks/gsd-validate-commit.sh"}
{"command": "bash /Users/.../.claude/hooks/gsd-phase-boundary.sh"}
```

## Impact

Every matching hook event fires `bash /path/to/missing.sh` which exits non-zero. Depending on how Claude Code handles hook failures, this can produce visible error messages or silently degrade behavior on every Bash command (validate-commit), every Write/Edit (phase-boundary), and every session start (session-state).

## Expected behavior

Either:
1. The installer should install the referenced `.sh` files, or
2. The installer should not add these entries to `settings.json` if the scripts don't ship

## Workaround

Manually remove the three stale entries from `~/.claude/settings.json`.

## Environment

- GSD version: 1.32.0
- Platform: macOS (Darwin 25.4.0)
- Runtime: claude
- Install scope: global

Issue #1755, opened 2026-04-05
v1.32.0 reinstall doesn't copy .sh hooks, causing PreToolUse:Bash hook errors


richard-agentic:
## Bug
After reinstalling GSD v1.32.0, three shell hook files are missing from `~/.claude/hooks/`, but `settings.json` still references them. This causes `PreToolUse:Bash hook error` on every `Bash`, `Write`, and `Edit` tool use.

## Missing files
- `gsd-phase-boundary.sh`
- `gsd-validate-commit.sh`
- `gsd-session-state.sh`

## Reproduction
1. Install GSD v1.32.0 on a system that already had an older version (so `settings.json` contains hook references).
2. Run `gsd-update` or reinstall.
3. Try running any `/gsd:*` command or any `Bash` tool call in Claude Code.
4. Observe `PreToolUse:Bash hook error`.

## Root cause
The installer appears to restore/merge `settings.json` hook entries, but does **not** copy the corresponding `.sh` files into `~/.claude/hooks/`. The JS hooks (`gsd-context-monitor.js`, etc.) are present; only the shell hooks are absent.

## Workaround
Manually download the missing `.sh` files from the repo into `~/.claude/hooks/` and make them executable.

## Suggested fix
Ensure the installer copies `hooks/*.sh` into `~/.claude/hooks/` during upgrade/reinstall, or skips adding stale references if the files aren't being installed.

trek-e:
Confirmed — this is the same root cause as #1754.

**Root cause:** In v1.32.0, `scripts/build-hooks.js` only copies `.js` hooks to `hooks/dist/`. The three `.sh` community hooks (`gsd-validate-commit.sh`, `gsd-session-state.sh`, `gsd-phase-boundary.sh`) were never added to the `HOOKS_TO_COPY` array. Since `hooks/dist/` is gitignored, the npm package ships without them. But the installer still registers them in `settings.json` → file-not-found → "hook error" on every Bash/Write/Edit tool call.

**Fixed in:** commit 8d732ee (part of v1.33.0 on `main`). The fix adds the three `.sh` files to `HOOKS_TO_COPY` with chmod preservation.

**Not yet on npm** — v1.33.0 publish is blocked by #1752.

**Workaround:** Same as #1754 — either install from GitHub (`npx github:gsd-build/get-shit-done`) or remove the stale hook entries from `~/.claude/settings.json`.

Marking as duplicate of the underlying cause (fixed on main, pending npm publish).

Issue #1754, opened 2026-04-05
PreToolUse:Bash hook error - The GSD workflow reference files aren't installed


rpgdev:
### GSD Version

1.33.0

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

v22.22.1

### Shell

bin/zsh

### Installation Method

npx get-shit-done-cc@latest (fresh run)

### What happened?

I tried to create a new milestone off of a TODO and it kept giving me this error.                                      

### What did you expect?

Smooth running of the skill.

### Steps to reproduce

1. install latest GSD
2. run `gsd-new-milestone` skill (I miss the colon separator there btw)

### Error output / logs

```shell
❯ /gsd-new-milestone based on todo 3                                                                                                                         
                                                                                                                                                             
  Searched for 6 patterns, read 2 files, listed 1 directory (ctrl+o to expand)                                                                               
  ⎿  PreToolUse:Bash hook error                                                                                                                              
  ⎿  PreToolUse:Bash hook error                                                                                      
  ⎿  PreToolUse:Bash hook error                                                                                                                              
  ⎿  PreToolUse:Bash hook error                                                                                                                              
                                                                                                                                                             
● The GSD workflow reference files aren't installed, but I have the skill objective and the todo context. Let me proceed with the new-milestone workflow     
  manually. First, let me load the current project state.                                                                                                    
                                                                                                                                                             
  Read 3 files (ctrl+o to expand)
```

### GSD Configuration

```json
{
  "mode": "yolo",
  "parallelization": true,
  "commit_docs": true,
  "model_profile": "balanced",
  "workflow": {
    "research": true,
    "plan_check": true,
    "verifier": true,
    "auto_advance": false,
    "nyquist_validation": true,
    "_auto_chain_active": true
  },
  "granularity": "standard",
  "git": {
    "branching_strategy": "none"
  }
}%
```

### GSD State (if relevant)

```markdown

```

### Runtime settings.json (if relevant)

```json

```

### How often does this happen?

Every time (100% reproducible)

### Impact

Major — Core feature is broken, no workaround

### Workaround (if any)

_No response_

### Additional context

The move to skills seems to have broken gsd a little bit.

### Privacy Checklist

- [x] I have reviewed all pasted output for PII (usernames, paths, API keys) and redacted where necessary

trek-e:
Hi @rpgdev — thanks for the report.

[...]

Issue #1752, opened 2026-04-05
chore: publish v1.33.0 to npm


trek-e:
## Context

v1.33.0 was released on GitHub but not published to npm. The automated release pipeline could not authenticate with the npm registry.

## What succeeded
- ✅ Full test suite (2161/2161 pass)
- ✅ Hooks build (9/9 validated and copied)
- ✅ CHANGELOG.md updated
- ✅ package.json bumped to 1.33.0
- ✅ Git tag `v1.33.0` created and pushed
- ✅ GitHub release created with zip artifact and SHA-256 checksums
- ✅ Release verified via GitHub API

## What needs manual intervention
- ❌ npm publish — `npm whoami` returned 401 Unauthorized

## Steps to complete
```bash
npm login
npm publish --access public
```

## Verify
```bash
npm info get-shit-done-cc version  # Should show 1.33.0
```

PR #1757, opened 2026-04-05
fix(install): skip registering community hooks when .sh files missing

### LINKED ISSUE #1711: Installer registers 3 community hooks in settings.json but never copies them to hooks/
## Bug

The installer (`bin/install.js`) registers all 9 hooks in the target `settings.json`, but only copies the 6 `.js` hooks to the `hooks/` directory. The 3 `.sh` community hooks are never deployed, causing hook execution errors on every session.

## Missing files

These are registered in `settings.json` but never written to `hooks/`:

| Hook | Event | Matcher |
|------|-------|---------|
| `gsd-session-state.sh` | SessionStart | — |
| `gsd-phase-boundary.sh` | PostToolUse | `Write\|Edit` |
| `gsd-validate-commit.sh` | PreToolUse | `Bash` |

## Root cause

`scripts/build-hooks.js` correctly lists all 9 hooks in `HOOKS_TO_COPY` and copies them to `hooks/dist/`. However, the installer appears to only deploy the `.js` files from dist while still writing all 9 hook entries into `settings.json`.

## Impact

- **Every Bash tool call** triggers a failed hook lookup for `gsd-validate-commit.sh`
- **Every session start** triggers a failed hook lookup for `gsd-session-state.sh`
- **Every Write/Edit** triggers a failed hook lookup for `gsd-phase-boundary.sh`

The hooks themselves are opt-in no-ops (they exit 0 unless `.planning/config.json` has `hooks.community: true`), so the fix is either:

1. **Include the `.sh` files in the install** (they're harmless no-ops without opt-in), or
2. **Don't register them in `settings.json`** unless the user has opted in

Option 2 seems cleaner — don't register hooks that will always be no-ops.

## Reproduction

```bash
npx get-shit-done-cc --claude --local
ls .claude/hooks/
# Only .js files present — no .sh files
cat .claude/settings.json
# All 9 hooks registered, including the 3 missing .sh files
```

## Environment

- GSD v1.32.0
- Claude Code CLI
- macOS Darwin 25.4.0

---

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

MiaoDX:
## Summary

- The 3 community `.sh` hooks (`gsd-validate-commit.sh`, `gsd-session-state.sh`, `gsd-phase-boundary.sh`) are registered in `settings.json` during install, but the `.sh` files are **not included in the published npm package** — `hooks/dist/` only contains `.js` files at publish time
- This causes persistent `PreToolUse hook error` / `PostToolUse hook error` on every tool call, since the shell tries to execute a missing script
- Guard each community hook registration with an `fs.existsSync()` check on the deployed hook file, so hooks are only registered when actually available

## Root cause

`package.json` `files` includes `hooks/dist`, and `prepublishOnly` runs `build-hooks.js` which copies both `.js` and `.sh` to `hooks/dist/`. However, the published v1.32.0 and v1.33.0 tarballs only contain `.js` files in `hooks/dist/` — the `.sh` files are absent. The installer then registers the hooks in `settings.json` pointing to files that were never deployed.

## Test plan

- [ ] Install GSD globally via `npx -y get-shit-done-cc --claude --global` — confirm no `.sh` hooks are registered in `~/.claude/settings.json` when the files are missing
- [ ] Verify no `PreToolUse hook error` messages appear during normal Claude Code usage
- [ ] When `.sh` files ARE present in `hooks/dist/` (e.g. local dev build), confirm hooks are still registered normally

Fixes #1711

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -5688,68 +5688,83 @@ function install(isGlobal, runtime = 'claude') {
     // and exits silently (no-op) if not enabled. This lets users enable
     // them per-project by adding: "hooks": { "community": true }
 
-    // Configure commit validation hook (Conventional Commits enforcement, opt-in)
-    const validateCommitCommand = isGlobal
-      ? 'bash ' + targetDir.replace(/\\/g, '/') + '/hooks/gsd-validate-commit.sh'
-      : 'bash ' + dirName + '/hooks/gsd-validate-commit.sh';
-    const hasValidateCommitHook = settings.hooks[preToolEvent].some(entry =>
-      entry.hooks && entry.hooks.some(h => h.command && h.command.includes('gsd-validate-commit'))
-    );
+    // Only register community hooks whose .sh files were actually deployed.
+    // In v1.32.0 the .sh files were missing from hooks/dist/ (build-hooks.js
+    // didn't include them), so the installer registered entries pointing to
+    // non-existent scripts, causing "hook error" on every tool call. See #1711.
+    const hooksDeployed = path.join(targetDir, 'hooks');
 
-    if (!hasValidateCommitHook) {
-      settings.hooks[preToolEvent].push({
-        matcher: 'Bash',
-        hooks: [
-          {
-            type: 'command',
-            command: validateCommitCommand,
-            timeout: 5
-          }
-        ]
-      });
-      console.log(`  ${green}✓${reset} Configured commit validation hook (opt-in via config)`);
+    // Configure commit validation hook (Conventional Commits enforcement, opt-in)
+    const validateCommitFile = path.join(hooksDeployed, 'gsd-validate-commit.sh');
+    if (fs.existsSync(validateCommitFile)) {
+      const validateCommitCommand = isGlobal
+        ? 'bash ' + targetDir.replace(/\\/g, '/') + '/hooks/gsd-validate-commit.sh'
+        : 'bash ' + dirName + '/hooks/gsd-validate-commit.sh';
+      const hasValidateCommitHook = settings.hooks[preToolEvent].some(entry =>
+        entry.hooks && entry.hooks.some(h => h.command && h.command.includes('gsd-validate-commit'))
+      );
+
+      if (!hasValidateCommitHook) {
+        settings.hooks[preToolEvent].push({
+          matcher: 'Bash',
+          hooks: [
+            {
+              type: 'command',
+              command: validateCommitCommand,
+              timeout: 5
+            }
+          ]
+        });
+        console.log(`  ${green}✓${reset} Configured commit validation hook (opt-in via config)`);
+      }
     }
 
     // Configure session state orientation hook (opt-in)
-    const sessionStateCommand = isGlobal
-      ? 'bash ' + targetDir.replace(/\\/g, '/') + '/hooks/gsd-session-state.sh'
-      : 'bash ' + dirName + '/hooks/gsd-session-state.sh';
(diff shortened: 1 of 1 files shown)


MiaoDX:
## Updated root cause analysis

The initial PR description was slightly off. Here's the full picture:

**v1.32.0** (latest on npm): `scripts/build-hooks.js` `HOOKS_TO_COPY` only has 6 `.js` entries — the 3 `.sh` community hooks are not listed. So `prepublishOnly` → `build-hooks.js` never copies them to `hooks/dist/`, and the published tarball doesn't contain them. But `install.js` unconditionally registers all 3 `.sh` hooks in `settings.json`.

**Result**: every Bash tool call triggers `PreToolUse:Bash hook error` because `gsd-validate-commit.sh` doesn't exist on disk.

**`main` already has the `build-hooks.js` fix** (`.sh` files added to `HOOKS_TO_COPY`), but it hasn't shipped to npm yet (v1.33.0 is tagged, not published).

**This PR** adds defense-in-depth: `install.js` now checks `fs.existsSync()` on each `.sh` hook before registering it. This prevents this class of bug from recurring.

Force-pushed to replace the previous commit.

MiaoDX:
Note: the build-side fix already landed in [`8d732ee`](https://github.com/gsd-build/get-shit-done/commit/8d732ee4c0eee2ef25d4c71a21a1754ad26ed437) ("fix: resolve #1656 and #1657 — bash hooks missing from dist, SDK install prompt"), but hasn't shipped to npm yet (v1.32.0 is the latest published). This PR adds the installer-side guard so the issue can't recur even if hooks/dist drifts again.

trek-e (commented):
## Review: Duplicate of existing fix

This PR addresses the same root cause that was already fixed on `main` in commit `8d732ee` (build-hooks.js now includes `.sh` files in `HOOKS_TO_COPY`). The fix takes a different approach (guard at registration time vs. fix at build time), but the underlying issue no longer exists in the current codebase.

**What this PR does:** Wraps each community hook registration in `fs.existsSync()` so missing `.sh` files are silently skipped.

**What `main` already does:** `build-hooks.js` now copies `.sh` files to `hooks/dist/`, so they are always present after build. The registration code runs against files that exist.

**Why the existing fix is preferable:** Guarding registration silently hides a broken build pipeline. If `.sh` files are missing, that's a build bug that should be caught and fixed — not papered over at install time. The root fix ensures the build produces correct output.

This PR is a duplicate of the fix already on `main`. Closing.

Related issues: #1711, #1754, #1755

trek-e:
Closing as duplicate — the root cause (missing .sh files in hooks/dist/) was fixed on main in commit 8d732ee by adding .sh files to build-hooks.js HOOKS_TO_COPY. The npm publish blocker (#1752) is the remaining issue preventing users from getting the fix.

MiaoDX:
Updated approach: replaced the silent `fs.existsSync` installer guard with a **regression test** that fails loudly if install.js registers hooks not in `build-hooks.js` `HOOKS_TO_COPY`.

Verified the test catches the exact v1.32.0 bug:
```
✖ every registered hook is in HOOKS_TO_COPY so it gets built into dist
  install.js registers hooks missing from build-hooks.js HOOKS_TO_COPY: 
  gsd-validate-commit.sh, gsd-session-state.sh, gsd-phase-boundary.sh.
  These will not be packaged into hooks/dist/ and will cause "hook error" at runtime.
```

The build-side fix landed in [`8d732ee`](https://github.com/gsd-build/get-shit-done/commit/8d732ee) — this test ensures it can't regress.

PR #1774, opened 2026-04-05
fix(install): register gsd-workflow-guard.js in settings.json hooks

### LINKED ISSUE #1767: gsd-workflow-guard.js installed to disk but never registered in settings.json
## Bug

`gsd-workflow-guard.js` is included in `scripts/build-hooks.js` `HOOKS_TO_COPY` (line 23), copied to `hooks/dist/` during build, and installed to `~/.claude/get-shit-done/hooks/` during install. However, it is **never registered** in `settings.json` — there is no hook registration block for it anywhere in `bin/install.js`.

## Impact

- The file sits on disk as dead weight after every install
- Documentation (ARCHITECTURE.md, FEATURES.md, USER-GUIDE.md) describes it as an opt-in `PreToolUse` hook enabled via `hooks.workflow_guard: true` in `.planning/config.json`, but the hook is never invoked by the runtime since it's not in settings.json
- The `hooks.workflow_guard` config flag is permanently inert

## Expected behavior

Either:
1. Register `gsd-workflow-guard.js` in settings.json (like the other community hooks) so it can actually be invoked, OR
2. Remove it from `HOOKS_TO_COPY` and `hooks/dist/` if it's not ready for use, and update docs accordingly

## Found during

Comprehensive install process audit for #1755. The uninstall file cleanup was fixed in that PR (added `gsd-workflow-guard.js` to the `gsdHooks` array), but the registration gap remains.

Tibsfox:
## Summary

- **Register the workflow guard hook in `settings.json`** during install, closing the gap where `gsd-workflow-guard.js` was copied to disk but never wired up as a PreToolUse entry
- **Add test coverage** to verify the registration pattern exists in `install.js`

## Background

GSD hooks have a two-step install pattern: the hook file is **copied to disk** (via `HOOKS_TO_COPY`) and then **registered in `settings.json`** so Claude Code actually invokes it. The workflow guard was missing the second step — it was dead weight on disk, and the `hooks.workflow_guard` config option in `.planning/config.json` was permanently inert no matter what value you set.

This is the same pattern used by the prompt-guard and read-guard hooks:

1. Build the command string (respecting `isGlobal` vs local install)
2. Check if the hook is already registered (idempotent — safe to re-run install)
3. Push the hook entry with the appropriate `matcher` and `timeout`

The hook remains **opt-in at runtime** — `gsd-workflow-guard.js` checks `hooks.workflow_guard` in `.planning/config.json` and exits silently if not enabled, so this change has zero impact on users who haven't explicitly opted in.

## What changed

| File | Change |
|------|--------|
| `bin/install.js` | Added registration block for `gsd-workflow-guard.js` as a PreToolUse hook (matcher: `Write\|Edit`, timeout: 5) — inserted after the read-guard registration and before the community hooks section |
| `tests/install-hooks-copy.test.cjs` | Added test verifying the `hasWorkflowGuardHook` check and `gsd-workflow-guard` string exist in `install.js` |

## Test plan

- [x] All 2202 existing tests pass (`npm test`)
- [x] New test verifies registration pattern exists in `install.js`
- [ ] Manual: fresh install should show "Configured workflow context guard hook (opt-in via config)" in output
- [ ] Manual: re-install should skip registration (idempotent check)
- [ ] Manual: enabling `hooks.workflow_guard: true` in `.planning/config.json` should now actually trigger the hook on Write/Edit

Fixes #1767

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -5630,6 +5630,30 @@ function install(isGlobal, runtime = 'claude') {
       console.log(`  ${green}✓${reset} Configured read-before-edit guard hook`);
     }
 
+    // Configure PreToolUse hook for workflow context guard (#1767)
+    // Advises when file edits happen outside a GSD workflow. Opt-in via
+    // hooks.workflow_guard: true in .planning/config.json (default: false).
+    const workflowGuardCommand = isGlobal
+      ? buildHookCommand(targetDir, 'gsd-workflow-guard.js')
+      : 'node ' + dirName + '/hooks/gsd-workflow-guard.js';
+    const hasWorkflowGuardHook = settings.hooks[preToolEvent].some(entry =>
+      entry.hooks && entry.hooks.some(h => h.command && h.command.includes('gsd-workflow-guard'))
+    );
+
+    if (!hasWorkflowGuardHook) {
+      settings.hooks[preToolEvent].push({
+        matcher: 'Write|Edit',
+        hooks: [
+          {
+            type: 'command',
+            command: workflowGuardCommand,
+            timeout: 5
+          }
+        ]
+      });
+      console.log(`  ${green}✓${reset} Configured workflow context guard hook (opt-in via config)`);
+    }
+
     // Community hooks — registered on install but opt-in at runtime.
     // Each hook checks .planning/config.json for hooks.community: true
     // and exits silently (no-op) if not enabled. This lets users enable
--- a/tests/install-hooks-copy.test.cjs
+++ b/tests/install-hooks-copy.test.cjs
@@ -197,6 +197,17 @@ describe('install.js source correctness', () => {
     );
   });
 
+  test('gsd-workflow-guard.js has settings.json registration', () => {
+    assert.ok(
+      src.includes('hasWorkflowGuardHook'),
+      'install.js should check for existing workflow guard hook before registering'
+    );
+    assert.ok(
+      src.includes("'gsd-workflow-guard'") || src.includes('"gsd-workflow-guard"'),
+      'install.js should register gsd-workflow-guard in settings.json hooks'
+    );
+  });
+
   test('phantom gsd-check-update.sh is not in uninstall hook list', () => {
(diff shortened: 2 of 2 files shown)


trek-e (changes requested):
## Duplicate of #1772

This PR fixes the same bug (#1767) as PR #1772 (by trek-e), which was opened earlier and targets the same branch. Both PRs:

- Add the same `workflowGuardCommand` path construction (identical `buildHookCommand` / `'node ' + dirName` pattern)
- Add the same `hasWorkflowGuardHook` dedup check
- Add the same `.push()` registration block with `matcher: 'Write|Edit'` and `timeout: 5`
- Reference the same issue (`Fixes #1767`)

The only differences are cosmetic:

1. **Placement**: This PR inserts the block *before* the community hooks comment; #1772 inserts it *after* (inside the community hooks section, before validate-commit). Neither placement is wrong, but #1772's position groups it with the other opt-in community hooks, which is where it logically belongs.

2. **Test coverage**: This PR adds 2 assertions to the existing `install-hooks-copy.test.cjs`. PR #1772 adds a dedicated test file with 4 tests, including an anti-pattern guard that scans all JS hooks in `gsdHooks` and verifies each has a corresponding command construction -- catching this entire class of bug for any future hooks. The anti-pattern guard is the more valuable contribution.

3. **Console message**: Minor wording difference (`"Configured workflow context guard hook (opt-in via config)"` vs `"Configured workflow guard hook (opt-in via hooks.workflow_guard)"`). #1772's message is more specific about the config key.

**Recommendation**: Close this PR in favor of #1772, which was first, has stronger test coverage, and better placement.

trek-e:
Closing in favor of #1772, which fixes the same bug (#1767) with:

- **Better placement** — inside the community hooks section where opt-in hooks belong
- **Stronger test coverage** — 4 dedicated tests including an anti-pattern guard that scans all JS hooks in `gsdHooks` to verify each has a registration block, preventing this entire class of bug for any future hooks
- **More specific console output** — references the exact config key `hooks.workflow_guard`

Thank you for the contribution — the fix itself was correct.

Issue #1395, opened 2026-03-26
Support for multiple Developers on a project


ShaePhillip:
### Problem or motivation

When working on the same project with multiple developers (each working on different features) when committing to github the .planning folder gets committed even if it is in the gitignore file, GSD forecfully commits these folders. Which then ruins other devs planning structure. 

### Proposed solution

Have an option to not commit the .planning folder, but keep it locally. 

### Which area does this affect?

Planning system (phases, roadmap, state)

### Applicable runtimes

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Codex
- [ ] Copilot
- [ ] Antigravity
- [ ] Cursor
- [ ] Windsurf
- [x] All runtimes

### Additional context

This might already be available, I am not the most technical person and am only getting started in the Claude Code and open source space

MicBer007:
I get this issue a lot when running GSD with Claude Code. The agent sometimes picks up that I've set "commit_docs" to false, but sometimes just forcefully commits anyway. This causes big issues down the line, with some developers on our team having an entire day's planning overwritten, etc.

Tibsfox:
Good news — this is actually already supported. GSD checks for `.planning/` in `.gitignore` before committing and skips if it's ignored (`commands.cjs` lines 244-249):

```javascript
if (isGitIgnored(cwd, '.planning')) {
  const result = { committed: false, hash: null, reason: 'skipped_gitignored' };
  output(result, raw, 'skipped');
  return;
}
```

There's also a `commit_docs` config option in `.planning/config.json` — setting it to `false` will skip all planning commits regardless of gitignore.

**For multi-developer setups**, the recommended approach is:

1. Add `.planning/` to your `.gitignore`
2. Each developer gets their own local planning state that doesn't conflict
3. GSD will detect the gitignore and skip committing planning artifacts

If you want planning artifacts tracked but per-developer, you could also use the workspace feature (`/gsd:new-workspace`) which creates isolated planning directories.

This might be worth adding to the docs/FAQ since it's a common question for team setups.

Tibsfox:
Thanks for the follow-up — that's useful context. I investigated further.

All GSD workflows use `gsd-tools commit` for `.planning/` artifacts, which correctly checks `commit_docs` before committing. The raw `git add`/`git commit` calls in workflows are only for source code (test files, etc.), not planning artifacts.

[...]

Issue #1250, opened 2026-03-20
commit_docs gitignore auto-detection not propagated to init output — causes planning files to be re-committed


ctolmsted:
## Summary

When `.planning/` is in `.gitignore`, the `init` commands (e.g., `init phase-op`, `init execute-phase`) still return `commit_docs: true` in their JSON output if the config file doesn't explicitly set `commit_docs: false`. The documented "auto-detection" behavior only exists inside `cmdCommit`, not in `loadConfig()` or `init` output.

This causes planning files to be repeatedly committed to git despite `.planning/` being in `.gitignore`.

## Version

1.26.0

## Steps to Reproduce

1. Add `.planning/` to `.gitignore`
2. Do NOT set `commit_docs: false` in `.planning/config.json` (rely on documented auto-detection)
3. Run any init command:
   ```bash
   node gsd-tools.cjs init phase-op 1
   ```
4. Observe the output includes `"commit_docs": true`

## Root Cause

There are two separate issues interacting:

### Issue 1: `loadConfig()` doesn't implement auto-detection

`planning-config.md` (line 56) documents:

> **Auto-detection:** If `.planning/` is gitignored, `commit_docs` is automatically `false` regardless of config.json.

But `loadConfig()` in `core.cjs` reads the config file and defaults to `true`. It never checks `.gitignore`. The auto-detection only exists as a separate gate inside `cmdCommit()` (line 232 of `commands.cjs`).

Since all `init` commands call `loadConfig()` and return its `commit_docs` value, every workflow sees `commit_docs: true` in its init JSON even when `.planning/` is gitignored.

### Issue 2: LLM executor agents interpret `commit_docs: true` as intent

Workflows parse `commit_docs` from init JSON (e.g., `execute-plan.md` line 22). When `gsd-tools.cjs commit` returns `{"committed": false, "reason": "skipped_gitignored"}`, the LLM executor agent has conflicting signals:

- Init context says `commit_docs: true` → "planning docs should be committed"  
- `gsd-tools.cjs commit` returned "skipped" → "the commit didn't happen"

The agent resolves this conflict by running raw `git add .planning/... && git commit` to "complete" the step, bypassing the gitignore gate entirely. This is consistent with the executor's task-commit protocol (`git-integration.md` lines 83-109) which uses raw git for code commits.

## Evidence

[...]

Issue #968, opened 2026-03-08
`commit-docs` is not a valid gsd-tools command


CVirus:
### GSD Version

1.22.4

### Runtime

Claude Code

### What happened?

> Disclaimer: This bug was detected by a human programmer, but the report itself is AI-generated. I've reviewed it thoroughly, though my JavaScript experience is limited. I don't want to waste open-source maintainers' time by asking you to review an AI-written report — so please let me know if more effort is needed from my side for you to look into this.

Two workflows reference `gsd-tools.cjs commit-docs` which doesn't exist. The valid command is `commit`. This causes a silent failure during the commit step — the docs commit is skipped entirely.

Affected files:
- `workflows/validate-phase.md` line 131: `node "./.claude/get-shit-done/bin/gsd-tools.cjs" commit-docs "docs(phase-${PHASE}): add/update validation strategy"`
- `workflows/plan-phase.md` line 240: `commit-docs "docs(phase-${PHASE}): add validation strategy"`

Available commands from `gsd-tools.cjs`: `state, resolve-model, find-phase, commit, verify-summary, verify, frontmatter, template, generate-slug, current-timestamp, list-todos, verify-path-exists, config-ensure-section, init`

### What did you expect?

The workflows should use `commit` (with `--files` flag) instead of `commit-docs`, or `commit-docs` should be added as a command/alias in gsd-tools.

### Steps to reproduce

1. Run `/gsd:validate-phase` on a completed phase
2. Workflow reaches Step 7 (Commit) and runs `gsd-tools.cjs commit-docs "docs(phase-01): ..."`
3. Exits with code 1: `Error: Unknown command: commit-docs`

### Relevant logs or error messages

```shell
Bash(node "./.claude/get-shit-done/bin/gsd-tools.cjs" commit-docs "docs(phase-01): add/update validation strategy" 2>&1)
  ⎿  Error: Exit code 1
     Error: Unknown command: commit-docs
```

PR #1802, opened 2026-04-05
fix(workflows): respect commit_docs:false in worktree merge and quick task commits

### LINKED ISSUE #1783: .planning Folder get's commited via Bash Comands
### GSD Version

1.32.0

### Runtime

OpenCode

### Operating System

Windows

### Node.js Version

v24.12.0

### Shell

_No response_

### Installation Method

### What happened?

After using **/gsd-new-project** I need to choose if the **.planning** Folder should get committet as well. I choose "no". The **.planning** get's included within the **.gitignore** File, but there seems to be hardcoded Stuff within GSD which tells the AI to use Bash Commands to commit the **.planning** Folder.

### What did you expect?

That my choice of not commiting the **.planning** Folder will get respected

### Steps to reproduce

1. Install GSD via **npx get-shit-done-cc@latest**
2. Choose **OpenCode**
3. Choose **Local**
4. Choose **1** to not install the _gsd-sdk_
5. Type: **/gsd-new-project** and type your idea in it
6. Go through the init process of **/gsd-new-project** until it asks about Version Control
7. Choose to NOT commit the **.planning** Folder
8. Go on with the project and you will see that the **.planning** Folder get's committet via Bash Commands regardless

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

Minor — Cosmetic or edge case

### Workaround (if any)

_No response_

### Additional context

_No response_

### Privacy Checklist

Tibsfox:
## Summary

When a user chooses `commit_docs: false` during `/gsd-new-project` setup, the expectation is that `.planning/` files stay local and never get committed. The `gsd-tools.cjs commit` wrapper correctly checks this flag (see `commands.cjs` line 265) and skips `.planning/` files — but three locations in the workflow markdown files used raw `git add .planning/` commands that bypassed the wrapper entirely.

This meant that even after explicitly opting out, users would still see `.planning/STATE.md` and `.planning/ROADMAP.md` showing up in their commits. A confusing experience when you've specifically said "no" to committing docs!

## Root cause

Three raw `git add` invocations bypass the `gsd-tools.cjs commit` wrapper:

1. **`execute-phase.md` (line ~529)** — The worktree merge protection step amends the merge commit with `.planning/STATE.md` and `.planning/ROADMAP.md` without checking `commit_docs`
2. **`quick.md` (line ~637)** — Identical worktree merge protection code (duplicated from execute-phase)
3. **`quick.md` (line ~776)** — `git add ${file_list}` stages all artifacts including `.planning/STATE.md` before the `gsd-tools.cjs commit` call gets a chance to filter them

## Fix

Each of the three locations now checks the `commit_docs` config flag via `gsd-tools.cjs config-get commit_docs` before staging any `.planning/` files:

- For the worktree merge protection blocks (locations 1 and 2): the `git add` + `git commit --amend` is wrapped in an `if [ "$COMMIT_DOCS" != "false" ]` guard
- For the file_list staging (location 3): when `commit_docs` is false, `.planning/` paths are filtered out of the file list before `git add`

The `gsd-tools.cjs commit` wrapper on the next line already handles this correctly — these raw `git add` calls were the only paths that bypassed the check.

## Safety

- When `commit_docs` is `true` (the default), behavior is completely unchanged
- The `config-get` call falls back to `"true"` if the config tool is unavailable, so existing installs without the flag are unaffected
- The fix uses the same `gsd-tools.cjs config-get` mechanism that the commit wrapper already relies on — no new dependencies

## Test plan

- [x] New test file `tests/commit-docs-bypass.test.cjs` verifies that every `git add .planning/` occurrence in both workflow files is preceded by a `commit_docs` config check
- [x] Full test suite passes (2205 tests, 0 failures)

Fixes #1783

### DIFF
--- a/get-shit-done/workflows/execute-phase.md
+++ b/get-shit-done/workflows/execute-phase.md
@@ -526,8 +526,12 @@ Execute each selected wave in sequence. Within a wave: parallel if `PARALLELIZAT
        # Amend merge commit with restored files if any changed
        if ! git diff --quiet .planning/STATE.md .planning/ROADMAP.md 2>/dev/null || \
           [ -n "$DELETED_FILES" ]; then
-         git add .planning/STATE.md .planning/ROADMAP.md 2>/dev/null || true
-         git commit --amend --no-edit 2>/dev/null || true
+         # Only amend the commit with .planning/ files if commit_docs is enabled (#1783)
+         COMMIT_DOCS=$(node "$HOME/.claude/get-shit-done/bin/gsd-tools.cjs" config-get commit_docs 2>/dev/null || echo "true")
+         if [ "$COMMIT_DOCS" != "false" ]; then
+           git add .planning/STATE.md .planning/ROADMAP.md 2>/dev/null || true
+           git commit --amend --no-edit 2>/dev/null || true
+         fi
        fi
 
        # Remove the worktree
--- a/get-shit-done/workflows/quick.md
+++ b/get-shit-done/workflows/quick.md
@@ -634,8 +634,11 @@ After executor returns:
 
        if ! git diff --quiet .planning/STATE.md .planning/ROADMAP.md 2>/dev/null || \
           [ -n "$DELETED_FILES" ]; then
-         git add .planning/STATE.md .planning/ROADMAP.md 2>/dev/null || true
-         git commit --amend --no-edit 2>/dev/null || true
+         COMMIT_DOCS=$(node "$HOME/.claude/get-shit-done/bin/gsd-tools.cjs" config-get commit_docs 2>/dev/null || echo "true")
+         if [ "$COMMIT_DOCS" != "false" ]; then
+           git add .planning/STATE.md .planning/ROADMAP.md 2>/dev/null || true
+           git commit --amend --no-edit 2>/dev/null || true
+         fi
        fi
 
        git worktree remove "$WT" --force 2>/dev/null || true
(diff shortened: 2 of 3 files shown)


trek-e (commented):
## Adversarial Review — PR #1802

**Verdict: Approve with one minor observation**

### Linked Issue

`Fixes #1783` is present in the PR body. Issue #1783 carries labels `bug` and `confirmed` — the confirmed-bug requirement from CONTRIBUTING.md is satisfied.

### Fix Correctness

All three bypass locations are real. Verified against `main`:

- `execute-phase.md` line 529: `git add .planning/STATE.md .planning/ROADMAP.md` — unguarded on main, guarded in this PR.
- `quick.md` line 637: identical worktree merge path — unguarded on main, guarded in this PR.
- `quick.md` line 776: `git add ${file_list}` where `file_list` explicitly includes `.planning/STATE.md` — unguarded on main, filtered in this PR.

The guard logic is correct. `config-get commit_docs` falling back to `"true"` on error preserves existing behavior for installs without the flag.

### Regression Test

`tests/commit-docs-bypass.test.cjs` is present and covers all three locations with structural source-code assertions (not runtime execution). This is the right approach for workflow markdown — a runtime test would require a full project setup; a structural test verifies the invariant directly.

One observation: the test operates on a 10-line backward window for the `git add .planning/` checks and a 500-char backward window for the regex scan. These windows are tight. If the guard is ever refactored to appear further above the `git add` (e.g., inside a function call or earlier in a larger block), the tests could pass while the guard is actually absent. This is a minor fragility, not a blocker — the current code satisfies the window constraints and the test would have caught the original bug. Worth noting for future maintainers.

### Overlap with Parallel Worktree Fix

No parallel branch or open PR for #1783 exists other than this one. The "parallel fix" mentioned in context does not appear to have materialized as a separate PR. There is no conflict or duplication to resolve.

### Historical Pattern

The git log shows this same class of bypass has been fixed in PRs #482, #968, #1044, #1250, #1395/#1550. This PR fixes two new occurrences in the worktree merge paths of `execute-phase.md` and `quick.md` that were not covered by those prior fixes. The fix is genuine, not a re-fix of already-patched code.

### Security

No concerns. No new dependencies. No shell injection vectors introduced.

### Style

Follows workflow conventions. Guard uses the same `gsd-tools.cjs config-get` mechanism that the commit wrapper already relies on. The fallback `|| echo "true"` is consistent with the existing pattern in `commands.cjs`.

### Summary

- Linked confirmed bug: yes
- Regression test present: yes
- Fix is correct and complete for the three identified bypass locations: yes
- No overlap with other open PRs: yes
- CI stated as passing (2205 tests, 0 failures): consistent with the diff

This is a clean, well-scoped fix that addresses the root cause without introducing new behavior for default configurations.

PR #1824, opened 2026-04-05
fix(install): preserve non-array hook entries during uninstall

### LINKED ISSUE #1825: fix(install): uninstall silently deletes non-array hook entries
### Describe the bug

The `uninstall()` function in `bin/install.js` silently deletes hook entries that don't have a `hooks` array property. When iterating over hook event lists (`SessionStart`, `PostToolUse`, etc.), entries without `entry.hooks` or where `entry.hooks` is not an array are mapped to `null` and filtered out by `.filter(Boolean)`.

This means any user-owned or third-party hook entry stored in a non-standard shape is permanently removed during GSD uninstall.

### To reproduce

1. Add a hook entry to `settings.json` under `hooks.SessionStart` that does not contain a `hooks` array (e.g. `{ "type": "custom", "command": "echo hello" }`)
2. Run GSD uninstall
3. The custom entry is silently deleted

### Expected behavior

Non-array hook entries should be preserved unchanged during uninstall. Only entries containing GSD hooks should be modified or removed.

### Actual behavior

All entries without a `hooks` array are deleted.

### Root cause

Introduced in commit `175d89e` when the per-event uninstall loops were consolidated. The old code used `return true` to preserve unknown shapes; the new code uses `return null` which `.filter(Boolean)` removes.

### Environment

- GSD version: v1.9.9+
- Node.js: 22+
- OS: all platforms

jeremymcs:
## Fix PR

---

## Linked Issue

Fixes #1825

---

## What was broken

`uninstall()` silently deleted hook entries that don't have a `hooks` array property. Any user-owned or third-party hook entry stored in a non-standard shape was permanently removed during GSD uninstall.

## What this fix does

Changes `return null` to `return entry` for non-array hook entries so they pass through uninstall filtering unchanged.

## Root cause

Introduced in commit `175d89e` when per-event uninstall loops were consolidated into a single loop. The old code used `return true` to preserve unknown entry shapes; the refactored code used `return null` which `.filter(Boolean)` removes.

### How I verified the fix

Added a regression test (`non-array hook entries are preserved during uninstall (#1825)`) that creates entries without a `hooks` array, runs the filter, and asserts they survive. Also fixed the mirrored `filterGsdHooks` helper in the test file to match the corrected production code.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] CHANGELOG.md updated if this is a user-facing fix

## Breaking changes

None

### DIFF
--- a/tests/install-hooks-copy.test.cjs
+++ b/tests/install-hooks-copy.test.cjs
@@ -325,7 +325,7 @@ describe('uninstall settings cleanup preserves user hooks', () => {
   function filterGsdHooks(entries) {
     return entries
       .map(entry => {
-        if (!entry.hooks || !Array.isArray(entry.hooks)) return null;
+        if (!entry.hooks || !Array.isArray(entry.hooks)) return entry;
         entry.hooks = entry.hooks.filter(h => !isGsdHookCommand(h.command));
         return entry.hooks.length > 0 ? entry : null;
       })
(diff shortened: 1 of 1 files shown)


trek-e (commented):
## Verification complete — merge-ready

**Issue link:** `Fixes #1825` — labeled `bug` + `confirmed-bug`. Issue-first rule satisfied. Fix PR template used correctly.

**CI:** All 5 checks green — ubuntu/22, ubuntu/24, macos/24, security scan, issue link check.

**Regression test:** Present in `tests/install-hooks-copy.test.cjs` — `non-array hook entries are preserved during uninstall (#1825)`. Test covers three entry shapes (non-array custom, GSD hook, non-array webhook), uses `JSON.parse(JSON.stringify(...))` deep-copy guard before mutation, and asserts both count and identity of surviving entries. Follows project conventions (node:test, node:assert/strict, no try/finally).

**Fix diff:** Single-line change in `bin/install.js` (`return null` → `return entry`). Correct. The `filterGsdHooks` helper in the test file was also mirrored correctly.

**Scope:** No scope creep. Only `bin/install.js` and `tests/install-hooks-copy.test.cjs` touched, both directly related to the reported bug.

**Review:** APPROVED by trek-e (jeremymcs also listed as approver per the PR record).

**Verdict: merge-ready.** No outstanding blockers. Safe to merge now.

PR #1841, opened 2026-04-06
fix(install): preserve non-array hook entries during uninstall

### LINKED ISSUE #1825: fix(install): uninstall silently deletes non-array hook entries

trek-e:
## Summary

- During uninstall, `settings.json` hook event arrays are iterated to strip GSD hook entries
- The map callback returned `null` for entries where `entry.hooks` was missing or not an array, causing them to be silently removed by `.filter(Boolean)`
- Fixed by returning the entry unchanged instead of `null` for these unrecognised shapes

## Root cause

In `bin/install.js` line 4435, the consolidation in commit `175d89e` changed `return true` (preserve) to `return null` (discard) for non-array hook entries. This silently deletes third-party hook registrations with non-standard shapes.

## Fix

`return null` → `return entry` for the non-array guard in the per-event uninstall loop.

## Test

Added `uninstall preserves non-array hook entries (#1825)` to `tests/install-hooks-copy.test.cjs` in the "uninstall settings cleanup preserves user hooks" describe block. The test was confirmed failing before the fix and passing after.

Fixes #1825

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -4432,7 +4432,7 @@ function uninstall(isGlobal, runtime = 'claude') {
         const before = JSON.stringify(settings.hooks[eventName]);
         settings.hooks[eventName] = settings.hooks[eventName]
           .map(entry => {
-            if (!entry.hooks || !Array.isArray(entry.hooks)) return null;
+            if (!entry.hooks || !Array.isArray(entry.hooks)) return entry;
             // Filter out individual GSD hooks, keep user hooks
             entry.hooks = entry.hooks.filter(h => !isGsdHookCommand(h.command));
             return entry.hooks.length > 0 ? entry : null;
--- a/tests/install-hooks-copy.test.cjs
+++ b/tests/install-hooks-copy.test.cjs
@@ -325,7 +325,7 @@ describe('uninstall settings cleanup preserves user hooks', () => {
   function filterGsdHooks(entries) {
     return entries
       .map(entry => {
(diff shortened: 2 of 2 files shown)


trek-e:
Closing as duplicate — #1824 (by @jeremymcs) was approved for the same fix and predates this PR. Merging that one instead to preserve contributor credit.

PR #1903, opened 2026-04-07
fix(core): add atomicWriteFileSync helper for crash-safe file writes

### LINKED ISSUE #1915: fix: fs.writeFileSync is not crash-safe — truncated files on process kill
## Description

All file writes in the GSD CLI use direct \`fs.writeFileSync()\` which is not atomic on POSIX — a process killed mid-write (SIGKILL, power loss, disk full) can produce a truncated file that is unparseable on next read.

The highest-risk target is STATE.md, which is the most frequently written file (every state patch, plan completion, phase transition).

## Impact

A truncated STATE.md, ROADMAP.md, or config.json blocks all further GSD operations until manually repaired. The probability is low per-write but accumulates over long sessions with many state transitions.

## Suggested fix

Add an \`atomicWriteFileSync()\` helper using the standard write-to-temp-then-rename pattern:

\`\`\`javascript
const tmp = filePath + '.tmp.' + process.pid;
fs.writeFileSync(tmp, content, encoding);
fs.renameSync(tmp, filePath); // atomic on POSIX
\`\`\`

Apply first to STATE.md writes as the highest-frequency target, then extend to other write paths.

Tibsfox:
Fixes #1915

## Summary

- Add `atomicWriteFileSync(filePath, content, encoding)` helper using write-to-temp + `renameSync`
- Apply to STATE.md write paths (`writeStateMd` and `readModifyWriteStateMd`) as the highest-frequency target
- Prevents truncated files when process is killed mid-write

## Context

All file writes in the GSD CLI use direct `fs.writeFileSync()` which is not atomic on POSIX — a process killed mid-write (SIGKILL, power loss, disk full) can produce a truncated file that is unparseable on next read.

The standard durable-write pattern is write-to-temp-then-rename:
```javascript
const tmp = filePath + '.tmp.' + process.pid;
fs.writeFileSync(tmp, content, encoding);
fs.renameSync(tmp, filePath); // atomic on POSIX
```

STATE.md is the highest-frequency write target (every state patch, plan completion, phase transition), so it benefits most from this protection. Other write paths (ROADMAP.md, config.json) can adopt this helper in follow-up PRs.

The helper includes cleanup of the temp file on error to prevent orphaned `.tmp` files.

## Test plan

- [x] All existing state and core tests pass (2650 passing)
- [x] `atomicWriteFileSync` exported and available for other modules
- [ ] Verify crash-safety: write large content, SIGKILL mid-write, confirm file is old or new, never truncated

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -418,6 +418,29 @@ function isGitIgnored(cwd, targetPath) {
   }
 }
 
+// ─── Atomic file writes ─────────────────────────────────────────────────────
+
+/**
+ * Atomic file write: writes to a temp file then renames into place.
+ * Prevents truncated files when the process is killed mid-write.
+ *
+ * The rename(2) syscall is atomic on POSIX — the destination file is
+ * either the old content or the new content, never a partial write.
+ * On Windows, fs.renameSync is not guaranteed atomic but still avoids
+ * the most common truncation scenarios.
+ */
+function atomicWriteFileSync(filePath, content, encoding = 'utf-8') {
+  const tmpPath = filePath + '.tmp.' + process.pid;
+  try {
+    fs.writeFileSync(tmpPath, content, encoding);
+    fs.renameSync(tmpPath, filePath);
+  } catch (err) {
+    // Clean up the temp file on failure (best-effort)
+    try { fs.unlinkSync(tmpPath); } catch { /* already gone or never created */ }
+    throw err;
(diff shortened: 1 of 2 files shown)


trek-e:
Closing in favor of #1943, which is a more comprehensive fix: adds regression tests (`tests/atomic-write.test.cjs`) and applies `atomicWriteFileSync` to `config.cjs` and `roadmap.cjs` in addition to `core.cjs` and `state.cjs`.

PR #1904, opened 2026-04-07
fix(core): register held locks for cleanup on process exit

### LINKED ISSUE #1916: fix: lock files persist after process.exit() — blocks subsequent operations
## Description

When \`error()\` calls \`process.exit(1)\` inside a locked region, the \`finally\` block that releases the lock never executes. This leaves stale lock files on disk until the staleness timeout (10 seconds for state lock, 30 seconds for planning lock). During that window, any other process trying to acquire the lock spin-waits or fails.

## Impact

After a GSD command fails with \`error()\` while holding a lock, subsequent commands are blocked for 10-30 seconds. In parallel agent execution, this can cascade — one agent's failure blocks all others waiting for the same lock.

## Suggested fix

Track held lock files in a module-level Set and register a \`process.on('exit')\` handler to clean them up. The \`exit\` event fires even on \`process.exit(1)\`, providing immediate cleanup on the most common abnormal exit path. This complements the existing staleness detection (which handles truly orphaned locks from SIGKILL).

Tibsfox:
Fixes #1916

## Summary

- Track held lock files in a module-level Set (`_heldLocks`)
- Register a `process.on('exit')` handler that cleans up any remaining locks
- Applied to both `withPlanningLock` (core.cjs) and `acquireStateLock` (state.cjs)

## Context

When `error()` calls `process.exit(1)` inside a locked region, the `finally` block that releases the lock never executes. This leaves stale lock files on disk until the staleness timeout (10 seconds for state lock, 30 seconds for planning lock). During that window, any other process trying to acquire the lock spin-waits.

The fix registers locks on acquisition and unregisters them on normal release. The `process.on('exit')` handler is the safety net — it fires even on `process.exit(1)` and cleans up any locks that were never normally released.

This complements the existing staleness detection (which recovers from truly orphaned locks after timeout) by providing immediate cleanup on the most common abnormal exit path.

## Test plan

- [x] All existing lock, state, and core tests pass (2650 passing)
- [x] `_heldLocks` exported for cross-module use (state.cjs imports from core.cjs)
- [ ] Verify: trigger `error()` inside locked region, confirm lock file cleaned up immediately

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -24,6 +24,14 @@ const WORKSTREAM_SESSION_ENV_KEYS = [
   'ZELLIJ_SESSION_NAME',
 ];
 
+// Track held lock files for cleanup on unexpected exit (e.g. process.exit in error())
+const _heldLocks = new Set();
+process.on('exit', () => {
+  for (const lockPath of _heldLocks) {
+    try { fs.unlinkSync(lockPath); } catch { /* ignore */ }
+  }
+});
+
 let cachedControllingTtyToken = null;
 let didProbeControllingTtyToken = false;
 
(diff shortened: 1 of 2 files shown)


trek-e:
Closing — superseded by #1944, which addresses all four locking issues (#1909, #1916, #1925, #1927) in a single comprehensive fix with a unified regression test suite.

PR #1921, opened 2026-04-07
fix(core): resolve @file: references in gsd-tools stdout

### LINKED ISSUE #1891: Initialize Context State Error
### GSD Version

1.34.1

### Runtime

OpenCode

### Operating System

Windows

### Node.js Version

v24.14.1

### Shell

Powershell 7

### Installation Method

### What happened?

I ran /gsd-debug and I've got the following Error:

> #Initialize context state
> $ INIT=$(node "D:/Development/OpenCode Projects/My-Project-I/.opencode/get-shit-done/bin/gsd-tools.cjs" state load)
> if [[ "$INIT" == @file:* ]]; then INIT=$(cat "${INIT#@file:}"); fi
> ParserError: 
> Line |
>    2 |  if [[ "$INIT" == @file:* ]]; then INIT=$(cat "${INIT#@file:}"); fi
>      |    ~
>      | Missing '(' after 'if' in if statement.

I assume, that this error happens because I use Powershell 7, which may not know such syntax.

### What did you expect?

That there is no error when initializing the context.

### Steps to reproduce

1. Install GSD using `npx get-shit-done-cc@latest`
2. Choose OpenCode and Local
3. Run /gsd-debug
4. Should show an Parsing Error (if it's creating a context, which is not always the case)

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

Most of the time

### Impact

Minor — Cosmetic or edge case

### Workaround (if any)

_No response_

### Additional context

_No response_

### Privacy Checklist

Tibsfox:
Fixes #1891

## Summary

- Intercept stdout in `gsd-tools.cjs` to transparently resolve `@file:` references before they reach the calling shell
- Eliminates the need for bash-specific `if [[ "$INIT" == @file:* ]]` checks in 40+ workflow files
- The existing bash checks become harmless no-ops and can be removed in a follow-up cleanup

## Context

When `output()` in `core.cjs` produces JSON larger than 50KB, it writes the content to a temp file and outputs `@file:/path` to stdout. Every workflow file then needed a bash-specific check to detect and resolve this:

```bash
INIT=$(node gsd-tools.cjs state load)
if [[ "$INIT" == @file:* ]]; then INIT=$(cat "${INIT#@file:}"); fi
```

This `[[ ... ]]` syntax is bash-specific and breaks on PowerShell (the default shell for OpenCode on Windows), producing `ParserError: Missing '(' after 'if' in if statement`.

The `--pick` code path in `gsd-tools.cjs` already handled `@file:` resolution internally (line 336). This PR extends the same pattern to the normal (non-pick) output path, so `@file:` references are always resolved before reaching stdout.

## How it works

`gsd-tools.cjs` intercepts `fs.writeSync(1, ...)` calls (the same technique the `--pick` path uses), captures the output, checks for the `@file:` prefix, reads the temp file if found, and writes the actual content to stdout. The workflow shell never sees `@file:`.

## Test plan

- [x] New test `bug-1891-file-resolution.test.cjs` — 3 assertions verifying @file: resolution exists in both code paths
- [x] All 170 config + state tests pass (no regressions from stdout interception)
- [ ] Manual: on Windows/PowerShell, verify `state load` output no longer starts with `@file:`

### DIFF
--- a/get-shit-done/bin/gsd-tools.cjs
+++ b/get-shit-done/bin/gsd-tools.cjs
@@ -355,7 +355,27 @@ async function main() {
     return;
   }
 
-  await runCommand(command, args, cwd, raw);
+  // Intercept stdout to transparently resolve @file: references (#1891).
+  // core.cjs output() writes @file:<path> when JSON > 50KB. The --pick path
+  // already resolves this, but the normal path wrote @file: to stdout, forcing
+  // every workflow to have a bash-specific `if [[ "$INIT" == @file:* ]]` check
+  // that breaks on PowerShell and other non-bash shells.
+  const origWriteSync2 = fs.writeSync;
+  const outChunks = [];
+  fs.writeSync = function (fd, data, ...rest) {
+    if (fd === 1) { outChunks.push(String(data)); return; }
+    return origWriteSync2.call(fs, fd, data, ...rest);
+  };
+  try {
+    await runCommand(command, args, cwd, raw);
+  } finally {
+    fs.writeSync = origWriteSync2;
+  }
+  let captured = outChunks.join('');
+  if (captured.startsWith('@file:')) {
+    captured = fs.readFileSync(captured.slice(6), 'utf-8');
+  }
+  origWriteSync2.call(fs, 1, captured);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


trek-e:
Closing and recreating as a maintainer PR from the rebased branch in the main repo (the upstream fork branch could not be pushed to). The fix is identical — same commit rebased onto current main.

Issue #1925, opened 2026-04-07
fix: 8 state commands have TOCTOU races — read outside lock, write inside


Tibsfox:
## Description

8 state commands read STATE.md outside the lock, modify in memory, then acquire the lock only for the write. Two concurrent calls (e.g., parallel wave agents) can both read the same content and the second write clobbers the first's changes.

**Affected commands:** `cmdStateUpdate`, `cmdStateAdvancePlan`, `cmdStateRecordMetric`, `cmdStateUpdateProgress`, `cmdStateAddDecision`, `cmdStateAddBlocker`, `cmdStateResolveBlocker`, `cmdStateRecordSession`, `cmdStateBeginPhase`

**Correct pattern already exists:** `cmdStatePatch` uses `readModifyWriteStateMd()` which holds the lock across the entire read-modify-write cycle.

## Impact

During parallel agent execution (wave-based execution with multiple plans), concurrent state updates can lose data. For example, two agents finishing their plans simultaneously: both read the same `Current Plan: 2`, both advance to `3`, the second write overwrites the first's decision/metric/blocker entries.

## Suggested fix

Migrate all 8 commands to use the `readModifyWriteStateMd()` pattern that `cmdStatePatch` already demonstrates. This is a mechanical refactor — the correct pattern exists in the same file.

Issue #1909, opened 2026-04-07
perf: acquireStateLock uses CPU-burning busy-wait instead of Atomics.wait


Tibsfox:
## Description

The state lock retry mechanism in `state.cjs` uses a busy-wait spin loop that pegs a CPU core for the full retry delay:

```javascript
while (Date.now() - start < delay) {}
```

The same codebase already uses the correct `Atomics.wait()` pattern in `withPlanningLock` (`core.cjs:619`), which yields the thread to the OS scheduler. The state lock should be aligned with this pattern.

## Impact

During parallel agent execution with lock contention, one CPU core is fully consumed for each waiting process instead of yielding to the OS scheduler. This can degrade performance on machines running multiple concurrent GSD agents.

## Suggested fix

Replace the spin loop with `Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, delay)` to match the existing `withPlanningLock` implementation.

PR #1928, opened 2026-04-07
fix(config): add planning lock to setConfigValue

### LINKED ISSUE #1927: fix: config.json has no locking — concurrent config-set clobbers changes
## Description

`setConfigValue()` in `config.cjs` does a read-modify-write on `config.json` without any locking. Two concurrent `config set` commands to different keys causes the last writer to clobber the first writer's changes.

## Impact

During parallel agent execution, concurrent config updates silently lose data. For example, one agent sets `workflow.research=true` while another sets `workflow.discuss_mode=adaptive` — only the last write survives.

## Suggested fix

Wrap `setConfigValue` in `withPlanningLock(cwd, ...)`, reusing the existing planning lock infrastructure that `roadmap.cjs` already uses.

Tibsfox:
Fixes #1927

## Summary

- Wrap `setConfigValue()` read-modify-write in `withPlanningLock()` to prevent concurrent config-set commands from clobbering each other
- Import `withPlanningLock` from `core.cjs` into `config.cjs`

## Context

`setConfigValue()` does an unlocked read-modify-write on `config.json`. During parallel wave execution, two agents calling `config set` with different keys would lose one agent's changes. The planning lock (`withPlanningLock`) already exists and is used by `roadmap.cjs` for ROADMAP.md writes — this PR extends the same protection to config writes.

Create-only operations (`cmdConfigNewProject`, `ensureConfigFile`) don't need locking since they write fresh configs rather than modifying existing content.

## Test plan

- [x] New test `config-locking.test.cjs` — 2 assertions verifying lock usage
- [x] All 77 existing config tests pass (no regressions)

### DIFF
--- a/get-shit-done/bin/lib/config.cjs
+++ b/get-shit-done/bin/lib/config.cjs
@@ -4,7 +4,7 @@
 
 const fs = require('fs');
 const path = require('path');
-const { output, error, planningRoot, CONFIG_DEFAULTS } = require('./core.cjs');
+const { output, error, planningRoot, withPlanningLock, CONFIG_DEFAULTS } = require('./core.cjs');
 const {
   VALID_PROFILES,
   getAgentToModelMapForProfile,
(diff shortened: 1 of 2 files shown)


trek-e:
Closing — superseded by #1944, which addresses all four locking issues (#1909, #1916, #1925, #1927) in a single comprehensive fix with a unified regression test suite.

PR #1942, opened 2026-04-07
fix(milestone): replace test()+replace() with replace()+comparison in cmdRequirementsMarkComplete

### LINKED ISSUE #1922: fix: requirements-mark-complete can miss matches due to global regex lastIndex state
## Description

`cmdRequirementsMarkComplete` in `milestone.cjs` uses `test()` then `replace()` on the same global-flag (`gi`) regex objects. With the global flag, `test()` advances `lastIndex`, so the subsequent `replace()` starts searching from the wrong position and can miss the match it just found.

The same function also constructs the table pattern regex twice — once for `test()` and again for `replace()` — which is wasteful and was added as a workaround for this exact bug.

## Impact

When marking requirements complete, the checkbox or table row update can silently fail — the function reports the requirement as "not found" when it actually exists in the document. This is most likely to occur when the match is near the start of the content (position 0), since `test()` advances `lastIndex` past it.

## Suggested fix

Replace the `test()` + `replace()` pattern with direct `replace()` + string comparison:

```javascript
// Before (buggy):
const pattern = new RegExp(`...`, 'gi');
if (pattern.test(content)) {         // advances lastIndex
  content = content.replace(pattern, '...');  // starts from wrong position
}

// After (correct):
const pattern = new RegExp(`...`, 'gi');
const after = content.replace(pattern, '...');
if (after !== content) {  // detect change without test()
  content = after;
}
```

Also remove the unnecessary `g` flag from the "done check" patterns that only use `test()` for existence checking.

trek-e:
## Fix

Closes #1922

## What changed

`cmdRequirementsMarkComplete` in `get-shit-done/bin/lib/milestone.cjs` used `test()` + `replace()` on the same `/gi` regex for the checkbox pattern, and created the table pattern regex twice (once for `test()`, once for `replace()`) as a workaround for the same issue.

The fix applies the pattern the issue recommends throughout:

```js
// Before
const pattern = new RegExp(`...`, 'gi');
if (pattern.test(content)) {
  content = content.replace(pattern, '...');
}

// After
const pattern = new RegExp(`...`, 'gi');
const after = content.replace(pattern, '...');
if (after !== content) {
  content = after;
}
```

This eliminates:
- The fragile `test()`+`replace()` pattern on the checkbox regex
- The double regex construction in the table path (the old workaround)
- The unnecessary `g` flag on the done-check patterns (`doneCheckbox`, `doneTable`) that are only used once with `test()`

## Tests added

Two regression tests in `tests/milestone.test.cjs` under the `requirements mark-complete command` suite:

1. Checkbox + table row both update correctly when only those two lines exist (no surrounding content)
2. Checkbox updates correctly when the requirement is the very first character of the file (position 0)

Full suite: 2658 tests, 0 failures.

### DIFF
--- a/get-shit-done/bin/lib/milestone.cjs
+++ b/get-shit-done/bin/lib/milestone.cjs
@@ -42,19 +42,17 @@ function cmdRequirementsMarkComplete(cwd, reqIdsRaw, raw) {
 
     // Update checkbox: - [ ] **REQ-ID** → - [x] **REQ-ID**
     const checkboxPattern = new RegExp(`(-\\s*\\[)[ ](\\]\\s*\\*\\*${reqEscaped}\\*\\*)`, 'gi');
-    if (checkboxPattern.test(reqContent)) {
-      reqContent = reqContent.replace(checkboxPattern, '$1x$2');
+    const afterCheckbox = reqContent.replace(checkboxPattern, '$1x$2');
+    if (afterCheckbox !== reqContent) {
+      reqContent = afterCheckbox;
       found = true;
     }
 
     // Update traceability table: | REQ-ID | Phase N | Pending | → | REQ-ID | Phase N | Complete |
     const tablePattern = new RegExp(`(\\|\\s*${reqEscaped}\\s*\\|[^|]+\\|)\\s*Pending\\s*(\\|)`, 'gi');
-    if (tablePattern.test(reqContent)) {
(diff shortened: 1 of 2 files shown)


trek-e:
Closing — #1923 was opened earlier for the same fix and is already mergeable. Closing this duplicate to avoid double-merging changes to `milestone.cjs`.

PR #1959, opened 2026-04-08
fix(executor): enforce acceptance_criteria as hard gate, not advisory text

### LINKED ISSUE #1958: Executor agents silently drop planned work — acceptance_criteria not enforced
## Problem

Executor agents (`gsd-executor`) silently skip planned work items without flagging them as incomplete. The `<acceptance_criteria>` blocks in plans are treated as advisory — agents read "MANDATORY" in the workflow text and ignore it anyway.

### Real-world impact

In a 5-plan phase execution, 3 executor agents collectively dropped:
- **HLS.js streaming** — `hls.js` was in package.json but never imported in MediaPlayer
- **Virtual scrolling** — `virtua` was in package.json but MediaGrid used manual scroll
- **4 components** — InspireCards, CinemaCurtain, AiFab, timeline page — all planned, none built
- **Skin-aware dashboard** — generic heading instead of per-skin conditionals

All of these had acceptance criteria that would have caught them. The executor wrote SUMMARY.md claiming success, and the verifier only partially caught the gaps because it checks `must_haves` truths (higher-level), not task-level acceptance criteria.

## Root Cause

`execute-plan.md` line 143 says:
```
MANDATORY acceptance_criteria check: After completing each task, if it has <acceptance_criteria>, 
verify EVERY criterion before moving to the next task.
```

But this is just text guidance. The agent has no structured verification loop, no failure handling, and no blocking gate. It reads the word "MANDATORY" the same way it reads any other instruction — as a suggestion it can deprioritize when running low on context or encountering complexity.

## Proposed Fix

Make acceptance criteria a **hard gate** with a concrete verification loop:

1. After each task, execute every acceptance criterion as a grep/file-check/CLI command
2. Log each result as PASS/FAIL with command output
3. If ANY fail → fix immediately → re-run ALL criteria
4. BLOCK the next task until the gate clears
5. If a criterion can't be satisfied after 2 attempts → log as deviation (not silent skip)

Also strengthen the plan-level self-check (SUMMARY creation step):
- Re-run ALL acceptance criteria from every task before finalizing
- Re-run the plan-level `<verification>` commands
- Only then write Self-Check: PASSED/FAILED

## Additional Context

- The verifier (`gsd-verifier`) catches some gaps via `must_haves` truth checking, but this is a second pass that costs extra tokens and only catches high-level assertions
- Acceptance criteria are already well-written in plans — they just need enforcement
- This would also reduce verification cycles since executors would self-correct during execution

Version: 1.28.0

shootdaj:
## Summary

- Replace advisory "MANDATORY" text with a structured 5-step verification loop that blocks task progression
- Self-check step now re-runs ALL acceptance criteria and plan-level verification before finalizing SUMMARY

## Problem

The `MANDATORY acceptance_criteria check` instruction on line 148 of `execute-plan.md` is purely advisory. Executor agents read it and silently skip criteria when they run low on context or hit complexity. In a real 5-plan phase execution, 3 agents collectively dropped HLS.js streaming, virtual scrolling (virtua), 4 UI components, and skin-aware rendering — all of which had acceptance criteria that would have caught them.

## Changes

**Per-task gate (line 148):** Replaced single-line advisory with a 5-step verification loop:
1. Execute each criterion via grep/file-check/CLI
2. Log PASS/FAIL with output
3. Fix failures immediately, re-run all criteria
4. BLOCK next task until gate clears
5. After 2 failed fix attempts, log as deviation (not silent skip)

**Self-check (line 110):** Added two verification steps before SUMMARY finalization:
- Re-run ALL `<acceptance_criteria>` from every task
- Re-run plan-level `<verification>` commands

## Test plan

- [ ] Executor agent runs acceptance criteria after each task
- [ ] Failing criteria blocks progression to next task
- [ ] Self-check catches criteria missed during per-task execution
- [ ] Deviations logged when criteria can't be satisfied

Closes #1958

### DIFF
--- a/get-shit-done/workflows/execute-plan.md
+++ b/get-shit-done/workflows/execute-plan.md
@@ -110,6 +110,8 @@ Pattern B only (verify-only checkpoints). Skip for A/C.
 3. After ALL segments: aggregate files/deviations/decisions → create SUMMARY.md → commit → self-check:
    - Verify key-files.created exist on disk with `[ -f ]`
    - Check `git log --oneline --all --grep="{phase}-{plan}"` returns ≥1 commit
+   - Re-run ALL `<acceptance_criteria>` from every task — if any fail, fix before finalizing SUMMARY
+   - Re-run the plan-level `<verification>` commands — log results in SUMMARY
    - Append `## Self-Check: PASSED` or `## Self-Check: FAILED` to SUMMARY
 
    **Known Claude Code bug (classifyHandoffIfNeeded):** If any segment agent reports "failed" with `classifyHandoffIfNeeded is not defined`, this is a Claude Code runtime bug — not a real failure. Run spot-checks; if they pass, treat as successful.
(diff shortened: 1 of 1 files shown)


trek-e (approved):
## Review: APPROVE

### Summary

The change upgrades the per-task acceptance criteria check from advisory text to a structured verification loop with explicit PASS/FAIL logging and a hard block on proceeding. The self-check step now also re-runs all acceptance criteria and verification commands before finalizing SUMMARY.

### Assessment

The 5-step loop structure (execute → log PASS/FAIL → fix if fail → re-run all → log deviation after 2 failures) is a significant improvement over the previous single-line advisory. The "log as deviation after 2 attempts" fallback correctly prevents infinite loops while preserving visibility into what failed.

### Observations

This is a markdown/workflow change, not code — there are no unit tests for workflow behavior. The test plan items are marked unchecked, which is expected for workflow files. The core benefit is the explicit instruction structure that executor agents will follow more reliably.

The change is scoped to `execute-plan.md` only — consistent with the issue scope.

### No issues found

PR #2001, opened 2026-04-08
fix(hooks): skip read-guard advisory on Claude Code runtime

### LINKED ISSUE #1984: gsd-read-guard.js fires unconditionally on Claude Code, where the runtime already enforces read-before-edit
### GSD Version
1.34.2

### Runtime
Claude Code

### What happened?

`hooks/gsd-read-guard.js` injects a `READ-BEFORE-EDIT REMINDER` advisory before **every** `Write`/`Edit` call against an existing file, regardless of whether the model already read the file in the current session. On Claude Code this fires as a false positive on essentially every edit, because Claude reliably follows the read-before-edit pattern natively.

Example (real session, single edit to a file I had just Read 5 lines earlier):

```
[Edit ai.py] →
PreToolUse:Edit hook additional context: READ-BEFORE-EDIT REMINDER:
You are about to modify "ai.py" which already exists. If you have
not already used the Read tool to read this file in the current
session, you MUST Read it first before editing. ...
```

The edit succeeds (Claude Code's runtime correctly tracks read state), but the advisory still gets injected as a `<system-reminder>` block that the model has to reason about and acknowledge.

### What did you expect?

The hook should be a no-op on runtimes that already enforce read-before-edit at the runtime level (Claude Code). The original motivation in #1628 was OpenCode + MiniMax M2.5, which doesn't track read state and falls into infinite retry loops — that's a real bug worth fixing, but the fix shouldn't apply universally.

### Impact

- **Token waste**: Every Write/Edit eats ~80 tokens of injected reminder + the model's implicit acknowledgement.
- **Noise in tool flow**: Each edit produces a `<system-reminder>` that has to be addressed in the assistant turn, polluting transcripts and slightly degrading focus on the actual task.
- **Confuses users**: The reminder reads as if the model did something wrong, prompting "what hook is that?" debugging detours (this issue is being filed after exactly such a detour).

### Proposed fixes (any of these would work)

1. **Detect Claude Code at runtime and exit 0 early.** Claude Code sets `CLAUDECODE=1` in the environment when invoking hooks. A two-line check at the top of `gsd-read-guard.js` skips injection on Claude Code while keeping the safety net for OpenCode/MiniMax.
2. **Inspect the hook input payload** for runtime-identifying fields (if PreToolUse payload exposes any).
3. **Make the hook opt-in via GSD config** (`safetyRails.readGuard: true`), defaulting on for OpenCode and off for Claude Code.

Option 1 is the smallest diff and matches how the bug was originally scoped (#1628 was OpenCode-specific). Sketch:

\`\`\`js
// Skip on Claude Code — its runtime already enforces read-before-edit
if (process.env.CLAUDECODE === '1') process.exit(0);
\`\`\`

### Steps to reproduce

1. Install GSD on Claude Code.
2. In any session, \`Read\` a file, then \`Edit\` the same file.
3. Observe \`READ-BEFORE-EDIT REMINDER\` injected even though the file was just read.

### Privacy Checklist
- [x] Reviewed for PII

Tibsfox:
## Summary

- Skip the `gsd-read-guard.js` advisory when running inside Claude Code, which natively enforces read-before-edit at the runtime level
- Detection uses `CLAUDE_SESSION_ID` env var — the established Claude Code detection pattern already used in `core.cjs:15` and `workstream-flag.md`
- Saves ~80 tokens per Write/Edit call and eliminates noisy `<system-reminder>` blocks in Claude Code sessions
- Non-Claude runtimes (OpenCode, Gemini, Cursor, etc.) continue to receive the advisory as before

### Context

The read guard was added for non-Claude models (e.g. MiniMax M2.5 on OpenCode) that don't follow read-before-edit natively and get stuck in infinite retry loops. Claude Code doesn't have this problem — it tracks read state internally. The advisory was harmless but added token waste and noise on every edit.

### Test isolation

The `runHook()` test helper now sanitizes `CLAUDE_SESSION_ID` from the inherited environment so the positive-path tests pass even when the test suite runs inside a Claude Code session.

Closes #1984

## Test plan

- [x] New test: `skips advisory on Claude Code runtime (CLAUDE_SESSION_ID set)` — verifies empty output
- [x] All 13 read-guard tests pass
- [x] `runHook()` helper sanitizes env to prevent test failures inside Claude Code
- [x] Adversarial security review passed (0 critical/high findings)

### DIFF
--- a/hooks/gsd-read-guard.js
+++ b/hooks/gsd-read-guard.js
@@ -36,6 +36,11 @@ process.stdin.on('end', () => {
       process.exit(0);
     }
 
+    // Claude Code natively enforces read-before-edit — skip the advisory (#1984)
+    if (process.env.CLAUDE_SESSION_ID) {
+      process.exit(0);
+    }
+
     const filePath = data.tool_input?.file_path || '';
     if (!filePath) {
       process.exit(0);
--- a/tests/read-guard.test.cjs
+++ b/tests/read-guard.test.cjs
@@ -26,14 +26,17 @@ const HOOK_PATH = path.join(__dirname, '..', 'hooks', 'gsd-read-guard.js');
  * Run the read guard hook with a given tool input payload.
  * Returns { exitCode, stdout, stderr }.
  */
-function runHook(payload) {
+function runHook(payload, envOverrides = {}) {
   const input = JSON.stringify(payload);
+  // Sanitize CLAUDE_SESSION_ID so positive-path tests work inside Claude Code sessions
+  const env = { ...process.env, CLAUDE_SESSION_ID: '', ...envOverrides };
   try {
     const stdout = execFileSync(process.execPath, [HOOK_PATH], {
       input,
       encoding: 'utf-8',
       timeout: 5000,
       stdio: ['pipe', 'pipe', 'pipe'],
+      env,
     });
     return { exitCode: 0, stdout: stdout.trim(), stderr: '' };
   } catch (err) {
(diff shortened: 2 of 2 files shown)


trek-e (approved):
## Review: APPROVE

### Summary

Correct, minimal fix. The `CLAUDE_SESSION_ID` check is the established Claude Code detection pattern (already used in `core.cjs:15`). Placing it before any file path processing means the hook exits immediately with no output on Claude Code, which is the desired behavior.

### Test coverage

- New test passes `CLAUDE_SESSION_ID: 'test-session-123'` via env override — correctly exercises the new code path
- `runHook()` sanitizes `CLAUDE_SESSION_ID: ''` in the base env so existing positive-path tests still pass inside Claude Code sessions — good defensive practice
- 13 existing tests all continue to pass

### No issues found

The fix is scoped to the exact problem reported in #1984. Non-Claude runtimes are unaffected (they don't set `CLAUDE_SESSION_ID`).

Issue #1977, opened 2026-04-08
Defense-in-depth: executor worktree commit safety hardening


Asentient:
## Context

Filed after investigating production data loss caused by executor subagents silently deleting files during worktree commits. Root cause is a compound failure across Claude Code (#44965 — worktree branches from wrong commit) and model behavior (Opus ignoring staging instructions). GSD's executor prompt is **already correct** — this issue is about adding enforcement that doesn't rely on model compliance.

Related: [anthropics/claude-code#45108](https://github.com/anthropics/claude-code/issues/45108)

---

## 1. Bug: `worktree_branch_check` is labeled Windows-only

The `execute-plan.md` workflow includes a `<worktree_branch_check>` block that verifies the merge-base after worktree creation. This is documented as a "known Windows issue" — but the same bug (EnterWorktree branching from `main` instead of current HEAD) was reproduced on **Linux (Proxmox LXC)**.

**Fix:** Remove the platform qualifier. The check should be universal and always included when `isolation: "worktree"` is used.

---

## 2. Enhancement: Post-commit deletion verification

After each executor commit, GSD should verify no unexpected file deletions occurred:

```bash
# After git commit succeeds:
DELETIONS=$(git diff --diff-filter=D --name-only HEAD~1 HEAD)
if [ -n "$DELETIONS" ]; then
  echo "WARNING: Commit includes file deletions:"
  echo "$DELETIONS"
  # Optionally: abort, revert, or flag for orchestrator review
fi
```

**Why:** The executor prompt already says "NEVER `git add -A`" but the model ignores this ~5-10% of the time. This catch-net is cheap (one git command), non-blocking in the happy path, and would have caught all three incidents immediately instead of during milestone audit days later.

**Where:** Add to the commit protocol section in `gsd-executor.md`, after the commit step.

---

## 3. Enhancement: Merge-back deletion check

Before fast-forward merging a worktree branch onto the main branch, the orchestrator should verify no unexpected deletions:

```bash
# Before merge:
DELETIONS=$(git diff --diff-filter=D --name-only HEAD...worktree-branch)
if [ -n "$DELETIONS" ]; then
  echo "BLOCKED: Worktree branch contains deletions of $COUNT files"
  echo "$DELETIONS"
  # Do not merge — flag for user review
fi
```

**Why:** Even if the executor fails to catch it at commit time, the orchestrator can catch it before the damage reaches the main branch. This is the last line of defense.

[...]

PR #2028, opened 2026-04-10
fix(worktree): use reset --hard in worktree_branch_check

### LINKED ISSUE #2015: Worktree executor creates branch from master instead of feature branch HEAD
## Problem

When spawning `gsd-executor` with `isolation="worktree"` from a feature branch, the worktree is created based on `master` (or the default branch) instead of the current feature branch HEAD. The executor then works against stale code, and its commits contain a massive diff between master and the feature branch — making merge back impossible.

## Reproduction

1. Be on a feature branch ~100 commits ahead of master (e.g., `feat/KNOWL-81-mcp-server`)
2. Run `/gsd:quick` which spawns executor with `isolation="worktree"`
3. Executor's first commit shows the entire feature branch as deletions + its actual changes

## Expected behavior

Worktree should be based on the current branch HEAD (`git rev-parse HEAD` at spawn time).

## Workaround in workflow

The workflow includes a `worktree_branch_check`:

```
Run: git merge-base HEAD ${EXPECTED_BASE}
If differs → git reset --soft ${EXPECTED_BASE}
```

This doesn't actually fix the problem because `reset --soft` only moves the HEAD pointer — the working tree files remain from master. A `reset --hard` would fix it but is intentionally avoided as a destructive operation.

## Impact

- ~66K tokens wasted on executor that produced unusable commits
- Manual merge/cherry-pick failed (too many conflicts from master↔feature diff)
- Had to apply changes manually (4 files, 5 minutes) — the executor's work was thrown away

## Environment

- Linux (not Windows — the comment in the workflow says "known issue on Windows")
- Claude Code CLI
- GSD 1.x (latest as of 2026-04-09)

## Suggested fix

Either:
1. Pass the current branch name to `EnterWorktree` so it creates the worktree from the correct ref
2. Use `reset --hard` in the branch check (the worktree is disposable — hard reset is safe there)
3. Validate the worktree base BEFORE spawning the executor and abort early if wrong

trek-e:
## What

The `worktree_branch_check` in `execute-phase.md` and `quick.md` used `git reset --soft` as the recovery action when `EnterWorktree` created a branch from `main`/`master` instead of the current feature branch HEAD.

## Why

`reset --soft` moves the HEAD pointer but leaves the working tree files untouched — so the executor would start with stale code from main and produce commits containing the entire feature branch as deletions (a ~100-commit diff). The user's actual task changes would be buried in that noise and the resulting commits were impossible to merge.

## How

Replace `git reset --soft {EXPECTED_BASE}` with `git reset --hard {EXPECTED_BASE}` in both workflow files. `--hard` resets both HEAD and the working tree to the correct base commit. This is safe in a fresh worktree that has no user changes yet.

Adds 4 regression tests (2 per workflow) asserting that the check uses `--hard` and contains no `--soft`.

Closes #2015


trek-e (commented):
## Review — fix(worktree): use reset --hard in worktree_branch_check

### Correctness

The fix is correct. `reset --soft` only moves the HEAD pointer, leaving the working tree files from the wrong base (main/master) in place. The executor then stages and commits those stale files, producing a commit that contains the entire feature branch as deletions. `reset --hard` resets both HEAD and the working tree to the expected base commit — correct behavior in a fresh worktree with no user changes.

### Coverage

Both `execute-phase.md` and `quick.md` updated consistently. The platform note correctly broadens from 'Windows-only' to 'Windows/macOS' per #1977 item 1.

### Tests

4 regression tests in `tests/bug-2015-worktree-base-branch.test.cjs` — 2 per workflow. Each test extracts the `<worktree_branch_check>` block by regex and asserts (a) `reset --soft` is absent and (b) `reset --hard` is present. Correct pattern.

### Issue link

`Closes #2015` present in PR body.

**No issues found. Ready to merge.**

PR #2035, opened 2026-04-10
fix(install): guard writeSettings against null settingsPath for cline runtime

### LINKED ISSUE #2044: cline install crashes with ERR_INVALID_ARG_TYPE: null passed to fs.writeFileSync
## Bug Description

Running `npx get-shit-done-cc` and selecting the **cline** runtime crashes with:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received null
    at Object.writeFileSync (node:fs:...)
    at writeSettings (bin/install.js:...)
    at finishInstall (bin/install.js:...)
```

## Root Cause

`install()` returns `settingsPath: null` for cline (it uses `.clinerules` instead of `settings.json`), but `finishInstall()` was missing the `!isCline` guard that all other non-settings runtimes have — codex, copilot, cursor, windsurf, and trae are all guarded, cline was not. So `writeSettings(null, ...)` was called unconditionally, passing `null` to `fs.writeFileSync`.

## Steps to Reproduce

1. Run `npx get-shit-done-cc --cline --local` in any directory
2. Observe `ERR_INVALID_ARG_TYPE` crash

## Expected Behavior

Install completes without error; `.clinerules` is written; `settings.json` is not written.

## Fix

Add `!isCline` to the `writeSettings` guard in `finishInstall()`:

```js
if (!isCodex && !isCopilot && !isKilo && !isCursor && !isWindsurf && !isTrae && !isCline) {
  writeSettings(settingsPath, settings);
}
```

brkay54:
Closes #2044

## Summary

- When installing GSD with the `cline` runtime, the installer crashed with `ERR_INVALID_ARG_TYPE` because `fs.writeFileSync` received `null` as the path argument.

- `install()` returns `settingsPath: null` for `cline` (it uses `.clinerules` instead of `settings.json`), but `finishInstall()` was missing `!isCline` in the guard that prevents calling `writeSettings` for null-path runtimes — unlike `codex`, `copilot`, `cursor`, `windsurf`, and `trae` which were all guarded.

- Fix: add `const isCline = runtime === 'cline'` to `finishInstall()` and include `!isCline` in the `writeSettings` guard condition.

## Test plan

- [x] Run installer with `cline` runtime — completes without error
- [x] `.clinerules` is written to the target directory
- [x] `settings.json` is not written for `cline` runtime
- [x] Regression test added to `tests/cline-install.test.cjs`:
  - `finishInstall does not throw ERR_INVALID_ARG_TYPE for cline runtime` — asserts no crash with null settingsPath
  - `settings.json is not written for cline runtime` — asserts writeSettings is not invoked for cline

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -6124,6 +6124,7 @@ function finishInstall(settingsPath, settings, statuslineCommand, shouldInstallS
   const isCursor = runtime === 'cursor';
   const isWindsurf = runtime === 'windsurf';
   const isTrae = runtime === 'trae';
+  const isCline = runtime === 'cline';
 
   if (shouldInstallStatusline && !isOpencode && !isKilo && !isCodex && !isCopilot && !isCursor && !isWindsurf && !isTrae) {
     settings.statusLine = {
@@ -6134,7 +6135,7 @@ function finishInstall(settingsPath, settings, statuslineCommand, shouldInstallS
   }
 
   // Write settings when runtime supports settings.json
-  if (!isCodex && !isCopilot && !isKilo && !isCursor && !isWindsurf && !isTrae) {
+  if (!isCodex && !isCopilot && !isKilo && !isCursor && !isWindsurf && !isTrae && !isCline) {
     writeSettings(settingsPath, settings);
   }
 
--- a/bin/install.js
+++ b/bin/install.js
@@ -6484,6 +6484,7 @@ if (process.env.GSD_TEST_MODE) {
     validateHookFields,
     preserveUserArtifacts,
     restoreUserArtifacts,
+    finishInstall,
   };
 } else {
 
--- a/tests/cline-install.test.cjs
+++ b/tests/cline-install.test.cjs
@@ -154,6 +154,18 @@ describe('Cline install (local)', () => {
     assert.ok(fs.existsSync(engineDir), 'get-shit-done directory must exist after install');
   });
 
+  test('does not crash with ERR_INVALID_ARG_TYPE (regression: null settingsPath guard)', () => {
+    // Before fix: finishInstall() called writeSettings(null, ...) because isCline guard was missing
+    // After fix:  !isCline is in the writeSettings guard, matching codex/copilot/cursor/windsurf/trae
+    assert.doesNotThrow(() => install(false, 'cline'), /ERR_INVALID_ARG_TYPE/);
+  });
+
+  test('settings.json is not written for cline runtime', () => {
+    install(false, 'cline');
+    const settingsJson = path.join(tmpDir, '.cline', 'settings.json');
+    assert.ok(!fs.existsSync(settingsJson), 'settings.json must not be written for cline runtime');
+  });
+
   test('installed engine files have no leaked .claude paths', () => {
     install(false, 'cline');
     const engineDir = path.join(tmpDir, 'get-shit-done');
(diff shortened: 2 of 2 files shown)


trek-e (changes requested):
## Review: Changes Requested

### Issue-First Rule Violation

This PR has no linked issue. Per the contributing guidelines:

> PRs without a linked issue are closed without review, no exceptions.
> Use `Closes #NNN`, `Fixes #NNN`, or `Resolves #NNN` in the PR body.

The CI check confirms this: the **Issue link required** check is failing.

No matching issue was found under any of the likely searches (cline null settingsPath, cline writeSettings, cline ERR_INVALID_ARG_TYPE). Please file a [Bug Report issue](https://github.com/gsd-build/get-shit-done/issues/new?template=bug_report.yml) first, get it labeled `confirmed-bug`, then update this PR body with `Fixes #NNN`.

---

### Missing Regression Test

Per CONTRIBUTING.md, a bug fix **requires a regression test**:

> A regression test is required. Write the test first — it must demonstrate the original failure before your fix is applied, then pass after the fix. A PR that fixes a bug without a regression test will be asked to add one.

There is an existing test file at `tests/cline-install.test.cjs` that covers cline install behavior. A regression test must be added there that calls `finishInstall()` (or exercises the full install path) with `runtime === 'cline'` and asserts it does not throw `ERR_INVALID_ARG_TYPE`.

---

### Code Review (the logic itself is correct)

The fix is technically sound:
- `install()` already early-returns `{ settingsPath: null, ... }` for cline at line 5817-5835, matching the pattern for codex, copilot, cursor, windsurf, and trae.
- `finishInstall()` was the only place missing `!isCline` in the `writeSettings` guard. The placement of `const isCline = runtime === 'cline'` and the updated guard condition are both correct.
- No security issues, no ESM imports, no style violations.

One minor observation: the statusline guard at line 6128 also does not exclude `isCline`, but this is safe in practice because cline is excluded from `statuslineRuntimes` upstream, so `shouldInstallStatusline` will always be `false` for cline. It is not a blocker, but worth noting for completeness.

---

### Steps to Unblock

1. File a Bug Report issue describing the `ERR_INVALID_ARG_TYPE` crash on cline install.
2. Once labeled `confirmed-bug`, update this PR body with `Fixes #NNN`.
3. Add a regression test to `tests/cline-install.test.cjs`.
4. All three test matrix jobs are currently green — keep them green.

trek-e:
The issue backing this PR is now filed as #2044. To unblock review, please:

1. Add `Closes #2044` to the PR description body.
2. Add a regression test (e.g. in a new test file or extending an existing installer test) that verifies `finishInstall` does not call `writeSettings` when the runtime is cline.

Once those two items are addressed, this PR can be re-reviewed and merged.

trek-e (approved):
All three requirements from the previous review are addressed:

- ✅ `Closes #2044` in the PR body — issue-first rule satisfied
- ✅ Regression tests added to `tests/cline-install.test.cjs` — both tests call `finishInstall(null, null, null, false, 'cline', false, tmpDir)` directly, which correctly exercises the null guard (the prior attempt that called `install()` would have short-circuited before reaching `finishInstall`)
- ✅ `finishInstall` exported via `GSD_TEST_MODE` to make the test callable

Code is correct, tests are correct, pattern matches the existing cline test file. Approving for merge.

PR #2054, opened 2026-04-10
perf(state): cache buildStateFrontmatter disk scan per process

### LINKED ISSUE #1967: enhancement: cache buildStateFrontmatter disk scan per gsd-tools invocation
## Pre-submission checklist

`writeStateMd()` — every STATE.md write triggers `buildStateFrontmatter()` which lists all phase directories and counts files in each.

## Current behavior

Every `writeStateMd()` call triggers `buildStateFrontmatter()` (state.cjs:709) which does:
1. `readdirSync(phasesDir)` to list all phase directories
2. Per-phase `readdirSync(path.join(phasesDir, dir))` to count plan/summary files
3. Filter each by `-PLAN.md` and `-SUMMARY.md` patterns

For a 50-phase project, this is 51 `readdirSync` calls on every single state field update. State-heavy operations like `cmdStateBeginPhase` (which updates 7+ fields) trigger this scan on every write.

## Proposed behavior

Cache the directory scan results for the duration of a single `gsd-tools` invocation using a module-level variable. The cache is valid for the entire process lifetime since GSD tools are short-lived single-command processes (same reasoning as the `isGitIgnored` cache merged in #1898).

Two options:
1. **Lazy frontmatter**: Only rebuild frontmatter when explicitly requested (e.g., before commit) rather than on every write
2. **Cached scan**: Cache the `readdirSync` results in a module-level Map, invalidated only if a new phase directory is created during the same process

## Reason and benefit

Reduces per-write I/O by 90%+ for state-heavy operations. Most visible during phase execution where `beginPhase`, `advancePlan`, `recordMetric`, and `updateProgress` all write STATE.md in rapid succession — each currently triggers a full disk scan.

## Scope of changes

- `get-shit-done/bin/lib/state.cjs` — add module-level cache for `buildStateFrontmatter` disk scan results, clear on phase directory creation

## Breaking changes

None. Output is identical — only the I/O cost per write changes.

## Alternatives considered

1. **Remove frontmatter sync from writes entirely** — rejected; frontmatter is consumed by hooks and `state json`, removing it breaks the pipeline
2. **Move frontmatter sync to a separate `state sync` command** — rejected; would require all workflows to explicitly call sync, high risk of stale frontmatter
3. **Use `fs.watch` to invalidate cache** — rejected; overengineered for short-lived processes

## Area affected

Core workflow (init, plan, build, verify) — STATE.md is written on every phase transition.

Tibsfox:
## Summary

- Cache the N+1 `readdirSync` calls in `buildStateFrontmatter()` using a module-level Map keyed by `cwd`
- The disk scan (phases directory + each phase subdirectory) was repeated on every call, but the phase structure doesn't change within a single `gsd-tools` CLI invocation
- Cache auto-clears at process exit — no explicit invalidation needed since each CLI invocation is a short-lived process

For a 50-phase project, this eliminates ~50 redundant `readdirSync` calls per repeated `buildStateFrontmatter` invocation within the same process.

Closes #1967

## Test plan

- [x] All 93 state tests pass
- [x] Cache is transparent — same output, fewer disk reads

### DIFF
--- a/get-shit-done/bin/lib/state.cjs
+++ b/get-shit-done/bin/lib/state.cjs
@@ -7,6 +7,11 @@ const path = require('path');
 const { escapeRegex, loadConfig, getMilestoneInfo, getMilestonePhaseFilter, normalizeMd, planningDir, planningPaths, output, error, atomicWriteFileSync } = require('./core.cjs');
 const { extractFrontmatter, reconstructFrontmatter } = require('./frontmatter.cjs');
 
+// Cache disk scan results from buildStateFrontmatter per cwd per process (#1967).
+// Avoids re-reading N+1 directories on every state write when the phase structure
+// hasn't changed within the same gsd-tools invocation.
+const _diskScanCache = new Map();
+
 /** Shorthand — every state command needs this path */
 function getStatePath(cwd) {
   return planningPaths(cwd).state;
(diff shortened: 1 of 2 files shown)


trek-e (changes requested):
## Review: perf(state): cache buildStateFrontmatter disk scan per process

### Issue link
`Closes #1967` is present and #1967 carries `approved-enhancement`. ✓

### Commit prefix
`perf(state):` — correct conventional commit prefix. ✓

### Security
No security concerns. The cache key is `cwd` (an absolute path string), which is controlled by the caller, not user-supplied input. No injection surface introduced. ✓

---

## Bugs / correctness issues

### 1. `isDirInMilestone.phaseCount` accessed after scope ends — cached value is correct, but the logic extraction has a latent ordering issue

In the original code, `isDirInMilestone` is used as both a filter function and as a `.phaseCount` property holder. In the refactored code, the cached object stores `totalPhases` using:

```js
totalPhases: isDirInMilestone.phaseCount > 0
  ? Math.max(phaseDirs.length, isDirInMilestone.phaseCount)
  : phaseDirs.length,
```

This is correct — `isDirInMilestone` is still in scope when the cache object is built. No functional regression here. ✓

### 2. Cache is never invalidated within a process that calls `cmdStateUpdate` + `cmdStateAdvancePlan` in sequence

This is the critical correctness defect.

`gsd-tools.cjs` exposes subcommands as a CLI dispatcher. Each invocation is a single short-lived process — **that claim is true for the current CLI entry point**. However, `buildStateFrontmatter` is also called from:

- `readStateMd` (line ~840) — used by `cmdStateLoad` and `cmdStateJson`
- `writeStateMd` (line ~948) — used by `cmdStateUpdate`, `cmdStateAdvancePlan`, `cmdStatePatch`

The SDK (`sdk/src/gsd-tools.ts`) calls `gsd-tools.cjs` as a subprocess for each operation, so per-invocation caching is safe there too.

**But**: the `gsd-tools.cjs` dispatcher is also `require()`-able as a module. If any test or internal caller imports it directly and calls multiple state-mutating commands in sequence within the same Node process, the cache will serve stale disk data after the first write. The test suite does exactly this via `runGsdTools()` — but `runGsdTools` spawns a subprocess each call, so it's safe. However, this is an implicit assumption that is nowhere documented and not enforced.

This is a latent defect with low current risk but real future risk as the codebase evolves. **It needs either a cache invalidation call in `writeStateMd`, or a comment in `writeStateMd` explicitly stating that callers must be single-invocation processes.**

### 3. Missing cache invalidation on write

`writeStateMd` calls `buildStateFrontmatter` to compute the frontmatter before writing. After the write, the on-disk state has changed (a new PLAN.md or SUMMARY.md may have been created by the executor agent before the next state write). The cache will return the pre-write disk snapshot for any subsequent `buildStateFrontmatter` call within the same process.

For the current CLI model this is fine. But `execute-phase` uses `STATE.md.lock`-based locking and writes STATE.md multiple times per wave. If two writes occur within the same process (possible in the Node SDK path or if the dispatcher is ever extended to handle compound operations), stale cache data will produce incorrect frontmatter metrics.

**Fix**: add `_diskScanCache.delete(cwd)` at the top of `writeStateMd` before calling `buildStateFrontmatter`. This is a one-liner that makes the cache safe for any calling pattern, not just the current one.

---

## Test coverage

The PR does not add any cache-specific tests. The test plan says:

> - [x] Cache is transparent — same output, fewer disk reads

"Same output" is not tested — there is no test that calls a state-mutating operation twice within the same process and asserts that the second call returns correct (post-mutation) data. The existing 93 tests all go through `runGsdTools()` subprocess isolation, so they cannot detect a stale-cache regression.

**Required**: at minimum one test that exercises two sequential state writes in the same process and asserts the metrics are correct after both writes.

---

## Summary

| Check | Result |
|---|---|
| Issue link + label | ✓ |
| Conventional commit prefix | ✓ |
| Security | ✓ |
| CJS, no external deps | ✓ |
| Cache key correctness (per-cwd) | ✓ |
| Per-process lifetime scoping | ✓ for CLI; latent risk for module-import callers |
| Stale cache on write | **Bug** — `writeStateMd` must invalidate the cache for the written `cwd` |
| Cache-specific tests | **Missing** — no test covers the stale-read-after-write scenario |

## Required before merge

1. Add `_diskScanCache.delete(cwd)` at the start of `writeStateMd` (before `buildStateFrontmatter` is called) to prevent stale reads if the function is ever called more than once per process for the same `cwd`.
2. Add a test that calls two state-mutating operations on the same `cwd` within a single process invocation and asserts that the second operation sees the updated disk state, not a cached snapshot from the first.

Tibsfox:
Thanks @trek-e! Both items addressed in 67aeb04:

**1. Cache invalidation in writeStateMd** — Added `_diskScanCache.delete(cwd)` at the top of `writeStateMd` before `buildStateFrontmatter` is called. This makes the cache safe for any calling pattern — SDK module imports, future compound dispatcher operations, and direct test imports of state.cjs.

**2. Stale-read-after-write test** — Added `tests/bug-1967-cache-invalidation.test.cjs` which imports state.cjs directly and exercises two sequential `writeStateMd` calls with a new phase directory created between them. Asserts the second write sees the updated disk state (`total_phases: 2, completed_phases: 1`) instead of the cached pre-write snapshot. This test would have caught the missing invalidation.

trek-e (approved):

PR #2059, opened 2026-04-10
feat(hooks): auto-record session state on context exhaustion

### LINKED ISSUE #1974: enhancement: auto-record session state on context exhaustion in context monitor hook
## Pre-submission checklist

Context exhaustion recovery — the context monitor hook already detects CRITICAL (25% remaining) but only injects advisory text.

## Current behavior

`gsd-context-monitor.js` fires a CRITICAL warning when remaining context drops below 25%. This warning tells the agent to stop and save state. However:

1. The agent may not act on the warning (it's advisory text, not a hard gate)
2. If the session crashes from context exhaustion, no state is saved
3. HANDOFF.json is only created by voluntary `/gsd-pause-work`, not on crash

## Proposed behavior

When the CRITICAL threshold fires, the hook automatically runs `gsd-tools state record-session` with a crash-recovery marker. This is lighter than full HANDOFF.json but provides enough breadcrumbs for `/gsd-resume-work` to reconstruct context:

- Records the current timestamp, phase, and plan
- Sets `Stopped At` to indicate context exhaustion (not voluntary pause)
- Runs as a fire-and-forget subprocess — doesn't block the agent

The advisory warning to the agent remains unchanged.

## Reason and benefit

Context exhaustion is the most common crash scenario in long sessions. Currently, all session state is lost unless the user manually ran `/gsd-pause-work` before the crash. This enhancement provides automatic breadcrumbs at the moment the system detects the session is about to end.

## Scope of changes

| File | Change |
|------|--------|
| `hooks/gsd-context-monitor.js` | On CRITICAL threshold, spawn `gsd-tools state record-session` subprocess (fire-and-forget) |

## Breaking changes

None. The advisory warning is unchanged. The `record-session` call is additive — it writes to STATE.md which already has session fields.

## Alternatives considered

1. **Write full HANDOFF.json** — rejected; too heavy for a hook that runs on every tool use. `record-session` is a single CLI call that updates existing STATE.md fields.
2. **Make the hook blocking** — rejected; hooks should be fast. Fire-and-forget subprocess is the right pattern.
3. **Rely on SessionEnd hooks** — rejected; SessionEnd may not fire on context exhaustion (the session doesn't end cleanly).

## Area affected

Hooks — context monitoring during agent execution.

Tibsfox:
## Summary

- On CRITICAL context threshold (25% remaining), auto-record `Stopped At: context exhaustion at N%` in STATE.md via fire-and-forget subprocess
- Provides breadcrumbs for `/gsd-resume-work` when sessions crash from context exhaustion
- Subprocess is `detached` + `unref()`'d — doesn't block the hook or agent
- Advisory warning text unchanged — this is purely additive

### Why

Context exhaustion is the most common crash scenario in long sessions. Previously, all session state was lost unless the user manually ran `/gsd-pause-work` before the crash. This records the minimum breadcrumb (timestamp + context percentage) at the moment the system detects the session is ending.

Closes #1974

## Test plan

- [x] Hook change only — subprocess spawns gsd-tools with existing `state update` command
- [x] Fire-and-forget pattern: spawn → detach → unref → exit hook immediately
- [ ] Manual: run a session until CRITICAL fires, verify STATE.md has `Stopped At` field

### DIFF
--- a/hooks/gsd-context-monitor.js
+++ b/hooks/gsd-context-monitor.js
@@ -128,6 +128,21 @@ process.stdin.on('end', () => {
     // Detect if GSD is active (has .planning/STATE.md in working directory)
     const isGsdActive = fs.existsSync(path.join(cwd, '.planning', 'STATE.md'));
 
+    // On CRITICAL with active GSD project, auto-record session state as a
+    // breadcrumb for /gsd-resume-work (#1974). Fire-and-forget subprocess —
+    // doesn't block the hook or the agent.
+    if (isCritical && isGsdActive) {
+      try {
+        const gsdTools = path.join(process.env.HOME || '', '.claude', 'get-shit-done', 'bin', 'gsd-tools.cjs');
+        const timestamp = new Date().toISOString().split('T')[0];
+        require('child_process').spawn(
+          process.execPath,
+          [gsdTools, 'state', 'update', 'Stopped At', `context exhaustion at ${usedPct}% (${timestamp})`],
+          { cwd, detached: true, stdio: 'ignore' }
+        ).unref();
+      } catch { /* non-critical — don't let state recording break the hook */ }
+    }
+
     // Build advisory warning message (never use imperative commands that
     // override user preferences — see #884)
     let message;
--- a/hooks/gsd-context-monitor.js
+++ b/hooks/gsd-context-monitor.js
@@ -21,6 +21,7 @@
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const { spawn } = require('child_process');
 
 const WARNING_THRESHOLD = 35;  // remaining_percentage <= 35%
 const CRITICAL_THRESHOLD = 25; // remaining_percentage <= 25%
@@ -130,16 +131,27 @@ process.stdin.on('end', () => {
 
     // On CRITICAL with active GSD project, auto-record session state as a
     // breadcrumb for /gsd-resume-work (#1974). Fire-and-forget subprocess —
-    // doesn't block the hook or the agent.
-    if (isCritical && isGsdActive) {
+    // doesn't block the hook or the agent. Fires ONCE per CRITICAL session,
+    // guarded by warnData.criticalRecorded to prevent repeated overwrites
+    // of the "crash moment" record on every debounce cycle.
+    if (isCritical && isGsdActive && !warnData.criticalRecorded) {
       try {
-        const gsdTools = path.join(process.env.HOME || '', '.claude', 'get-shit-done', 'bin', 'gsd-tools.cjs');
-        const timestamp = new Date().toISOString().split('T')[0];
-        require('child_process').spawn(
+        // Runtime-agnostic path: this hook lives at <runtime-config>/hooks/
+        // and gsd-tools.cjs lives at <runtime-config>/get-shit-done/bin/.
+        // Using __dirname makes this work on Claude Code, OpenCode, Gemini,
+        // Kilo, etc. without hardcoding ~/.claude/.
+        const gsdTools = path.join(__dirname, '..', 'get-shit-done', 'bin', 'gsd-tools.cjs');
+        // Coerce usedPct to a safe number in case bridge file is malformed
+        const safeUsedPct = Number(usedPct) || 0;
+        const stoppedAt = `context exhaustion at ${safeUsedPct}% (${new Date().toISOString().split('T')[0]})`;
+        spawn(
           process.execPath,
-          [gsdTools, 'state', 'update', 'Stopped At', `context exhaustion at ${usedPct}% (${timestamp})`],
+          [gsdTools, 'state', 'record-session', '--stopped-at', stoppedAt],
           { cwd, detached: true, stdio: 'ignore' }
         ).unref();
+        warnData.criticalRecorded = true;
+        // Persist the sentinel so subsequent debounce cycles don't re-fire
+        fs.writeFileSync(warnPath, JSON.stringify(warnData));
       } catch { /* non-critical — don't let state recording break the hook */ }
     }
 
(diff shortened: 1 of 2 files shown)


trek-e (changes requested):
## Review: request-changes

Issue link confirmed: #1974 carries `approved-enhancement`. The approach (fire-and-forget subprocess on CRITICAL) is sound in concept. Three blocking defects before this can merge.

---

### Blocking — Repeated firing on every debounce cycle

The new block is placed after the debounce gate:

```js
// Reset debounce counter
warnData.callsSinceWarn = 0;
warnData.lastLevel = currentLevel;
fs.writeFileSync(warnPath, JSON.stringify(warnData));

// ... then later ...
if (isCritical && isGsdActive) {
  // subprocess spawned here — fires every DEBOUNCE_CALLS tool uses
}
```

The approved spec says the subprocess is a one-time breadcrumb. As written it fires repeatedly: once when CRITICAL is first reached, then again every `DEBOUNCE_CALLS` (5) tool uses for the rest of the session. The STATE.md `Stopped At` field gets overwritten on every cycle with a new timestamp, which corrupts the "crash moment" record.

Fix: add a sentinel in the warn state file to track whether the subprocess has already been spawned, and guard with `!warnData.criticalRecorded`. Set `warnData.criticalRecorded = true` before `writeFileSync`.

---

### Blocking — `gsd-tools.cjs` path is Claude-only

```js
const gsdTools = path.join(process.env.HOME || '', '.claude', 'get-shit-done', 'bin', 'gsd-tools.cjs');
```

This hook is installed for all runtimes (OpenCode → `~/.config/opencode/`, Gemini → `~/.gemini/`, Kilo → `~/.config/kilo/`). Hard-coding `~/.claude/` means the path will be wrong on all non-Claude installations. The installer already has a precedent for `CLAUDE_CONFIG_DIR` — use that env var as the override path, and fall back to the runtime-specific path written by the installer (the installer already rewrites `~/.claude/` path references at install time via `path.replacement` transforms).

The simplest correct fix: use `__dirname` — this hook is installed at `<runtime-config>/hooks/gsd-context-monitor.js`, and `gsd-tools.cjs` is at `<runtime-config>/get-shit-done/bin/gsd-tools.cjs`. So `path.join(__dirname, '..', 'get-shit-done', 'bin', 'gsd-tools.cjs')` resolves correctly regardless of runtime.

---

### Blocking — `state update` vs `state record-session`

Issue #1974 describes the command as `gsd-tools state record-session`. The implementation calls:

```js
[gsdTools, 'state', 'update', 'Stopped At', `context exhaustion at ${usedPct}% (${timestamp})`]
```

Confirm `state update` accepts a key/value pair in this positional form and writes the expected field to STATE.md. If `state update` is not the right subcommand (or takes a different signature), the subprocess will silently exit non-zero and nothing will be written. The outer `catch {}` swallows the error, so this will be invisible in production.

Add a test (see below) that exercises this path against a real temp STATE.md and asserts the field is written correctly.

---

### Non-blocking — `usedPct` should be coerced to a number

`usedPct` comes from `metrics.used_pct` (the temp bridge JSON file). `spawn` uses an array of args so there is no shell injection risk. However, if the metrics file is malformed or adversarially crafted, `usedPct` could be a string like `"foo"` or `"75; rm -rf ~"`. While not exploitable via `spawn`, it would write garbage into STATE.md. Add `const safeUsedPct = Number(usedPct) || 0` and use that in the template string.

---

### Non-blocking — inline `require` inconsistent with file style

```js
require('child_process').spawn(...)
```

All other `require` calls in this file are hoisted to the top. Extract to the top-level:

```js
const { spawn } = require('child_process');
```

---

### Missing tests

CONTRIBUTING.md requires tests for behavioral changes. The existing `hooks-opt-in.test.cjs` pattern (pipe JSON into the hook via `spawnSync`) can be adapted:

1. A test that verifies: when CRITICAL threshold fires with an active GSD project, the subprocess is spawned and STATE.md receives the `Stopped At` field.
2. A test that verifies: the subprocess does NOT fire a second time within the same debounce window (once the sentinel is set).
3. A test that verifies: when `isGsdActive` is false (no STATE.md), the subprocess is not spawned.

These can be written against the actual hook using the bridge metrics file written to `os.tmpdir()` — the same approach `hooks-opt-in.test.cjs` uses for the bash hooks, adapted for the JS context monitor hook.

Tibsfox:
Thanks @trek-e! All three blocking defects addressed in 7be9aff, plus the two style nits and new tests:

**1. Repeated firing** — Added `warnData.criticalRecorded` sentinel persisted to the warn state file. Fires exactly once per CRITICAL session instead of every 5 tool uses.

**2. Hardcoded ~/.claude/ path** — Replaced with `path.join(__dirname, '..', 'get-shit-done', 'bin', 'gsd-tools.cjs')`. The hook at `<runtime-config>/hooks/` resolves correctly to `<runtime-config>/get-shit-done/bin/` on all runtimes (Claude Code, OpenCode, Gemini, Kilo) without assuming a specific config dir.

**3. Wrong subcommand** — Switched from `state update` to `state record-session --stopped-at`. The dedicated command updates Last session, Last Date, Stopped At, and Resume File atomically under the state lock, which is exactly what the original issue proposed.

**Style nits** — Hoisted `const { spawn } = require('child_process')` to top of file. Coerced `usedPct` to `Number(usedPct) || 0` to sanitize the bridge metrics file.

**Tests** — Added `tests/bug-1974-context-exhaustion-record.test.cjs` (4 cases):
1. Subprocess spawns and writes "context exhaustion" to STATE.md on CRITICAL
2. Subprocess does NOT spawn when `.planning/STATE.md` is absent
3. Sentinel guard prevents second fire within same session
4. Hook source uses `__dirname`-based path (not hardcoded `~/.claude/`)

Test #1 actually exercises the real subprocess end-to-end via `spawnSync` and verifies STATE.md contains the "context exhaustion" field — this would have caught the missing `state update` signature bug.

trek-e (approved):

PR #2061, opened 2026-04-10
perf(workflow): default to inline execution for 1-2 task plans

### LINKED ISSUE #1979: enhancement: default to inline execution for 1-2 task plans to avoid subagent spawn overhead
## Pre-submission checklist

Plan execution dispatch in `execute-phase.md` — the decision between spawning a subagent (Pattern A/B) vs. inline execution (Pattern C).

## Current behavior

`execute-phase.md` supports three execution patterns:
- **Pattern A**: Spawn subagent per plan (default for multi-plan phases)
- **Pattern B**: Sequential inline with context carry (for worktree or when Task unavailable)
- **Pattern C**: Direct inline execution (currently used only when Task tool is unavailable)

Every `Task()` spawn creates a fresh conversation, which kills all prompt cache prefix from the parent context. For single-task plans, the overhead of spawning a new agent (~14,000 tokens of static content) outweighs the benefit of isolation.

Pattern C already exists and works correctly — it just isn't the default for small plans.

## Proposed behavior

For plans with 1-2 tasks, default to Pattern C (inline execution within the orchestrator's context) rather than spawning a subagent. This preserves the orchestrator's cache prefix and avoids the 14,000-token static overhead. Reserve subagent spawning (Pattern A) for plans with 3+ tasks where the fresh context benefit outweighs the cache loss.

The threshold could be driven by `workflow.inline_plan_threshold` config (default: 2) for user customization.

## Reason and benefit

Saves ~14,000 tokens per small plan executed inline. Small plans (1-2 tasks) are common in gap-closure phases, cleanup phases, and phases with many fine-grained plans. The Anthropic prompt caching documentation confirms cache is prefix-based and per-conversation — each subagent spawn starts cold.

## Scope of changes

| File | Change |
|------|--------|
| `get-shit-done/workflows/execute-phase.md` | Add plan-count check before dispatch: if tasks <= 2, use Pattern C; otherwise Pattern A |
| `get-shit-done/references/planning-config.md` | Document `workflow.inline_plan_threshold` config key |

## Breaking changes

None. Pattern C already exists and produces identical results. The change only affects which dispatch path is chosen by default.

## Alternatives considered

1. **Always use subagents** — current behavior; wastes ~14K tokens per small plan
2. **Make it opt-in via config** — could layer `workflow.inline_plan_threshold` on top, but a reasonable default of 2 serves most projects
3. **Always use inline** — rejected; large plans benefit from subagent isolation (fresh context, worktree support)

## Area affected

Core workflow (execute-phase) — plan dispatch logic.

Tibsfox:
## Summary

- Plans with 1-2 tasks now execute inline (Pattern C) instead of spawning a subagent (Pattern A)
- Avoids ~14K token subagent spawn overhead and preserves the orchestrator's prompt cache
- Configurable via \`workflow.inline_plan_threshold\` (default: 2, set to 0 to disable)
- Plans above the threshold continue using checkpoint-based routing as before

### Why

Small plans (gap closure, cleanup, single-task phases) pay the full subagent spawn cost for minimal work. Pattern C already exists and produces identical results — this just makes it the default for small plans.

### Changes

| File | Change |
|------|--------|
| \`execute-plan.md\` | Add task count check before checkpoint routing |
| \`planning-config.md\` | Document \`workflow.inline_plan_threshold\` in both tables |
| \`config.cjs\` | Add to VALID_CONFIG_KEYS |

Closes #1979

## Test plan

- [x] Full test suite passes (2918/2919, 1 pre-existing)
- [x] Pattern C already exists and is well-tested — this just routes more plans to it
- [ ] Manual: run \`/gsd-execute-phase\` on a 1-task plan, verify inline execution

### DIFF
--- a/get-shit-done/bin/lib/config.cjs
+++ b/get-shit-done/bin/lib/config.cjs
@@ -28,6 +28,7 @@ const VALID_CONFIG_KEYS = new Set([
   'git.branching_strategy', 'git.base_branch', 'git.phase_branch_template', 'git.milestone_branch_template', 'git.quick_branch_template',
   'planning.commit_docs', 'planning.search_gitignored',
   'workflow.subagent_timeout',
+  'workflow.inline_plan_threshold',
   'hooks.context_warnings',
   'features.thinking_partner',
   'context',
--- a/get-shit-done/references/planning-config.md
+++ b/get-shit-done/references/planning-config.md
@@ -35,6 +35,7 @@ Configuration options for `.planning/` directory behavior.
 | `git.quick_branch_template` | `null` | Optional branch template for quick-task runs |
 | `workflow.use_worktrees` | `true` | Whether executor agents run in isolated git worktrees. Set to `false` to disable worktrees — agents execute sequentially on the main working tree instead. Recommended for solo developers or when worktree merges cause issues. |
 | `workflow.subagent_timeout` | `300000` | Timeout in milliseconds for parallel subagent tasks (e.g. codebase mapping). Increase for large codebases or slower models. Default: 300000 (5 minutes). |
+| `workflow.inline_plan_threshold` | `2` | Plans with this many tasks or fewer execute inline (Pattern C) instead of spawning a subagent. Avoids ~14K token spawn overhead for small plans. Set to `0` to always spawn subagents. |
 | `manager.flags.discuss` | `""` | Flags passed to `/gsd-discuss-phase` when dispatched from manager (e.g. `"--auto --analyze"`) |
 | `manager.flags.plan` | `""` | Flags passed to plan workflow when dispatched from manager |
 | `manager.flags.execute` | `""` | Flags passed to execute workflow when dispatched from manager |
@@ -259,6 +260,7 @@ Set via `workflow.*` namespace in config.json (e.g., `"workflow": { "research": 
 | `workflow.skip_discuss` | boolean | `false` | `true`, `false` | Skip discuss phase entirely |
 | `workflow.use_worktrees` | boolean | `true` | `true`, `false` | Run executor agents in isolated git worktrees |
 | `workflow.subagent_timeout` | number | `300000` | Any positive integer (ms) | Timeout for parallel subagent tasks (default: 5 minutes) |
+| `workflow.inline_plan_threshold` | number | `2` | `0`–`10` | Plans with ≤N tasks execute inline instead of spawning a subagent |
 | `workflow.code_review` | boolean | `true` | `true`, `false` | Enable built-in code review step in the ship workflow |
 | `workflow.code_review_depth` | string | `"standard"` | `"light"`, `"standard"`, `"deep"` | Depth level for code review analysis in the ship workflow |
 | `workflow._auto_chain_active` | boolean | `false` | `true`, `false` | Internal: tracks whether autonomous chaining is active |
--- a/get-shit-done/workflows/execute-plan.md
+++ b/get-shit-done/workflows/execute-plan.md
@@ -61,10 +61,18 @@ PLAN_START_EPOCH=$(date +%s)
 
 <step name="parse_segments">
 ```bash
+TASK_COUNT=$(grep -c "^<task" .planning/phases/XX-name/{phase}-{plan}-PLAN.md 2>/dev/null || echo "0")
+INLINE_THRESHOLD=$(node "$HOME/.claude/get-shit-done/bin/gsd-tools.cjs" config-get workflow.inline_plan_threshold --default 2 2>/dev/null || echo "2")
 grep -n "type=\"checkpoint" .planning/phases/XX-name/{phase}-{plan}-PLAN.md
 ```
 
-**Routing by checkpoint type:**
+**Primary routing: task count threshold (#1979)**
+
+If `TASK_COUNT <= INLINE_THRESHOLD`: Use Pattern C (inline) regardless of checkpoint type. Small plans execute faster inline — avoids ~14K token subagent spawn overhead and preserves prompt cache. Configure threshold via `workflow.inline_plan_threshold` (default: 2).
+
+If `TASK_COUNT > INLINE_THRESHOLD`: Apply checkpoint-based routing below.
+
+**Checkpoint-based routing (plans with > threshold tasks):**
 
 | Checkpoints | Pattern | Execution |
 |-------------|---------|-----------|
--- a/get-shit-done/workflows/execute-plan.md
+++ b/get-shit-done/workflows/execute-plan.md
@@ -61,16 +61,17 @@ PLAN_START_EPOCH=$(date +%s)
 
 <step name="parse_segments">
 ```bash
-TASK_COUNT=$(grep -c "^<task" .planning/phases/XX-name/{phase}-{plan}-PLAN.md 2>/dev/null || echo "0")
+# Count tasks — match <task tag at any indentation level
+TASK_COUNT=$(grep -cE '^\s*<task[[:space:]>]' .planning/phases/XX-name/{phase}-{plan}-PLAN.md 2>/dev/null || echo "0")
 INLINE_THRESHOLD=$(node "$HOME/.claude/get-shit-done/bin/gsd-tools.cjs" config-get workflow.inline_plan_threshold --default 2 2>/dev/null || echo "2")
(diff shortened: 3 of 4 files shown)


trek-e (changes requested):
## Review

### Issue link
`Closes #1979` is present. Issue carries `approved-enhancement`. ✅

### Logic — `TASK_COUNT` grep pattern
The new routing check uses:
```bash
TASK_COUNT=$(grep -c "^<task" .planning/phases/XX-name/{phase}-{plan}-PLAN.md 2>/dev/null || echo "0")
```
The PLAN.md template and `gsd-planner.md` both show task elements indented inside a `<tasks>` block (e.g. `<task type="auto">`). In practice those lines are indented with spaces/tabs, so `^<task` (anchored to start-of-line) will **not match any tasks** and `TASK_COUNT` will always be `0`. That means `0 <= threshold (2)` is always true — every plan routes inline regardless of actual task count. The count should be `grep -c "<task "` (un-anchored, matches anywhere on the line), or anchor on the known indentation if it is always consistent.

### `threshold = 0` handling
The PR description and the config doc both state "Set to `0` to disable inline execution." The workflow comment `If TASK_COUNT <= INLINE_THRESHOLD` means `0 <= 0` is true whenever a plan has no tasks, and false otherwise. That is correct for the edge case, but the intent ("set to 0 to always spawn subagents") is not guarded — a plan that genuinely has 0 tasks (malformed or empty) would still route inline even when the user has disabled the feature. The routing block needs an explicit `if [[ "$INLINE_THRESHOLD" -gt 0 && "$TASK_COUNT" -le "$INLINE_THRESHOLD" ]]` guard. ✗

### Pattern C output compatibility
Pattern C is the existing "execute in main context" path — it already produces SUMMARY.md and commits as required. No compatibility issue here. ✅

### `config.cjs` — `VALID_CONFIG_KEYS`
`workflow.inline_plan_threshold` is added to the set. ✅

### `planning-config.md` documentation
Added to both tables (Quick Reference and Complete Field Reference). ✅

### Test coverage
The `config-field-docs.test.cjs` suite validates that every field in `VALID_CONFIG_KEYS` appears in `planning-config.md`, but it does **not** enumerate `workflow.inline_plan_threshold` specifically. Because the new key was added to both `VALID_CONFIG_KEYS` and `planning-config.md`, no existing test fails — and no new test was added.

Two gaps:

1. **No config validation test** for `workflow.inline_plan_threshold` (valid type, range clamping, `0` as a legal value, rejection of negative values). `tests/config.test.cjs` tests `config-set` and `config-get` for other keys; this key should be covered the same way.
2. **No routing-logic test** for the threshold check in `execute-plan.md`. The workflow is a `.md` file so it cannot be unit-tested directly, but the task-count grep (`grep -c "^<task"`) is testable in isolation (and exposes the anchor bug above). At minimum a test for the bash snippet in a fixture PLAN.md would catch the regression.

The 2918/2919 pass rate is consistent with these gaps — the pre-existing failure is unrelated, but the new feature path has zero coverage. ✗

### Summary of required changes

| # | File | Required fix |
|---|------|-------------|
| 1 | `execute-plan.md` | Fix `grep -c "^<task"` → `grep -c "<task "` (remove `^` anchor; tasks are indented in real PLAN.md files) |
| 2 | `execute-plan.md` | Guard inline routing with `[[ "$INLINE_THRESHOLD" -gt 0 && … ]]` so threshold=0 reliably disables the feature |
| 3 | `tests/config.test.cjs` | Add `config-set workflow.inline_plan_threshold 3` / `config-get` round-trip test |

Items 1 and 2 are correctness bugs. Item 3 is a test coverage gap required by the contributing guide for enhancements ("Tests covering the enhanced behavior are required").

trek-e (changes requested):
## Review

### Issue link
\`Closes #1979\` is present. Issue carries \`approved-enhancement\`. ✅

### Logic — \`TASK_COUNT\` grep pattern
The new routing check uses:

\`\`\`bash
TASK_COUNT=$(grep -c "^<task" .planning/phases/XX-name/{phase}-{plan}-PLAN.md 2>/dev/null || echo "0")
\`\`\`

The PLAN.md template and \`gsd-planner.md\` both show task elements indented inside a \`<tasks>\` block (e.g. \`    <task type="auto">\`). In practice those lines are indented with spaces/tabs, so \`^<task\` (anchored to start-of-line) will **not match any tasks** and \`TASK_COUNT\` will always be \`0\`. That means \`0 <= threshold (2)\` is always true — every plan routes inline regardless of actual task count.

Fix: use \`grep -c "<task "\` (un-anchored).

### \`threshold = 0\` handling
The config doc and PR summary both state "Set to \`0\` to disable inline execution." The condition \`TASK_COUNT <= INLINE_THRESHOLD\` evaluates \`0 <= 0\` as true for empty/malformed plans even when the user has disabled the feature. The routing block needs an explicit \`[[ "$INLINE_THRESHOLD" -gt 0 && "$TASK_COUNT" -le "$INLINE_THRESHOLD" ]]\` guard so zero means "always spawn." ✗

### Pattern C output compatibility
Pattern C is the existing "execute in main context" path — it already produces SUMMARY.md and commits. No compatibility issue. ✅

### \`config.cjs\` — \`VALID_CONFIG_KEYS\`
\`workflow.inline_plan_threshold\` is added to the set. ✅

### \`planning-config.md\` documentation
Added to both tables (Quick Reference and Complete Field Reference). ✅

### Test coverage
Two gaps, both required by CONTRIBUTING.md for enhancements:

1. **No config test** for \`workflow.inline_plan_threshold\` — \`tests/config.test.cjs\` covers other keys via \`config-set\`/\`config-get\` round-trips; this key should be covered the same way (valid value, \`0\` as legal value).
2. **No routing test** — the task-count grep is testable in isolation and would have caught the \`^<task\` anchor bug. A fixture PLAN.md with known task count, run through the grep, is sufficient.

The 2918/2919 pass rate is consistent with zero new tests for this feature. ✗

### Required changes

| # | File | Fix |
|---|------|-----|
| 1 | \`execute-plan.md\` | \`grep -c "^<task"\` → \`grep -c "<task "\` (remove \`^\` anchor; tasks are indented in real PLAN.md files) |
| 2 | \`execute-plan.md\` | Guard condition: \`[[ "$INLINE_THRESHOLD" -gt 0 && "$TASK_COUNT" -le "$INLINE_THRESHOLD" ]]\` so threshold=0 reliably disables inline routing |
| 3 | \`tests/config.test.cjs\` | Add \`config-set\`/\`config-get\` round-trip test for \`workflow.inline_plan_threshold\` |

Items 1 and 2 are correctness bugs (1 silently routes every plan inline; 2 mishandles the documented opt-out). Item 3 is a test gap required by the contributing guide.

trek-e (changes requested):
Review: Issue #1979 carries approved-enhancement. Config.cjs and planning-config.md changes are correct. Two correctness bugs in execute-plan.md: (1) grep -c "^<task" anchors to start-of-line but tasks are indented in real PLAN.md files, so TASK_COUNT is always 0 and every plan routes inline — fix: remove the ^ anchor. (2) threshold=0 is documented to disable inline routing but the condition TASK_COUNT <= INLINE_THRESHOLD evaluates 0<=0 as true for empty plans even when the feature is disabled — fix: guard with INLINE_THRESHOLD -gt 0. Missing test coverage for both the new config key and the routing logic, required by CONTRIBUTING.md for enhancements.

Tibsfox:
Thanks @trek-e! Excellent catches — both bugs fixed in e5cc0bb:

**1. Grep anchor bug** — Changed from `grep -c "^<task"` to `grep -cE '^\s*<task[[:space:]>]'`. This matches tasks at any indentation level and avoids false positives on `<tasks>` or `</task>`. Verified against both the existing template (`phase-prompt.md` — unindented) and indented task formats.

**2. Threshold=0 guard** — Added explicit `INLINE_THRESHOLD > 0` guard so setting the key to 0 actually disables inline routing instead of evaluating `0 <= 0` as true for empty plans.

**3. Missing tests** — Added `tests/inline-plan-threshold.test.cjs` with 8 tests covering:
- config-set accepts the key and threshold=0
- VALID_CONFIG_KEYS and planning-config.md contain the entry
- Routing pattern matches indented tasks and rejects `<tasks>`/`</task>`
- Routing is guarded by `INLINE_THRESHOLD > 0`

trek-e (approved):

PR #2062, opened 2026-04-10
feat(config): support global skills from ~/.claude/skills/ in agent_skills

### LINKED ISSUE #1992: agent_skills config should support global skills from ~/.claude/skills/
## Problem

`agent_skills` in `.planning/config.json` only resolves skill paths relative to the project root via `validatePath()` in `security.cjs`. Skills installed globally at `~/.claude/skills/` (e.g., `shadcn`, `frontend-design`, `supabase-postgres-best-practices`) cannot be injected into sub-agents because the security guard rejects any path that escapes the project directory.

**Current behavior:**

```json
"agent_skills": {
  "gsd-executor": ["shadcn"]
}
```

`buildAgentSkillsBlock()` looks for `<projectRoot>/shadcn/SKILL.md` → not found → skipped.

Symlinks don't work either because `fs.realpathSync()` resolves them to the real path outside the project, which `validatePath` then rejects.

**Code reference:** `bin/lib/init.cjs`, `buildAgentSkillsBlock()` (lines ~1443-1480).

## Use Case

Injecting third-party or user-installed global skills into GSD sub-agents during execution. For example:

- `shadcn` → `gsd-executor`, `gsd-ui-researcher` (component API knowledge while coding UI)
- `frontend-design` → `gsd-ui-researcher` (design patterns for UI spec generation)
- `supabase-postgres-best-practices` → `gsd-executor`, `gsd-planner` (query optimization awareness)

These skills are installed globally via shadcn CLI or copy-paste and live at `~/.claude/skills/<name>/SKILL.md`. They are not project-specific, so duplicating them into every repo is not practical.

## Proposed Solution

Support a `global:` prefix that resolves to the runtime's global skills directory:

```json
"agent_skills": {
  "gsd-executor": ["global:shadcn", "global:supabase-postgres-best-practices"],
  "gsd-ui-researcher": ["global:frontend-design", "global:shadcn"]
}
```

Resolution logic in `buildAgentSkillsBlock()`:

```
if skillPath starts with "global:"
  → resolve to ~/.claude/skills/<name>/SKILL.md (or equivalent runtime dir)
  → skip validatePath (already within the runtime's controlled directory)
else
  → current behavior (relative to project root, validatePath enforced)
```

The global skills directory is already a trusted location managed by the runtime (Claude Code installs GSD skills there), so the security concern of path traversal doesn't apply.

## Alternatives Considered

- **Copy skills into each project** — works but duplicates files and desynchronizes on updates.
- **Symlinks** — rejected by `validatePath` because `realpathSync` resolves outside project root.
- **Inheriting skills from parent session** — blocked by Claude Code platform limitation (see #382).

## Environment

- GSD v1.34.2
- Claude Code on Windows 11
- Skills installed at `~/.claude/skills/` (global)

Tibsfox:
## Summary

- Add \`global:\` prefix for \`agent_skills\` config entries that resolve to \`~/.claude/skills/<name>/SKILL.md\`
- Allows injecting globally-installed skills (shadcn, supabase, frontend-design, etc.) into GSD sub-agents without duplicating them into every project
- Project-relative paths continue to use \`validatePath()\` containment checks as before

### Usage

\`\`\`json
"agent_skills": {
  "gsd-executor": ["global:shadcn", "global:supabase-postgres-best-practices"],
  "gsd-ui-researcher": ["global:frontend-design", "global:shadcn"]
}
\`\`\`

### Security

- Skill names validated against \`/^[a-zA-Z0-9_-]+$/\` — prevents path traversal
- \`~/.claude/skills/\` is a trusted runtime-controlled directory (where Claude Code installs skills)
- No changes to \`validatePath()\` or \`security.cjs\` — global skills bypass it because the target directory is already trusted
- Invalid names and missing SKILL.md files produce warnings, not errors

Closes #1992

## Test plan

- [x] All 11 agent-skills tests pass
- [x] Existing project-relative skill paths unchanged
- [ ] Manual: add \`"global:shadcn"\` to agent_skills config, verify it resolves to ~/.claude/skills/shadcn/SKILL.md

### DIFF
--- a/get-shit-done/bin/lib/init.cjs
+++ b/get-shit-done/bin/lib/init.cjs
@@ -1470,6 +1470,24 @@ function buildAgentSkillsBlock(config, agentType, projectRoot) {
   for (const skillPath of skillPaths) {
     if (typeof skillPath !== 'string') continue;
 
+    // Support global: prefix for skills installed at ~/.claude/skills/ (#1992)
+    if (skillPath.startsWith('global:')) {
+      const skillName = skillPath.slice(7);
+      // Sanitize: skill name must be alphanumeric, hyphens, or underscores only
+      if (!/^[a-zA-Z0-9_-]+$/.test(skillName)) {
+        process.stderr.write(`[agent-skills] WARNING: Invalid global skill name "${skillName}" — skipping\n`);
+        continue;
+      }
+      const globalSkillDir = path.join(require('os').homedir(), '.claude', 'skills', skillName);
+      const globalSkillMd = path.join(globalSkillDir, 'SKILL.md');
+      if (!fs.existsSync(globalSkillMd)) {
+        process.stderr.write(`[agent-skills] WARNING: Global skill not found at "~/.claude/skills/${skillName}/SKILL.md" — skipping\n`);
+        continue;
+      }
+      validPaths.push({ ref: `${globalSkillDir}/SKILL.md`, display: `~/.claude/skills/${skillName}` });
+      continue;
+    }
+
     // Validate path safety — must resolve within project root
     const pathCheck = validatePath(skillPath, projectRoot);
     if (!pathCheck.safe) {
@@ -1484,12 +1502,12 @@ function buildAgentSkillsBlock(config, agentType, projectRoot) {
       continue;
     }
 
-    validPaths.push(skillPath);
+    validPaths.push({ ref: `${skillPath}/SKILL.md`, display: skillPath });
   }
 
   if (validPaths.length === 0) return '';
 
-  const lines = validPaths.map(p => `- @${p}/SKILL.md`).join('\n');
+  const lines = validPaths.map(p => `- @${p.ref}`).join('\n');
   return `<agent_skills>\nRead these user-configured skills:\n${lines}\n</agent_skills>`;
 }
 
--- a/get-shit-done/bin/lib/init.cjs
+++ b/get-shit-done/bin/lib/init.cjs
@@ -1456,6 +1456,8 @@ function cmdInitRemoveWorkspace(cwd, name, raw) {
  */
 function buildAgentSkillsBlock(config, agentType, projectRoot) {
   const { validatePath } = require('./security.cjs');
+  const os = require('os');
+  const globalSkillsBase = path.join(os.homedir(), '.claude', 'skills');
 
   if (!config || !config.agent_skills || !agentType) return '';
 
@@ -1473,15 +1475,28 @@ function buildAgentSkillsBlock(config, agentType, projectRoot) {
     // Support global: prefix for skills installed at ~/.claude/skills/ (#1992)
     if (skillPath.startsWith('global:')) {
       const skillName = skillPath.slice(7);
+      // Explicit empty-name guard before regex for clearer error message
+      if (!skillName) {
+        process.stderr.write(`[agent-skills] WARNING: "global:" prefix with empty skill name — skipping\n`);
+        continue;
+      }
       // Sanitize: skill name must be alphanumeric, hyphens, or underscores only
       if (!/^[a-zA-Z0-9_-]+$/.test(skillName)) {
         process.stderr.write(`[agent-skills] WARNING: Invalid global skill name "${skillName}" — skipping\n`);
         continue;
       }
-      const globalSkillDir = path.join(require('os').homedir(), '.claude', 'skills', skillName);
+      const globalSkillDir = path.join(globalSkillsBase, skillName);
       const globalSkillMd = path.join(globalSkillDir, 'SKILL.md');
       if (!fs.existsSync(globalSkillMd)) {
         process.stderr.write(`[agent-skills] WARNING: Global skill not found at "~/.claude/skills/${skillName}/SKILL.md" — skipping\n`);
+        continue;
+      }
+      // Symlink escape guard: validatePath resolves symlinks and enforces
+      // containment within globalSkillsBase. Prevents a skill directory
+      // symlinked to an arbitrary location from being injected (#1992).
+      const pathCheck = validatePath(globalSkillMd, globalSkillsBase, { allowAbsolute: true });
+      if (!pathCheck.safe) {
+        process.stderr.write(`[agent-skills] WARNING: Global skill "${skillName}" failed path check (symlink escape?) — skipping\n`);
         continue;
       }
       validPaths.push({ ref: `${globalSkillDir}/SKILL.md`, display: `~/.claude/skills/${skillName}` });
--- a/tests/agent-skills.test.cjs
+++ b/tests/agent-skills.test.cjs
@@ -205,3 +205,90 @@ describe('config-set agent_skills', () => {
     );
   });
 });
+
+// ─── global: prefix support (#1992) ──────────────────────────────────────────
+
+describe('agent-skills global: prefix', () => {
+  let tmpDir;
+  let fakeHome;
+  let globalSkillsDir;
+
+  beforeEach(() => {
+    tmpDir = createTempProject();
+    // Create a fake HOME with ~/.claude/skills/ structure
+    fakeHome = fs.mkdtempSync(path.join(require('os').tmpdir(), 'gsd-1992-home-'));
+    globalSkillsDir = path.join(fakeHome, '.claude', 'skills');
+    fs.mkdirSync(globalSkillsDir, { recursive: true });
+  });
+
+  afterEach(() => {
+    cleanup(tmpDir);
+    fs.rmSync(fakeHome, { recursive: true, force: true });
+  });
+
+  function createGlobalSkill(name) {
+    const skillDir = path.join(globalSkillsDir, name);
+    fs.mkdirSync(skillDir, { recursive: true });
+    fs.writeFileSync(path.join(skillDir, 'SKILL.md'), `# ${name}\nGlobal skill content.\n`);
+    return skillDir;
+  }
+
+  test('global:valid-skill resolves to $HOME/.claude/skills/valid-skill/SKILL.md', () => {
+    createGlobalSkill('valid-skill');
+    writeConfig(tmpDir, {
+      agent_skills: { 'gsd-executor': ['global:valid-skill'] },
+    });
+
+    const result = runGsdTools(['agent-skills', 'gsd-executor'], tmpDir, { HOME: fakeHome, USERPROFILE: fakeHome });
+    assert.ok(result.output.includes('valid-skill/SKILL.md'), `should reference the global skill: ${result.output}`);
+    assert.ok(result.output.includes('<agent_skills>'), 'should emit agent_skills block');
+  });
+
+  test('global:invalid!name is rejected by regex and skipped', () => {
+    writeConfig(tmpDir, {
+      agent_skills: { 'gsd-executor': ['global:invalid!name'] },
+    });
+
+    const result = runGsdTools(['agent-skills', 'gsd-executor'], tmpDir, { HOME: fakeHome, USERPROFILE: fakeHome });
+    // No valid skills → empty output, command succeeds
+    assert.strictEqual(result.output, '', 'should skip invalid name without crashing');
+  });
+
+  test('global:missing-skill is skipped when directory is absent', () => {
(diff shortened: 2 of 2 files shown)


trek-e (changes requested):
## Review: request-changes

Linked issue #1992 carries `approved-enhancement` — the issue link is valid. The core approach (prefix-based dispatch into a separate code path) is sound. However there are three blocking issues and one style fix required before this can merge.

---

### [BLOCKING] Symlink escape is unguarded

The regex `/^[a-zA-Z0-9_-]+$/` blocks `../` traversal in the name itself, but it cannot prevent a skill _directory_ that is itself a symlink pointing outside `~/.claude/skills/`. The existing `validatePath()` in `security.cjs` is specifically designed for this: it calls `fs.realpathSync()` on both the base and the target, then checks containment. The new global path skips that entirely.

**Fix:** after constructing `globalSkillDir`, call `validatePath(globalSkillMd, path.join(require('os').homedir(), '.claude', 'skills'))` (with `allowAbsolute: true`) and skip on failure. That re-uses the existing symlink-resolution logic instead of reimplementing it partially.

```js
const globalSkillsBase = path.join(require('os').homedir(), '.claude', 'skills');
const pathCheck = validatePath(globalSkillMd, globalSkillsBase, { allowAbsolute: true });
if (!pathCheck.safe) {
  process.stderr.write(`[agent-skills] WARNING: Global skill "${skillName}" failed path check — skipping\n`);
  continue;
}
```

---

### [BLOCKING] No tests for the `global:` prefix path

The diff adds zero test coverage. The PR description mentions "11 agent-skills tests pass" — those are the pre-existing tests; none cover the new code path. Per CONTRIBUTING.md, an enhancement requires tests covering the enhanced behavior. The following cases must be covered in `tests/agent-skills.test.cjs`:

1. `global:valid-skill` resolves to `$HOME/.claude/skills/valid-skill/SKILL.md` and appears in output
2. `global:invalid!name` is rejected by the regex (skipped, no crash)
3. `global:missing-skill` where the directory doesn't exist → skipped, command still succeeds
4. Mix of `global:` and project-relative paths in the same array → both resolve correctly
5. `global:` with an empty name after the prefix (`skillPath === 'global:'`, `skillName === ''`) → rejected by regex

Note: tests must set `HOME` / `USERPROFILE` env vars (as the existing tests already do) to point at a controlled temp dir so the global skills base is deterministic and isolated.

---

### [BLOCKING] Empty `global:` prefix is silently mishandled

`'global:'.slice(7)` produces `''`. The regex `/^[a-zA-Z0-9_-]+$/` requires at least one character, so `''` fails the test and the skill is skipped with a confusing message: `Invalid global skill name "" — skipping`. This is the correct outcome but it's worth adding a guard with a clearer message:

```js
if (!skillName) {
  process.stderr.write(`[agent-skills] WARNING: "global:" prefix with empty skill name — skipping\n`);
  continue;
}
```

---

### [STYLE] Inline `require('os')` inside a loop

`require('os').homedir()` is called on every iteration of the `for` loop. `require()` is cached after the first call so this is not a performance issue, but it reads poorly. Hoist it to the top of `buildAgentSkillsBlock()` alongside the existing `const { validatePath } = require('./security.cjs')` line.

```js
const os = require('os');
```

---

### What looks good

- The regex character class is tight and correct for path component validation (no `.`, no `/`, no `\`, no null bytes).
- `startsWith('global:')` is prefix-only — it cannot match in the middle of a name.
- `warn, don't error` is the right behaviour for missing skills; it matches the existing project-relative path behaviour.
- The refactor of `validPaths` from `string[]` to `{ ref, display }[]` is clean and the `display` field is useful for future error messaging.
- The `os.homedir()` expansion is correct for Linux/macOS and Windows (Node's `os.homedir()` handles all three platforms).

---

### Summary

Two security issues (symlink escape, missing tests that would have caught it) and one correctness issue (empty-name edge case) must be fixed before merge. The style nit is minor but should be addressed in the same pass.

trek-e (changes requested):
## Review: request-changes

Linked issue #1992 carries `approved-enhancement` — the issue link is valid. The core approach (prefix-based dispatch into a separate code path) is sound. However there are three blocking issues and one style fix required before this can merge.

---

### [BLOCKING] Symlink escape is unguarded

The regex passes only names matching /^[a-zA-Z0-9_-]+$/, which is correct for preventing ../traversal in the name itself. But it cannot prevent a skill directory that is a symlink pointing outside ~/.claude/skills/. The existing validatePath() in security.cjs is built for exactly this: it calls fs.realpathSync() on both base and target, then checks containment. The new global path bypasses it entirely.

Fix: after constructing globalSkillDir, call validatePath(globalSkillMd, globalSkillsBase, { allowAbsolute: true }) and skip on failure.

  const globalSkillsBase = path.join(require('os').homedir(), '.claude', 'skills');
  const pathCheck = validatePath(globalSkillMd, globalSkillsBase, { allowAbsolute: true });
  if (!pathCheck.safe) {
    process.stderr.write('[agent-skills] WARNING: Global skill "' + skillName + '" failed path check -- skipping\n');
    continue;
  }

---

### [BLOCKING] No tests for the global: prefix path

The diff touches only init.cjs. Zero tests are added. Per CONTRIBUTING.md, an enhancement requires tests covering the enhanced behavior. The pre-existing 11 tests cover project-relative paths; none cover the new code path.

Required cases in tests/agent-skills.test.cjs:

1. global:valid-skill resolves to $HOME/.claude/skills/valid-skill/SKILL.md and appears in output
2. global:invalid!name is rejected by the regex and skipped without crashing
3. global:missing-skill where the directory does not exist -> skipped, command still succeeds
4. Mix of global: and project-relative paths in the same array -> both resolve correctly
5. global: with empty name after the prefix (skillName === '') -> rejected, clear warning

Tests must set HOME / USERPROFILE to a controlled temp dir (as existing tests already do) so the global skills base is isolated.

---

### [BLOCKING] Empty global: prefix name is mishandled with confusing message

'global:'.slice(7) produces ''. The regex /^[a-zA-Z0-9_-]+$/ requires at least one character, so '' fails the test and emits: Invalid global skill name "" -- skipping. The outcome is correct but the message is confusing. Add an explicit guard before the regex:

  if (!skillName) {
    process.stderr.write('[agent-skills] WARNING: "global:" with empty skill name -- skipping\n');
    continue;
  }

---

### [STYLE] Inline require('os') inside a loop

require('os').homedir() is called on every loop iteration. require() is cached after the first call so this is not a performance issue, but it reads poorly alongside the hoisted const { validatePath } = require('./security.cjs') at the top of the function. Hoist: const os = require('os');

---

### What looks correct

- The regex /^[a-zA-Z0-9_-]+$/ is tight: no dot, no slash, no backslash, no null bytes, no Unicode normalization vectors. Sufficient for name validation.
- startsWith('global:') is prefix-only and cannot match mid-string.
- Warn-not-error for missing skills matches existing project-relative path behavior.
- The validPaths refactor from string[] to { ref, display }[] is clean.
- os.homedir() handles macOS, Linux, and Windows correctly.

---

### Summary

Fix the symlink bypass (use validatePath for the resolved global path), add the required tests (enhancement PR without tests will be asked to add them regardless), and add the empty-name guard. The style nit should be addressed in the same pass.

trek-e (changes requested):
Review: 3 blocking issues, 1 style nit. Issue 1992 approved-enhancement label confirmed. Symlink escape unguarded (use validatePath on resolved global path). No tests added for new code path. Empty global: name produces confusing warning. Inline require(os) in loop. See comments.

trek-e (changes requested):
## Review: request-changes

Linked issue #1992 carries `approved-enhancement` — the issue link is valid. The core approach (prefix-based dispatch) is sound. Three blocking issues and one style nit must be fixed before merge.

---

### [BLOCKING] Symlink escape is unguarded

`/^[a-zA-Z0-9_-]+$/` prevents `../` traversal in the _name_ but cannot prevent a skill directory that is itself a symlink pointing outside `~/.claude/skills/`. The existing `validatePath()` in `security.cjs` is built for exactly this — it calls `fs.realpathSync()` on both base and target then checks containment. The new global path bypasses it entirely.

**Fix:** after constructing `globalSkillDir`, call `validatePath(globalSkillMd, globalSkillsBase, { allowAbsolute: true })` and skip on failure. That re-uses the existing symlink-resolution logic.

```js
const globalSkillsBase = path.join(require('os').homedir(), '.claude', 'skills');
const pathCheck = validatePath(globalSkillMd, globalSkillsBase, { allowAbsolute: true });
if (!pathCheck.safe) {
  process.stderr.write(`[agent-skills] WARNING: Global skill "${skillName}" failed path check — skipping\n`);
  continue;
}
```

---

### [BLOCKING] No tests for the `global:` prefix path

The diff touches only `init.cjs`. Zero tests are added. Per CONTRIBUTING.md, an enhancement PR requires tests covering the enhanced behavior. The pre-existing 11 tests cover project-relative paths; none cover the new code path.

Required cases in `tests/agent-skills.test.cjs`:

1. `global:valid-skill` resolves to `$HOME/.claude/skills/valid-skill/SKILL.md` and appears in output
2. `global:invalid!name` is rejected by the regex, skipped without crashing
3. `global:missing-skill` (directory absent) — skipped, command still succeeds
4. Mix of `global:` and project-relative paths in same array — both resolve correctly
5. `global:` with empty name after prefix (`skillName === ''`) — rejected with clear warning

Tests must pass `HOME`/`USERPROFILE` env vars (as existing tests already do) pointing at a controlled temp dir.

---

### [BLOCKING] Empty `global:` prefix produces confusing warning

`'global:'.slice(7)` produces `''`. The regex requires at least one character, so `''` fails and emits: `Invalid global skill name "" — skipping`. The outcome is correct but the message is confusing. Add an explicit guard before the regex:

```js
if (!skillName) {
  process.stderr.write(`[agent-skills] WARNING: "global:" prefix with empty skill name — skipping\n`);
  continue;
}
```

---

### [STYLE] Inline `require('os')` inside a loop

`require('os').homedir()` is called on every loop iteration. `require()` is cached so no performance issue, but it reads poorly. Hoist alongside the existing `const { validatePath } = require('./security.cjs')` at the top of the function.

---

### What looks correct

- Regex `/^[a-zA-Z0-9_-]+$/` is tight: no dot, slash, backslash, null bytes, or Unicode normalization vectors. Sufficient for name-only validation.
- `startsWith('global:')` is prefix-only; cannot match mid-string.
- Warn-not-error for missing skills matches existing project-relative path behavior.
- `validPaths` refactor from `string[]` to `{ ref, display }[]` is clean.
- `os.homedir()` handles macOS, Linux, and Windows correctly.

---

### Summary

Fix the symlink bypass by routing the resolved global path through `validatePath`. Add the required tests (no enhancement merges without them per CONTRIBUTING.md). Add the empty-name guard. Style nit in same pass.

Tibsfox:
Thanks @trek-e! All three blocking issues + the style nit addressed in 0a18fc3:

**1. Symlink escape guard** — Added `validatePath(globalSkillMd, globalSkillsBase, { allowAbsolute: true })` after the existence check. Re-uses the existing symlink-resolution and containment logic in `security.cjs`. A skill directory symlinked to an arbitrary location now fails the path check with a clear warning.

**2. Empty name guard** — Added explicit `if (!skillName)` guard before the regex so `global:` produces `"global:" prefix with empty skill name` instead of the confusing `Invalid global skill name ""`.

**3. Tests** — Added 5 new test cases in `tests/agent-skills.test.cjs`:
1. `global:valid-skill` resolves to `$HOME/.claude/skills/valid-skill/SKILL.md` and appears in output
2. `global:invalid!name` rejected by regex, skipped without crashing
3. `global:missing-skill` (directory absent) skipped gracefully
4. Mix of `global:` and project-relative paths both resolve
5. `global:` with empty name produces clear warning and skips

Tests use `fs.mkdtempSync` + `HOME=fakeHome` pattern to set up isolated `~/.claude/skills/` directories.

**4. Style fix** — Hoisted `require('os')` and `globalSkillsBase` calculation out of the loop, alongside the existing `validatePath` import at the top of `buildAgentSkillsBlock`.

All 16 agent-skills tests pass (11 existing + 5 new).

deadpoolawsome (approved):

trek-e (approved):

PR #2093, opened 2026-04-11
fix(workflow): carve out Other+empty exception from answer_validation retry loop

### LINKED ISSUE #2085: discuss-phase: "Other" (Chat more) selection triggers immediate question generation instead of waiting for user input
## Problem

When a user selects "Other" (displayed as "Chat more" in the UI) on an `AskUserQuestion` prompt during `/gsd-discuss-phase`, Claude immediately generates another set of questions instead of waiting for the user to type their input. This can cascade — if the user doesn't type fast enough, 2-3 rounds of unwanted questions fire off before they can respond.

## Steps to Reproduce

1. Run `/gsd-discuss-phase` for any phase
2. On any gray area question, select "Other" / "Chat more"
3. Observe: Claude immediately fires another `AskUserQuestion` or generates new questions
4. User's intended free-text input is never captured

## Expected Behavior

When the user selects "Other" / "Chat more":
1. Claude outputs a short plain-text prompt (e.g. "What would you like to discuss?")
2. Claude **stops and waits indefinitely** for the user's next message
3. Only after the user types their input and presses Enter does Claude respond
4. Claude reflects back the user's input, then resumes the normal question flow

## Actual Behavior

`AskUserQuestion` returns immediately with just the "Other" selection (no text body). Claude interprets this as a completed answer and proceeds to generate the next question or batch. The user never gets a chance to type their thoughts.

## Root Cause

The `AskUserQuestion` tool returns as soon as a selection is made. When "Other" is selected, there's no mechanism to block until the user provides follow-up text. The workflow instructions in `discuss-phase.md` say to "wait for them to type at the normal prompt" but there's no enforcement — Claude's natural behavior is to continue generating output after receiving a tool result.

## Suggested Fix

In `discuss-phase.md`, strengthen the "Other" handling in the `discuss_areas` step:

```markdown
**CRITICAL — "Other" / "Chat more" handling:**
When AskUserQuestion returns with "Other" selected and NO accompanying user text 
(empty string or whitespace-only), you MUST:
1. Output a single plain-text line: "What would you like to discuss?"
2. STOP generating. Do not call any tools. Do not output any further text.
3. Wait for the user's next message.
4. Only after receiving the user's message, reflect it back and continue.

Do NOT interpret an empty "Other" response as a signal to generate more questions.
```

## Impact

- Frustrating UX — user's intent to discuss is overridden by auto-generated questions
- Can waste context window with 2-3 rounds of unwanted questions
- Undermines the "thinking partner" philosophy of discuss-phase

trek-e:
## What

When a user selects "Other" (displayed as "Chat more") in a discuss-phase AskUserQuestion prompt with no accompanying text, the response body is empty. The existing answer_validation block at line 115 classified this as a generic empty response and retried the AskUserQuestion — causing 2-3 cascading rounds of unwanted questions before the user could type anything.

The intended behavior on line 795 (ask follow-up as plain text, wait for the user to type) was never reached because the retry path ran first.

## Why

"Other" + empty text is a distinct signal: the user wants to type freeform input. It is not an unanswered question. The answer_validation block needs an explicit exception for this case so it hands control back to the user instead of retrying.

## How

Added an "Other + empty text" exception at the top of the answer_validation block. When "Other" is selected with an empty body, the workflow outputs one plain-text prompt line and stops — it does not retry the AskUserQuestion or generate more questions. All other empty responses continue through the existing retry path unchanged.

The fix applies once, in the answer_validation block, so all three locations where "Other" appears in discuss-phase (gray-area picker, 4-question batch flow, loop-control menu) inherit the exception without redundant wording.

All 3015 tests pass.

Fixes #2085

### DIFF
--- a/get-shit-done/workflows/discuss-phase.md
+++ b/get-shit-done/workflows/discuss-phase.md
@@ -113,6 +113,15 @@ Phase: "API documentation"
 
 <answer_validation>
 **IMPORTANT: Answer validation** — After every AskUserQuestion call, check if the response is empty or whitespace-only. If so:
+
+**Exception — "Other" with empty text:** If the user selected "Other" (or "Chat more") and the response body is empty or whitespace-only, this is NOT an empty answer — it is a signal that the user wants to type freeform input. In this case:
+1. Output a single plain-text line: "What would you like to discuss?"
+2. STOP generating. Do not call any tools. Do not output any further text.
+3. Wait for the user's next message.
+4. After receiving their message, reflect it back and continue.
+Do NOT retry the AskUserQuestion or generate more questions when "Other" is selected with empty text.
+
+**All other empty responses:** If the response is empty or whitespace-only (and the user did NOT select "Other"):
 1. Retry the question once with the same parameters
 2. If still empty, present the options as a plain-text numbered list and ask the user to type their choice number
 Never proceed with an empty answer.


trek-e (commented):
Adversarial review — passed.

**Security / injection:** The new exception block is maintainer-controlled workflow instruction, not untrusted user content. No injection surface. The instruction to "output a single plain-text line" and "STOP generating" is a safe behavioral directive — it constrains the agent rather than expanding its permissions.

**Behavioral accuracy:** The exception carves out exactly the "Other + empty body" case before the existing retry logic. All other empty responses still hit the retry path. The ordering (exception first, fallthrough to retry) is correct. No silent behavioral change to non-Other flows.

**Edge case:** The condition relies on the agent correctly identifying "Other"/"Chat more" selection. If the AskUserQuestion response format doesn't clearly distinguish option selection from free-text entry, the agent may misclassify. This is a runtime concern, not a code defect — worth noting for follow-up.

**Closing keyword:** `Fixes #2085` present. No conflicts with other open PRs.

**Verdict:** Safe to merge.

trek-e (commented):
Adversarial review — passed.

Security / injection: The new exception block is maintainer-controlled workflow instruction, not untrusted user content. No injection surface. The instruction to output a single plain-text line and STOP generating is a safe behavioral directive — it constrains the agent rather than expanding its permissions.

Behavioral accuracy: The exception carves out exactly the Other + empty body case before the existing retry logic. All other empty responses still hit the retry path. The ordering (exception first, fallthrough to retry) is correct. No silent behavioral change to non-Other flows.

Edge case worth noting: The condition relies on the agent correctly identifying Other/Chat more selection vs free-text entry. If the AskUserQuestion response format doesn't clearly distinguish between them, the agent may misclassify — worth a follow-up if reports surface.

Closing keyword: Fixes #2085 present. No conflicts with other open PRs.

Verdict: Safe to merge.

PR #2198, opened 2026-04-14
fix: parse baseInt to integer in renameDecimalPhases to fix regex mismatch

### LINKED ISSUE #2197: fix: renameDecimalPhases receives string baseInt causing regex mismatch for zero-padded phases
## Problem

In `get-shit-done/bin/lib/phase.cjs`, the function `renameDecimalPhases` is called from `cmdPhaseRemove` with `baseInt` passed as the result of `.split('.')[0]` on a normalized phase string — which is a **string**, not a number.

### Faulty Call Site (~line 715)

```js
const renamed = isDecimal
  ? renameDecimalPhases(
      phasesDir,
      normalized.split('.')[0],           // returns string e.g. "06"
      parseInt(normalized.split('.')[1], 10)
    )
```

### Faulty Regex (~line 588)

```js
function renameDecimalPhases(phasesDir, baseInt, removedDecimal) {
  const decPattern = new RegExp(`^${baseInt}\\.(\\d+)-(.+)$`);
  // If baseInt is "06", pattern is: /^06\.(\d+)-(.+)$/
  // But dirs are named like "6.1-feature" (no leading zero in base).
  // Pattern NEVER matches — all renaming is silently skipped.
```

If the normalized phase has a leading zero (e.g. `06.2`), `split('.')[0]` returns `"06"`, but directories on disk are named `6.1-slug` (no leading zero in the integer). The regex never matches, so **no sibling decimal phases get renumbered** after a removal. Phase directories fall out of sync with ROADMAP.

## Steps to Reproduce

1. Create a project with phases `6.1`, `6.2`, `6.3`.
2. Run `gsd phase remove 06.1` (or `6.1`).
3. Observe: phases `6.2` and `6.3` on disk are **not** renamed to `6.1` and `6.2`.

## Expected Behavior

After removing a decimal phase, sibling decimal phases should be renumbered to fill the gap (e.g., `6.2` becomes `6.1`, `6.3` becomes `6.2`).

## Fix

Parse `baseInt` to a plain integer at the call site:

```js
const renamed = isDecimal
  ? renameDecimalPhases(
      phasesDir,
      parseInt(normalized.split('.')[0], 10),   // fix: parse to int
      parseInt(normalized.split('.')[1], 10)
    )
```

Or inside `renameDecimalPhases`:

```js
function renameDecimalPhases(phasesDir, baseInt, removedDecimal) {
  const base = parseInt(baseInt, 10);  // normalize to integer
  const decPattern = new RegExp(`^${base}\\.(\\d+)-(.+)$`);
  ...
}
```

kuishou68:
Closes #2197

## Problem
In `cmdPhaseRemove`, `renameDecimalPhases` was called with `normalized.split('.')[0]` which returns a **string** (e.g. `"06"`). Inside `renameDecimalPhases`, this string was used directly in a regex: `/^06\.(\d+)-(.+)$/`. Phase directories are named without leading zeros (e.g. `6.1-feature`), so the pattern never matched and no decimal siblings were renumbered after a phase removal.

## Fix
Parse the base part to an integer with `parseInt(..., 10)` at the call site:

```js
// Before
renameDecimalPhases(phasesDir, normalized.split('.')[0], ...)
// After  
renameDecimalPhases(phasesDir, parseInt(normalized.split('.')[0], 10), ...)
```

This is a one-character change that ensures the regex matches directory names correctly.

### DIFF
--- a/get-shit-done/bin/lib/phase.cjs
+++ b/get-shit-done/bin/lib/phase.cjs
@@ -712,7 +712,7 @@ function cmdPhaseRemove(cwd, targetPhase, options, raw) {
   let renamedDirs = [], renamedFiles = [];
   try {
     const renamed = isDecimal
-      ? renameDecimalPhases(phasesDir, normalized.split('.')[0], parseInt(normalized.split('.')[1], 10))
+      ? renameDecimalPhases(phasesDir, parseInt(normalized.split('.')[0], 10), parseInt(normalized.split('.')[1], 10))
       : renameIntegerPhases(phasesDir, parseInt(normalized, 10));
     renamedDirs = renamed.renamedDirs;
     renamedFiles = renamed.renamedFiles;


trek-e (commented):
Adversarial review — PR #2198

**Finding: Fix is correct and complete.**

Root cause confirmed: `normalized.split('.')[0]` returns a string (e.g. `"06"`), which gets interpolated into the regex at line 588 as `^06\.`. Phase directories are named without leading zeros (e.g. `6.1-feature`), so the regex never matches — decimal siblings are silently skipped during phase removal.

With `parseInt(normalized.split('.')[0], 10)` the regex becomes `^6\.`, which correctly matches `6.1-feature`, `6.2-auth`, etc.

**Verification checks:**
- Single call site at line 715 — no other paths that pass a string to `baseInt` parameter
- `parseInt("06", 10)` = 6, no NaN risk (normalized is always a validated phase string like "06.2")
- `removedDecimal` was already `parseInt`-wrapped at the call site — the PR brings `baseInt` in line with it
- No security surface: `baseInt` is only used in `new RegExp(...)` for the pattern match, never in shell or FS operations unvalidated

No issues found. Fix is targeted, correct, and adds no risk.

@trek-e

PR #2207, opened 2026-04-14
fix(hooks): add gsd-hook-version to bash hooks, extend check-update regex for bash syntax

### LINKED ISSUE #2206: fix(hooks): bash hooks missing gsd-hook-version header, check-update always flags them as stale
## Bug Description

Three bash hook files (`gsd-phase-boundary.sh`, `gsd-session-state.sh`, `gsd-validate-commit.sh`) are missing the `# gsd-hook-version` header. As a result, `gsd-check-update.js` always flags them as stale with `hookVersion: 'unknown'`, triggering the stale hooks warning in the statusline on every session — even immediately after a fresh install.

## Root Cause

`gsd-check-update.js` detects stale hooks by scanning for `// gsd-hook-version: <version>` in hook files. This regex uses JavaScript comment syntax (`//`) and does not match bash comment syntax (`#`). Since the bash hooks have no version header at all, they fall into the `else` branch unconditionally:

```js
// line 106 in gsd-check-update.js
const versionMatch = content.match(/\/\/ gsd-hook-version:\s*(.+)/);
if (versionMatch) {
  // ... version comparison
} else {
  // No version header at all — definitely stale (pre-version-tracking)
  staleHooks.push({ file: hookFile, hookVersion: 'unknown', installedVersion: installed });
}
```

## Proposed Fix

Two-part change:

**1. Add version header to bash hooks** (after shebang line):
```bash
#!/bin/bash
# gsd-hook-version: {{GSD_VERSION}}
```
Files: `hooks/gsd-phase-boundary.sh`, `hooks/gsd-session-state.sh`, `hooks/gsd-validate-commit.sh`

**2. Extend regex in `gsd-check-update.js`** to handle both comment syntaxes:
```js
// before
const versionMatch = content.match(/\/\/ gsd-hook-version:\s*(.+)/);
// after
```

## Steps to Reproduce

1. Fresh install of GSD v1.35.0
2. Wait for `gsd-check-update.js` to run (or trigger manually)
3. Observe `stale_hooks` in cache always contains all three bash hooks with `hookVersion: 'unknown'`
4. Statusline shows `⚠ stale hooks — run /gsd-update` (or `⚠ dev install — re-run installer`) despite hooks being current

## Impact

False positive stale hooks warning on every session after install. Users who install from source (version ahead of npm latest) see `⚠ dev install — re-run installer to sync hooks` permanently.

j2h4u:
Fixes #2206

## Summary

- Add `# gsd-hook-version: {{GSD_VERSION}}` header to `gsd-phase-boundary.sh`, `gsd-session-state.sh`, and `gsd-validate-commit.sh`
- Extend the version regex in `gsd-check-update.js` from `// gsd-hook-version:` to `(?:\/\/|#) gsd-hook-version:` to handle both JS and bash comment syntax

## Test plan

- [ ] Fresh install: verify `~/.cache/gsd/gsd-update-check.json` has no `stale_hooks` entries for the three bash files
- [ ] Statusline shows no stale hooks warning after install
- [ ] JS hooks still detected correctly (regex change is additive)

### DIFF
--- a/hooks/gsd-check-update.js
+++ b/hooks/gsd-check-update.js
@@ -103,7 +103,7 @@ const child = spawn(process.execPath, ['-e', `
         for (const hookFile of hookFiles) {
           try {
             const content = fs.readFileSync(path.join(hooksDir, hookFile), 'utf8');
-            const versionMatch = content.match(/\\/\\/ gsd-hook-version:\\s*(.+)/);
+            const versionMatch = content.match(/(?:\/\/|#) gsd-hook-version:\s*(.+)/);
             if (versionMatch) {
               const hookVersion = versionMatch[1].trim();
               if (isNewer(installed, hookVersion) && !hookVersion.includes('{{')) {
--- a/hooks/gsd-phase-boundary.sh
+++ b/hooks/gsd-phase-boundary.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# gsd-hook-version: {{GSD_VERSION}}
 # gsd-phase-boundary.sh — PostToolUse hook: detect .planning/ file writes
 # Outputs a reminder when planning files are modified outside normal workflow.
 # Uses Node.js for JSON parsing (always available in GSD projects, no jq dependency).
--- a/hooks/gsd-session-state.sh
+++ b/hooks/gsd-session-state.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# gsd-hook-version: {{GSD_VERSION}}
 # gsd-session-state.sh — SessionStart hook: inject project state reminder
 # Outputs STATE.md head on every session start for orientation.
 #
--- a/hooks/gsd-validate-commit.sh
+++ b/hooks/gsd-validate-commit.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# gsd-hook-version: {{GSD_VERSION}}
 # gsd-validate-commit.sh — PreToolUse hook: enforce Conventional Commits format
 # Blocks git commit commands with non-conforming messages (exit 2).
 # Allows conforming messages and all non-commit commands (exit 0).


trek-e:
Thanks for catching this and putting together a fix — the regex extension to handle bash \`#\` comment syntax is exactly the right direction and your approach is credited in the PR that landed this fix (#2224).

I wanted to leave some notes here in case they're useful for future contributions, because this bug turned out to have a structure that's easy to miss.

---

## Why this PR alone doesn't fully fix the issue

Your fix addresses two of the three required parts correctly: adding the \`# gsd-hook-version: {{GSD_VERSION}}\` header to the bash hooks, and extending the detector regex. The missing piece is in \`install.js\`.

When \`install.js\` copies \`.sh\` hooks, it uses \`copyFileSync\` — a raw byte copy with no substitution. That means the installed bash hooks contain the **literal string** \`{{GSD_VERSION}}\` rather than a real semver. The detector has a guard for this:

```js
if (isNewer(installed, hookVersion) && !hookVersion.includes('{{')) {
  staleHooks.push(...)
}
```

The \`!hookVersion.includes('{{')\` check silently skips hooks whose version is still a template placeholder. So with your fix applied:

1. Fresh install: \`.sh\` hooks contain \`# gsd-hook-version: {{GSD_VERSION}}\` (literal)
2. Detector reads the header, sees \`{{\`, silently skips the stale check — **false positive clears** ✓
3. After a future GSD update, your version is now higher than the hook's version... but the hook's version is still \`{{GSD_VERSION}}\` (placeholder)
4. Detector skips it again — **stale bash hooks are permanently undetectable** ✗

The fix appears to work on a fresh install but silently breaks the feature it's meant to protect.

---

## A useful debugging technique: trace the "after" state

A reliable way to catch this class of bug is to ask not just *"does the symptom go away?"* but *"what does the system look like after a future update with this fix in place?"*

Concretely: after applying a fix, run the installer into a temp directory and inspect the installed artifacts directly:

```bash
CLAUDE_CONFIG_DIR=/tmp/test-install node bin/install.js --claude --global --yes
grep "gsd-hook-version" /tmp/test-install/hooks/gsd-phase-boundary.sh
```

If that shows \`{{GSD_VERSION}}\` (the placeholder), the installer never ran substitution. If it shows \`1.36.0\` (a real version), it did. This kind of "install and inspect" check would have surfaced the missing substitution immediately.

---

## What the final fix does differently

PR #2224 adds all three parts:
1. Headers in the bash hook sources (your approach)
2. \`install.js\` reads+substitutes \`{{GSD_VERSION}}\` in \`.sh\` files at install time (was missing)
3. Detector regex handles both \`//\` and \`#\` syntax (your approach)

It also extracts the entire stale-detection subprocess from an inline \`node -e '<template literal>'\` into a dedicated \`gsd-check-update-worker.js\` — which eliminates the double-escaping problem that made the regex hard to reason about in the first place.

Thanks again for the contribution. The regex approach you identified was the right call.

Issue #2212, opened 2026-04-14
Shell hooks missing gsd-hook-version header causes stale hooks warning after fresh install


AnonymousScripting:
## Bug

After updating to v1.36.0 via `/gsd-update`, the statusline immediately shows `⚠ stale hooks — run /gsd-update` despite the update just completing successfully.

## Root Cause

The installer adds `// gsd-hook-version: X.Y.Z` headers to `.js` hook files but does **not** add `# gsd-hook-version: X.Y.Z` headers to `.sh` hook files.

The staleness check in `gsd-check-update.js` (lines ~82-120) scans all `MANAGED_HOOKS` for a version header via regex `/\/\/ gsd-hook-version:\s*(.+)/`. When a hook has no version header, it's treated as "pre-version-tracking" and flagged as stale — even if it was just installed seconds ago.

## Affected Files

- `gsd-phase-boundary.sh` — no version header
- `gsd-session-state.sh` — no version header
- `gsd-validate-commit.sh` — no version header

All JS hooks (`gsd-check-update.js`, `gsd-statusline.js`, etc.) have the correct header and pass the check.

## Cache Evidence

`~/.cache/gsd/gsd-update-check.json` after fresh v1.36.0 install:
```json
{
  "stale_hooks": [
    { "file": "gsd-phase-boundary.sh", "hookVersion": "unknown", "installedVersion": "1.36.0" },
    { "file": "gsd-session-state.sh", "hookVersion": "unknown", "installedVersion": "1.36.0" },
    { "file": "gsd-validate-commit.sh", "hookVersion": "unknown", "installedVersion": "1.36.0" }
  ]
}
```

## Fix

Either:

1. **Installer fix**: Add `# gsd-hook-version: X.Y.Z` as line 2 of each `.sh` hook during installation (matching what's done for `.js` hooks)
2. **Detection fix**: Update the regex in `gsd-check-update.js` to also match bash-style comments: `/[#\/]{1,2}\s*gsd-hook-version:\s*(.+)/`

## Workaround

Manually add the version header to each shell script after the shebang:
```bash
#!/bin/bash
# gsd-hook-version: 1.36.0
```
Then clear the cache:
```bash
rm -f ~/.cache/gsd/gsd-update-check.json ~/.claude/cache/gsd-update-check.json
```

## Environment

- GSD v1.36.0
- Runtime: Claude Code
- Install scope: Global (`~/.claude/`)
- Platform: macOS (Darwin 25.3.0)

[...]

Issue #2210, opened 2026-04-14
Shell hooks (.sh) always flagged as stale — version regex only matches JS comments


kwazema:
## Bug

The 3 managed `.sh` hooks are **always** reported as stale, even immediately after running `/gsd-update`. The `⚠ stale hooks — run /gsd-update` warning persists across sessions.

## Root cause

`gsd-check-update.js` (line ~106) uses a single regex to detect hook version headers:

```js
const versionMatch = content.match(/\/\/ gsd-hook-version:\s*(.+)/);
```

This only matches JavaScript-style comments (`//`). The 3 shell hooks use `#` comments, so they never contain a `// gsd-hook-version:` line — the checker always classifies them as `hookVersion: 'unknown'` → stale.

**Affected hooks** (all in `MANAGED_HOOKS` list):
- `gsd-phase-boundary.sh`
- `gsd-session-state.sh`
- `gsd-validate-commit.sh`

## Workaround

Adding `# // gsd-hook-version: 1.36.0` to each `.sh` file satisfies both bash (treats the line as a comment) and the JS regex (finds the pattern mid-line).

## Suggested fix

Either:
1. Update the regex to also match shell-style headers: `/(?:\/\/|#)\s*gsd-hook-version:\s*(.+)/`
2. Or have the GSD installer inject the `# // gsd-hook-version:` header when writing `.sh` hooks during `/gsd-update`

## Environment

- GSD version: 1.36.0
- OS: Windows 11 (bash via Git Bash)
- Claude Code CLI

trek-e:
Duplicate of #2136. The fix is in PR #2215 (stamp `gsd-hook-version` into shell hooks + installer substitution). Closing as duplicate — once PR #2215 merges the stale-hooks warning will be resolved.

Issue #2209, opened 2026-04-14
Stale hooks false positive: shipped .sh hooks missing version header flagged by gsd-check-update


YARSON-AI:
## Summary

After updating to 1.36.0, the statusline shows `⚠ stale hooks — run /gsd-update` on every session start, but `/gsd-update` reports the latest version is already installed. The warning is a false positive caused by an internal inconsistency in 1.36.0.

## Environment

- `get-shit-done-cc@1.36.0` (global install at `~/.claude/get-shit-done/`)
- Claude Code 2.1.107, macOS (Darwin 24.6.0)
- Hooks installed at `~/.claude/hooks/`

## Root cause

`hooks/gsd-check-update.js` (#2141) added the three bash hooks to `MANAGED_HOOKS`:

```js
const MANAGED_HOOKS = [
  // ...
  'gsd-phase-boundary.sh',
  'gsd-session-state.sh',
  'gsd-validate-commit.sh',
  // ...
];
```

For each managed hook it runs:

```js
const versionMatch = content.match(/\/\/ gsd-hook-version:\s*(.+)/);
if (versionMatch) { /* ... */ }
else {
  // No version header at all — definitely stale (pre-version-tracking)
  staleHooks.push({ file: hookFile, hookVersion: 'unknown', installedVersion: installed });
}
```

Two problems combine to guarantee a false positive on every install:

1. **The `.sh` files shipped in `hooks/dist/` on npm do not contain any `gsd-hook-version:` header.** Verified against the 1.36.0 tarball (`https://registry.npmjs.org/get-shit-done-cc/-/get-shit-done-cc-1.36.0.tgz`) — first 6 lines of all three files are plain descriptive comments, no version marker.
2. **The regex `// gsd-hook-version:` only matches `//`-style (JS) comments.** Even if a maintainer adds a shell-style `# gsd-hook-version: 1.36.0` line, it would still be flagged as stale because `#` doesn't match `//`.

Result: every fresh `/gsd-update` install immediately has 3 stale hooks recorded in `~/.cache/gsd/gsd-update-check.json`, and `gsd-statusline.js:226` renders the warning on every session.

## Repro

1. Install/update to 1.36.0.
2. Start a new Claude Code session.
3. Observe `⚠ stale hooks — run /gsd-update` in the statusline.
4. Run `/gsd-update` → reports "already latest".
5. Inspect cache:

```
$ cat ~/.cache/gsd/gsd-update-check.json
[...]

Issue #1834, opened 2026-04-05
bug: SessionStart hook references missing gsd-session-state.sh (and 2 other .sh hooks)


sriram-rach:
### Describe the bug

The installer writes three `.sh` hook entries into `.claude/settings.json`, but the corresponding shell scripts are never installed to `.claude/hooks/`. This causes a hook error on every `SessionStart` event and likely on `PostToolUse` and `PreToolUse` events involving the other two missing scripts.

Missing files referenced in `settings.json`:
1. `bash .claude/hooks/gsd-session-state.sh` — `SessionStart` hook
2. `bash .claude/hooks/gsd-phase-boundary.sh` — `PostToolUse` hook (matcher: `Write|Edit`)
3. `bash .claude/hooks/gsd-validate-commit.sh` — `PreToolUse` hook (matcher: `Bash`)

Only `.js` hooks are actually installed:
```
.claude/hooks/gsd-check-update.js
.claude/hooks/gsd-context-monitor.js
.claude/hooks/gsd-prompt-guard.js
.claude/hooks/gsd-read-guard.js
.claude/hooks/gsd-statusline.js
.claude/hooks/gsd-workflow-guard.js
```

### To reproduce

1. Install GSD v1.32.0 (`npm i -g get-shit-done-cc`)
2. Open a new Claude Code session in a project with GSD initialized
3. Observe `SessionStart` hook error for `gsd-session-state.sh`

Alternatively:
```bash
ls .claude/hooks/gsd-session-state.sh
# No such file or directory

ls .claude/hooks/gsd-phase-boundary.sh
# No such file or directory

ls .claude/hooks/gsd-validate-commit.sh
# No such file or directory
```

### Expected behavior

Either:
- The `.sh` hook files should be installed alongside the `.js` hooks, **or**
- The `settings.json` entries referencing them should not be written if the files don't exist

### Actual behavior

`settings.json` references three `.sh` hook files that are never created, causing errors on every session start (and potentially on every `Write|Edit` and `Bash` tool use).

### Environment

- GSD version: 1.32.0
- Node.js: v24
- OS: macOS (Darwin 25.4.0)
- Claude Code: 2.1.92

PR #2215, opened 2026-04-14
fix(hooks): stamp gsd-hook-version into shell hooks

### LINKED ISSUE #2136: MANAGED_HOOKS list missing bash hooks for staleness check
## Summary

`gsd-check-update.js` has a `MANAGED_HOOKS` whitelist used to detect stale hooks after a GSD update. It only includes `.js` hooks, missing the 3 bash hooks that GSD also ships:

**In MANAGED_HOOKS (6):**
- gsd-check-update.js
- gsd-context-monitor.js
- gsd-prompt-guard.js
- gsd-read-guard.js
- gsd-statusline.js
- gsd-workflow-guard.js

**Missing (3):**
- gsd-phase-boundary.sh
- gsd-session-state.sh
- gsd-validate-commit.sh

## Impact

After a GSD update, the staleness checker won't flag outdated bash hooks. If a breaking change ships in one of these 3 scripts, users won't get the "stale hooks" warning.

## Suggested fix

Either add the bash hooks to `MANAGED_HOOKS` (they'd need `# gsd-hook-version:` headers), or auto-discover hooks by scanning the hooks directory instead of maintaining a static list.

## Environment

- GSD 1.34.2
- Claude Code 2.1.104

### LINKED ISSUE #2206: fix(hooks): bash hooks missing gsd-hook-version header, check-update always flags them as stale

nitsan2dots:
## Linked Issue

Fixes #2136

Closes #2206, #2209, #2210, #2212 (duplicates reported independently).

> #2136 carries `bug`, `confirmed`, `fix-pending`, `area: hooks` labels. Its suggested remediation — _"add the bash hooks to MANAGED_HOOKS (they'd need `# gsd-hook-version:` headers)"_ — is exactly what this PR implements.

---

## What was broken

After a fresh install of GSD 1.36.0 (and earlier), every session immediately showed the warning:

```
⚠ stale hooks — run /gsd-update
```

even though the installed version already matched the latest release on npm. Running `/gsd-update` would bail with _"already on the latest version"_ and the warning never cleared.

## What this fix does

1. Adds a `# gsd-hook-version: {{GSD_VERSION}}` header line to the three shipped shell hooks (`gsd-phase-boundary.sh`, `gsd-session-state.sh`, `gsd-validate-commit.sh`), mirroring the convention already used by the six `.js` hooks.
2. Extends `bin/install.js` — both the bundled-hooks branch and the Codex branch — to substitute `{{GSD_VERSION}}` in `.sh` hooks at install time, the same way it already did for `.js` hooks.

## Root cause

`gsd-check-update.js` scans all managed hooks for a `gsd-hook-version` marker and compares against the installed VERSION. The `.js` hooks carried the marker and were substituted by the installer at `bin/install.js:5759`. The `.sh` hooks were simply `copyFileSync`'d — no header, no substitution — so the checker recorded them with `hookVersion: "unknown"` and persistently marked them stale.

### How I verified the fix

1. Ran `npm test` — all 3,868 tests pass (including the existing `#1834`/`#1755` regression suite which pins the outer `} else {` structure in the hook copy loop).
2. Built hooks and ran the installer into a temp directory; verified all three `.sh` hooks now contain `# gsd-hook-version: 1.36.0` (concrete version, not the template placeholder).
3. Cleared `~/.cache/gsd/gsd-update-check.json`; on next session the stale warning no longer appears.

### Regression test added

Yes — `tests/install-hooks-copy.test.cjs` now has two new cases:
- Asserts all three `.sh` hook sources carry `# gsd-hook-version: {{GSD_VERSION}}`.
- Asserts `install.js` performs `{{GSD_VERSION}}` substitution in the `.sh` path.

### Platforms tested

- [x] macOS
- [ ] Windows — logic is platform-agnostic; `chmod` already guarded with try/catch for Windows as before
- [ ] Linux — logic identical to macOS
- [x] N/A (file copy + string substitution — no platform-specific behavior)

### Runtimes tested

- [x] Claude Code — end-to-end install + verified stamped versions
- [ ] Gemini CLI, OpenCode, Kilo, Codex — same code path (the Codex-specific hook block was updated symmetrically); not manually verified
- [x] N/A for the bug itself — the warning fires on any runtime that installs the shell hooks

---

## Checklist

- [x] Linked issue has the `confirmed` + `fix-pending` labels (#2136)
- [x] Regression test added
- [x] All existing tests pass (`npm test`) — 3,868/3,868
- [x] CHANGELOG.md updated under `[Unreleased] → Fixed`

## Breaking changes

None. The substitution is additive: shell hooks previously had no version header, now they have one. Downstream code (`gsd-check-update.js`) already expects and handles this header.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- **Shell hooks falsely flagged as stale** — `gsd-phase-boundary.sh`, `gsd-session-state.sh`, and `gsd-validate-commit.sh` now ship with a `gsd-hook-version` header, and the installer substitutes `{{GSD_VERSION}}` in `.sh` hooks the same way it does for `.js` hooks. Resolves the "⚠ stale hooks — run /gsd-update" warning on fresh installs (#2136, #2206, #2209, #2210, #2212)
+
 ## [1.36.0] - 2026-04-14
 
 ### Added
--- a/bin/install.js
+++ b/bin/install.js
@@ -5761,10 +5761,16 @@ function install(isGlobal, runtime = 'claude') {
             // Ensure hook files are executable (fixes #1162 — missing +x permission)
             try { fs.chmodSync(destFile, 0o755); } catch (e) { /* Windows doesn't support chmod */ }
           } else {
-            fs.copyFileSync(srcFile, destFile);
-            // Ensure .sh hook files are executable (mirrors chmod in build-hooks.js)
+            // Non-.js hooks: .sh hooks also carry a gsd-hook-version header so
+            // gsd-check-update.js can detect staleness after updates — stamp the
+            // version in the same way we do for .js hooks.
             if (entry.endsWith('.sh')) {
+              let content = fs.readFileSync(srcFile, 'utf8');
+              content = content.replace(/\{\{GSD_VERSION\}\}/g, pkg.version);
+              fs.writeFileSync(destFile, content);
               try { fs.chmodSync(destFile, 0o755); } catch (e) { /* Windows doesn't support chmod */ }
+            } else {
+              fs.copyFileSync(srcFile, destFile);
             }
           }
         }
@@ -5876,9 +5882,13 @@ function install(isGlobal, runtime = 'claude') {
           fs.writeFileSync(destFile, content);
           try { fs.chmodSync(destFile, 0o755); } catch (e) { /* Windows */ }
         } else {
-          fs.copyFileSync(srcFile, destFile);
           if (entry.endsWith('.sh')) {
+            let content = fs.readFileSync(srcFile, 'utf8');
+            content = content.replace(/\{\{GSD_VERSION\}\}/g, pkg.version);
+            fs.writeFileSync(destFile, content);
             try { fs.chmodSync(destFile, 0o755); } catch (e) { /* Windows */ }
+          } else {
+            fs.copyFileSync(srcFile, destFile);
           }
         }
       }
--- a/hooks/gsd-phase-boundary.sh
+++ b/hooks/gsd-phase-boundary.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# gsd-hook-version: {{GSD_VERSION}}
 # gsd-phase-boundary.sh — PostToolUse hook: detect .planning/ file writes
 # Outputs a reminder when planning files are modified outside normal workflow.
 # Uses Node.js for JSON parsing (always available in GSD projects, no jq dependency).
--- a/hooks/gsd-session-state.sh
+++ b/hooks/gsd-session-state.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# gsd-hook-version: {{GSD_VERSION}}
 # gsd-session-state.sh — SessionStart hook: inject project state reminder
 # Outputs STATE.md head on every session start for orientation.
 #
--- a/hooks/gsd-validate-commit.sh
+++ b/hooks/gsd-validate-commit.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# gsd-hook-version: {{GSD_VERSION}}
 # gsd-validate-commit.sh — PreToolUse hook: enforce Conventional Commits format
 # Blocks git commit commands with non-conforming messages (exit 2).
 # Allows conforming messages and all non-commit commands (exit 0).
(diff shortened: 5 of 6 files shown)


trek-e:
This is a thorough, well-documented PR — the install.js substitution for \`.sh\` hooks is exactly the right fix for the stamping side of this bug, and your approach is credited in the PR that landed the complete fix (#2224).

A few notes in case they're useful, because the bug had a third part that's easy to miss.

---

## Why this PR alone doesn't fully fix the issue

You correctly identified that \`install.js\` was \`copyFileSync\`-ing the bash hooks without substituting \`{{GSD_VERSION}}\`, and you fixed that. The missing piece is in the detector itself.

The version check regex in \`gsd-check-update.js\` (at the time of this PR) was:

```js
const versionMatch = content.match(/\\/\\/ gsd-hook-version:\\s*(.+)/);
```

That double-escaped form evaluates in the subprocess as:
```js
content.match(/\/\/ gsd-hook-version:\s*(.+)/)
```

Which only matches JavaScript \`//\` comment syntax — never bash \`#\`. So even with your fix applied:

1. Fresh install: \`.sh\` hooks are correctly stamped with e.g. \`# gsd-hook-version: 1.36.0\` ✓
2. Detector reads the hook, tries to match \`// gsd-hook-version:\` against \`# gsd-hook-version:\`
3. No match — falls through to the \`else\` branch: \`hookVersion: 'unknown'\`
4. \`isNewer(installed, 'unknown')\` returns true
5. **Stale warning fires on every session** ✗

The fix stamps correctly but the detector still can't read what was stamped.

---

## A useful debugging technique: follow the data path end to end

This bug pattern — fix applied correctly in one place, but another place can't consume the result — is best caught by tracing the full data path rather than validating each piece independently.

In this case: start from the user-visible symptom ("stale warning fires") and follow it backwards to every code path that produces that outcome. The staleness check has three failure modes:
1. Hook has no header → \`hookVersion: 'unknown'\` → stale
2. Hook header uses wrong comment style → regex misses it → \`hookVersion: 'unknown'\` → stale
3. Hook header has unreplaced placeholder → \`{{\`-guard skips → silently undetectable

Your fix eliminates mode 3. Mode 2 was still present. An end-to-end test that actually runs the detector against an installed \`.sh\` hook and asserts \`stale_hooks\` is empty would have caught this — which is why PR #2224 includes exactly that as Part 4 of the regression test suite.

---

## What the final fix does differently

PR #2224 adds all three parts:
1. Headers in the bash hook sources (your approach, carried forward)
2. \`install.js\` reads+substitutes \`{{GSD_VERSION}}\` in \`.sh\` files at install time (your approach, carried forward)
3. Detector regex extended to handle both \`//\` and \`#\` comment syntax (was missing)

The fix also extracts the stale-detection logic from an inline \`node -e '<template literal>'\` subprocess into a dedicated \`gsd-check-update-worker.js\` — which makes the double-escaped regex legible, eliminates an entire category of escaping bugs, and makes the worker independently testable.

Your investigation and documentation of the install.js path were genuinely helpful. Thanks for the detailed PR.

PR #2224, opened 2026-04-14
fix(hooks): complete stale-hooks false-positive fix — stamp .sh version headers + fix detector regex

### LINKED ISSUE #2136: MANAGED_HOOKS list missing bash hooks for staleness check

trek-e:
## Problem

After every fresh install of GSD, the statusline shows:

```
⚠ stale hooks — run /gsd-update
```

Running \`/gsd-update\` bails with \"already on the latest version\". The warning never clears.

## Root Cause (three-way dependency failure)

The stale-hook detector in \`gsd-check-update.js\` checks installed hook files for a \`gsd-hook-version\` header. **Three things were wrong simultaneously:**

1. The three bash hooks (\`gsd-phase-boundary.sh\`, \`gsd-session-state.sh\`, \`gsd-validate-commit.sh\`) shipped with **no version header at all** — so they always fell into the \"unknown / stale\" branch.

2. \`install.js\` performed \`copyFileSync\` for \`.sh\` hooks instead of read+substitute+write, so \`{{GSD_VERSION}}\` would never be replaced at install time — even if the headers were added.

3. The detector regex \`/\\/\\/ gsd-hook-version:\\s*(.+)/\` only matched **JS \`//\` comment syntax**, never bash \`#\` syntax — embedded inside a \`node -e\` template literal where every regex character required double-escaping.

**All three must be fixed together.** Each pair of fixes is insufficient:

| Fix | PR 2207 (regex + headers) | PR 2215 (headers + install stamp) | This PR (all three) |
|---|---|---|---|
| False positive cleared immediately? | ✓ (accidentally — \`{{\`-guard bails on unreplaced placeholder) | ✗ regex still can't read \`#\` syntax | ✓ |
| Future stale detection works after an update? | ✗ bash hooks silently skip staleness forever | ✗ broken | ✓ |

**PR 2207 alone** introduces a permanent silent failure: the \`{{\`-guard suppresses the stale check forever because install never stamps a real version into \`.sh\` files. **PR 2215 alone** stamps correctly but the detector regex still can't see \`#\`-prefixed headers, so the false positive persists on every session.

## Fix (all three parts required)

1. **Bash hook sources**: add \`# gsd-hook-version: {{GSD_VERSION}}\` header on line 2 (after shebang) in all three \`.sh\` hooks
2. **\`install.js\` bundled path**: read+substitute \`{{GSD_VERSION}}\` in \`.sh\` files the same way it already does for \`.js\` hooks
3. **\`install.js\` Codex path**: same substitution for the Codex-specific hook copy block

## Architectural Improvement: Worker Extraction

Beyond the three-part bug fix, the stale-detection logic was extracted from an inline \`node -e '<template literal>'\` subprocess into a dedicated \`gsd-check-update-worker.js\` file. This eliminates the double-escaping problem permanently:

**Before:**
```js
// In gsd-check-update.js — regex had to be double-escaped inside template literal:
const code = `
  const versionMatch = content.match(/\\/\\/ gsd-hook-version:\\s*(.+)/);
`;
spawn(process.execPath, ['-e', code], ...);
```

**After:**
```js
// gsd-check-update.js — spawns the worker by file path, no escaping needed:
const workerPath = path.join(__dirname, 'gsd-check-update-worker.js');
spawn(process.execPath, [workerPath], { env: { GSD_CACHE_FILE, ... } });

// gsd-check-update-worker.js — plain JS, regex is readable:
```

Additional benefits:
- Worker uses \`execFileSync\` (array args) instead of \`execSync\` — safer and passes security hooks
- \`MANAGED_HOOKS\` allowlist is now independently readable by tests without spawning a subprocess
- Worker carries its own \`gsd-hook-version\` header so it participates in staleness detection

## Testing

**New regression test**: \`tests/bug-2136-sh-hook-version.test.cjs\` — 11 assertions across 5 describe blocks:

- Part 1: source bash hooks carry the \`{{GSD_VERSION}}\` placeholder header on line 2
- Part 2: the detector regex in the worker matches \`#\` syntax (and still matches \`//\` — no regression)
- Part 3a: \`install.js\` bundled path reads+substitutes \`.sh\` files at install time
- Part 3b: \`install.js\` Codex path does the same
- Part 4 (E2E): runs the real installer into a temp dir; verifies installed \`.sh\` hooks contain a concrete semver (not the literal placeholder), and the staleness detector reports zero stale hooks on a fresh install

**Existing regression tests updated** to follow the logic that moved to the worker:
- \`tests/orphaned-hooks.test.cjs\`: reads worker file directly for \`MANAGED_HOOKS\` assertions; new test verifies main hook spawns worker by file path (not \`-e\`)
- \`tests/managed-hooks.test.cjs\`: updated \`MANAGED_HOOKS_FILE\` pointer to worker
- \`tests/core.test.cjs\`: stale-hook-path regression test updated to read worker for \`configDir/hooks/\` assertion

**Full suite**: 3886/3886 tests pass.

## Credit

- Regex fix approach: PR #2207 (j2h4u / Maxim Brashenko)
- Install substitution approach: PR #2215 (nitsan2dots)

Closes #2136, #2206, #2209, #2210, #2212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- **Shell hooks falsely flagged as stale on every session** — `gsd-phase-boundary.sh`, `gsd-session-state.sh`, and `gsd-validate-commit.sh` now ship with a `# gsd-hook-version: {{GSD_VERSION}}` header; the installer substitutes `{{GSD_VERSION}}` in `.sh` hooks the same way it does for `.js` hooks; and the stale-hook detector in `gsd-check-update.js` now matches bash `#` comment syntax in addition to JS `//` syntax. All three changes are required together — neither the regex fix alone nor the install fix alone is sufficient to resolve the false positive (#2136, #2206, #2209, #2210, #2212)
+
 ## [1.36.0] - 2026-04-14
 
 ### Added
--- a/bin/install.js
+++ b/bin/install.js
@@ -5761,10 +5761,15 @@ function install(isGlobal, runtime = 'claude') {
             // Ensure hook files are executable (fixes #1162 — missing +x permission)
             try { fs.chmodSync(destFile, 0o755); } catch (e) { /* Windows doesn't support chmod */ }
           } else {
-            fs.copyFileSync(srcFile, destFile);
-            // Ensure .sh hook files are executable (mirrors chmod in build-hooks.js)
+            // .sh hooks carry a gsd-hook-version header so gsd-check-update.js can
+            // detect staleness after updates — stamp the version just like .js hooks.
             if (entry.endsWith('.sh')) {
+              let content = fs.readFileSync(srcFile, 'utf8');
+              content = content.replace(/\{\{GSD_VERSION\}\}/g, pkg.version);
+              fs.writeFileSync(destFile, content);
               try { fs.chmodSync(destFile, 0o755); } catch (e) { /* Windows doesn't support chmod */ }
+            } else {
+              fs.copyFileSync(srcFile, destFile);
             }
           }
         }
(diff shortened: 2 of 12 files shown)


celorodovalho:
Awesome! When do you plan to release? @trek-e

trek-e:
@celorodovalho The fix is merged and will ship in the next tagged release. Watch the [releases page](https://github.com/gsd-build/get-shit-done/releases) or run `/gsd-update` after the release drops.

PR #2299, opened 2026-04-16
feat(do): add weight labels to routing decisions

fr1j0:
## Summary

- Adds a **Weight** column (`🟢 small / 🟡 medium / 🔴 large`) to the `/gsd-do` routing table
- Adds weight definitions below the table so users know exactly what each level implies
- Includes weight labels in ambiguity-resolution options so users can calibrate before committing
- Adds a **Weight** line to the routing display shown before dispatch

## Problem

Users know their task is small but can't tell from command names alone whether `/gsd-do` will trigger an inline fix or a multi-step planning pipeline. This creates hesitation — people avoid `/gsd-do` because they don't trust it won't over-engineer their request.

## Solution

Surface the weight *before* dispatch. The routing display now shows:

```
Weight: 🟢 small — runs inline, no subagents, no planning artifacts
Routing to: /gsd-quick
```

Users can immediately see the cost of the dispatched command and trust the routing.

## Weight definitions

- 🟢 **small** — runs inline, no subagents, no planning artifacts, done in one pass
- 🟡 **medium** — may spawn one agent or produce a lightweight artifact, no full plan/build cycle
- 🔴 **large** — triggers a multi-step pipeline (discuss → plan → execute), creates planning artifacts, may take several minutes

## Scope & Impact

Only changes `workflows/do.md`. No breaking changes — purely additive to the display and routing table. All existing routing logic is preserved.

## ADR / Design Docs

None applicable.

## Screenshots / UI Changes

None. Text-based workflow output only.

## Testing

Routing table entries manually verified against command descriptions in skills directory.

### DIFF
--- a/get-shit-done/workflows/do.md
+++ b/get-shit-done/workflows/do.md
@@ -1,5 +1,5 @@
 <purpose>
-Analyze freeform text from the user and route to the most appropriate GSD command. This is a dispatcher — it never does the work itself. Match user intent to the best command, confirm the routing, and hand off.
+Analyze freeform text from the user and route to the most appropriate GSD command. This is a dispatcher — it never does the work itself. Match user intent to the best command, confirm the routing with a weight label, and hand off.
 </purpose>
 
 <required_reading>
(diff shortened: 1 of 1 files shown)


trek-e:
please re-read all contribution requirements and follow the process.

PR #2315, opened 2026-04-16
fix(statusline): read CLAUDE_CODE_AUTO_COMPACT_WINDOW for context meter

### LINKED ISSUE #2219: gsd-statusline.js: context meter assumes default autocompact buffer (16.5%), wrong when CLAUDE_CODE_AUTO_COMPACT_WINDOW is set
## Summary

`hooks/gsd-statusline.js` hardcodes the autocompact buffer at 16.5% of total context when computing the displayed "used %" meter. This is wrong for users who tune `CLAUDE_CODE_AUTO_COMPACT_WINDOW` to a non-default value, which Claude Code respects but the GSD hook doesn't.

## Repro

1. Set `"env": { "CLAUDE_CODE_AUTO_COMPACT_WINDOW": "400000" }` in `~/.claude/settings.json` (Opus 4.6 1M context — push compact threshold from default ~165k to 400k).
2. Use the session until Claude Code reports `remaining_percentage = 50`.
3. Statusline shows `60%` used (computed against the hardcoded 16.5% buffer), but actual usable context is now 60% of total (40% reserved for autocompact), so it should show `83%` used.

The same miscalibration silently affects the warning thresholds in `gsd-context-monitor.js` because it reads `used_pct` from the bridge file the statusline writes.

## Affected lines

`hooks/gsd-statusline.js`, currently:

\`\`\`js
// Claude Code reserves ~16.5% for autocompact buffer, so usable context
// is 83.5% of the total window. We normalize to show 100% at that point.
const AUTO_COMPACT_BUFFER_PCT = 16.5;
\`\`\`

## Proposed fix

Read `CLAUDE_CODE_AUTO_COMPACT_WINDOW` from the env (Claude Code passes its env to hooks) and compute the buffer % dynamically. Default to 16.5% when the env var is unset, preserving today's behavior:

\`\`\`js
const totalCtx = data.context_window?.total_tokens || 1_000_000;
const acw = parseInt(process.env.CLAUDE_CODE_AUTO_COMPACT_WINDOW || '0', 10);
const AUTO_COMPACT_BUFFER_PCT = acw > 0
  ? Math.min(100, (acw / totalCtx) * 100)
  : 16.5;
\`\`\`

Tested locally: default unchanged (50% remaining → 60% used), and with `CLAUDE_CODE_AUTO_COMPACT_WINDOW=400000` correctly reports 83% used at 50% remaining.

Happy to send a PR if the approach looks right.

willsheldon:
## Summary
- Context meter hardcoded 16.5% autocompact buffer, which is wrong when users set `CLAUDE_CODE_AUTO_COMPACT_WINDOW` to a non-default value
- Now reads the env var (passed through to hooks by Claude Code) and computes buffer % dynamically against `context_window_size`
- Falls back to 16.5% when unset, preserving current behavior

## Test plan
- [ ] Verify default behavior unchanged: unset env, `remaining_percentage=50` → `used=60%`
- [ ] Verify custom window: `CLAUDE_CODE_AUTO_COMPACT_WINDOW=400000`, `remaining_percentage=50` → `used=83%`
- [ ] Verify bridge file `used_pct` matches displayed percentage (context-monitor reads this)

Fixes #2219

### DIFF
--- a/hooks/gsd-statusline.js
+++ b/hooks/gsd-statusline.js
@@ -123,10 +123,15 @@ function runStatusline() {
     const session = data.session_id || '';
     const remaining = data.context_window?.remaining_percentage;
 
-    // Context window display (shows USED percentage scaled to usable context)
-    // Claude Code reserves ~16.5% for autocompact buffer, so usable context
-    // is 83.5% of the total window. We normalize to show 100% at that point.
-    const AUTO_COMPACT_BUFFER_PCT = 16.5;
+    // Context window display (shows USED percentage scaled to usable context).
+    // Claude Code reserves CLAUDE_CODE_AUTO_COMPACT_WINDOW tokens for the
+    // autocompact buffer; default ~165k of 1M = 16.5%. When users tune the
+    // env var, the meter tracks the actual threshold instead of the default.
+    const totalCtx = data.context_window?.context_window_size || 1_000_000;
+    const acw = parseInt(process.env.CLAUDE_CODE_AUTO_COMPACT_WINDOW || '0', 10);
+    const AUTO_COMPACT_BUFFER_PCT = acw > 0
+      ? Math.min(100, (acw / totalCtx) * 100)
+      : 16.5;
     let ctx = '';
     if (remaining != null) {
       // Normalize: subtract buffer from remaining, scale to usable range


trek-e (commented):
This fix was already merged in commit `53078d3` (PR #2219 close) — `gsd-statusline.js` on `main` already reads `CLAUDE_CODE_AUTO_COMPACT_WINDOW` and computes `AUTO_COMPACT_BUFFER_PCT` dynamically against `context_window?.total_tokens`.

One minor difference: this PR uses `data.context_window?.context_window_size` while main uses `data.context_window?.total_tokens`. If `context_window_size` is a more accurate field name for the Claude Code hook data format, it might be worth a follow-up PR to align that field name. But the core fix is already live.

Closing as superseded.

PR #2338, opened 2026-04-17
Fix graphify 0.4 build contract

fishskylky-tech:
## Summary
Align the graphify build contract with graphify 0.4.x actual output.

## Problem
Graphify 0.4.x no longer reliably produces `graphify-out/graph.html`, and the current CLI invocation is `graphify update .` rather than `graphify . --update`.
Because the GSD builder prompt and pre-flight artifact list still assume the old contract, `/gsd-graphify build` can report a failed build even when graphify successfully generated `graph.json` and `GRAPH_REPORT.md`.

## Changes
- remove `graph.html` from the graphify build artifact contract
- update the graphify builder prompt to use `graphify update .`
- update the builder prompt to accept `graph.json` files with `nodes[]` plus either `links[]` or `edges[]`
- update the graphify build unit test to match the current artifact contract

## Validation
- `node --test tests/graphify.test.cjs`
- verified locally against a real graphify 0.4.18 build where status/query/snapshot/diff returned non-zero edge counts and stable results after the contract fix

### DIFF
--- a/commands/gsd/graphify.md
+++ b/commands/gsd/graphify.md
@@ -153,13 +153,13 @@ gsd-tools path: $HOME/.claude/get-shit-done/bin/gsd-tools.cjs
 1. **Invoke graphify:**
    Run from the project root:
    ```
-   graphify . --update
+   graphify update .
    ```
    This builds the knowledge graph with SHA256 incremental caching.
    Timeout: up to 5 minutes (or as configured via graphify.build_timeout).
 
 2. **Validate output:**
-   Check that graphify-out/graph.json exists and is valid JSON with nodes[] and edges[] arrays.
+   Check that graphify-out/graph.json exists and is valid JSON with nodes[] plus either links[] or edges[].
    If graphify exited non-zero or graph.json is not parseable, output:
    ## GRAPHIFY BUILD FAILED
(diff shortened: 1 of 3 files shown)


trek-e (approved):
**Adversarial review — approved.**

Three-file change that aligns the build contract with graphify 0.4.x reality. Reviewed each file:

**commands/gsd/graphify.md** — CLI invocation updated (`graphify update .` vs old `graphify . --update`), `graph.html` removed from copy step, output description updated to reflect `links[]` OR `edges[]`. No remaining references to `graph.html` in the query/status/diff sections of this file. The validation description now accurately matches the relaxed JSON contract.

**get-shit-done/bin/lib/graphify.cjs** — `artifacts` array updated to drop `graph.html`. I verified that `graph.html` is not consumed anywhere else in `graphify.cjs` (query, status, diff logic does not reference it), so no consumers are broken by the removal.

**tests/graphify.test.cjs** — Test updated to match the new artifact list. No coverage gaps introduced.

One observation: the main branch still has the old `graph.html` in `artifacts` and the copy step. This PR correctly removes it — no stale references remain in the PR diff.

trek-e:
you need to create an issue that follows the template before this can be fully merged.

trek-e:
Closing for lack of following process defined in contributing.

PR #2401, opened 2026-04-18
fix(sdk): point checkAgentsInstalled at ~/.claude/agents

### LINKED ISSUE #2400: checkAgentsInstalled() reads wrong directory, reports agents_installed: false when agents are installed
## What happened

`init` query (and anything that calls `withProjectRoot`) reports `agents_installed: false` even when all GSD agents are correctly installed at `~/.claude/agents/`. This cascades into workflows (e.g. `execute-phase`) telling Claude that subagents aren't available, blocking parallel `gsd-executor` spawning.

## Root cause

`sdk/src/query/init.ts:85-86` defaults the agents directory to `~/.claude/get-shit-done/agents/`:

```ts
const agentsDir = process.env.GSD_AGENTS_DIR
  || join(homedir(), '.claude', 'get-shit-done', 'agents');
```

But the installer (`bin/install.js:5704-5771`) and `sdk/src/init-runner.ts:40` write/read agents at `~/.claude/agents/`:

```ts
const GSD_AGENTS_DIR = join(homedir(), '.claude', 'agents');
```

So the installer puts files in one place, and the init-time check looks somewhere else. The stale default appears to be inherited from the old `core.cjs` port (see comment on line 82 of `query/init.ts`) and was never updated to match `init-runner.ts` / the installer.

Built artifact `sdk/dist/query/init.js:77-78` has the same bug.

## Expected

When `~/.claude/agents/gsd-*.md` files are present (as the installer places them), `init` should return `agents_installed: true`.

## Reproduce

1. `npx get-shit-done-cc@latest` (install completes, writes `~/.claude/agents/gsd-*.md`)
2. `ls ~/.claude/agents/ | grep gsd-` → 31 files present
3. Trigger any init query (e.g. `/gsd-execute-phase`)
4. Output includes `agents_installed: false` → workflow refuses to spawn gsd-executor subagents

## Impact

**Blocker** for parallel phase execution. Workarounds exist (`GSD_AGENTS_DIR=~/.claude/agents`) but only if the user knows the bug exists.

## Workaround

Set `GSD_AGENTS_DIR=~/.claude/agents` before running GSD commands.

## Fix

One-line change in `sdk/src/query/init.ts` (rebuild `sdk/dist/query/init.js`):

```ts
const agentsDir = process.env.GSD_AGENTS_DIR
  || join(homedir(), '.claude', 'agents');
```

This matches the installer and `init-runner.ts`.

jeremymcs:
## Linked Issue

Fixes #2400

## What was broken

Every `init` query returned `agents_installed: false` on a clean GSD install, which made `/gsd-execute-phase` (and anything else that gates on `agents_installed`) refuse to spawn `gsd-executor` subagents even when all GSD agents were on disk.

## What this fix does

Change the default agents directory in `checkAgentsInstalled` from `~/.claude/get-shit-done/agents/` to `~/.claude/agents/` so it matches where the installer and `init-runner.ts` actually put agents.

## Root cause

`sdk/src/query/init.ts:85-86` was ported from the old `core.cjs` (see comment on line 82) and kept the pre-migration path. The installer (`bin/install.js:5704-5771`) and `sdk/src/init-runner.ts:40` were updated to `~/.claude/agents/`, but the init-time check was not. Verified in this repo: 31 `gsd-*.md` files exist at `~/.claude/agents/` and zero at `~/.claude/get-shit-done/agents/`, so the check returned `false` for every profile key in `MODEL_PROFILES`.

### How I verified the fix

- Added two regression tests in `sdk/src/query/init.test.ts` that drive `withProjectRoot` via `GSD_AGENTS_DIR`:
  - Populated dir with stubs for every `MODEL_PROFILES` key → `agents_installed: true`, `missing_agents: []`
  - Empty dir → `agents_installed: false`, non-empty `missing_agents`
- `npx vitest run src/query/init.test.ts` → 24/24 pass.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] N/A

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] N/A

---

## Checklist

- [ ] Linked issue has the `confirmed-bug` label *(awaiting maintainer triage on #2400)*
- [x] Regression test added
- [x] Existing tests pass (`init.test.ts`: 24/24)
- [x] CHANGELOG.md updated under `[Unreleased] › Fixed`
- [x] No dependencies added

## Breaking changes

None. `GSD_AGENTS_DIR` override behavior is unchanged; only the default path is corrected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent installation detection corrected so clean installs no longer report agents as missing
  * Workflow reference updated to use the correct default agents directory

* **New Features**
  * Installer now automatically installs and configures the SDK on PATH
  * New installer flags to opt in/out of SDK installation (--sdk / --no-sdk)


jeremymcs:
Swept two more stale references from the same bug class (commit b4608ce):

- **`get-shit-done/workflows/profile-user.md:12`** — the `/gsd-profile-user` workflow's required-reading `@`-import pointed at `$HOME/.claude/get-shit-done/agents/gsd-user-profiler.md`. Every other workflow (e.g. `discuss-phase.md:661`, `validate-phase.md:97`) uses `$HOME/.claude/agents/`. Fixed to match.
- **`sdk/src/init-runner.ts:40`** — `GSD_AGENTS_DIR` was a hardcoded constant with no env override. `sdk/src/query/init.ts:85` already honored the env var; now both do. Keeps SDK consumers able to redirect lookups consistently.

Full SDK unit suite: 1245/1245 pass.

jeremymcs:
Codex adversarial review surfaced a follow-up (commit 819e164): the prior default was hardcoded to `~/.claude/agents`, so users with `CLAUDE_CONFIG_DIR` set would still false-negative. Fixed by extracting `resolveAgentsDir()` into `query/helpers.ts` with the installer's precedence (`bin/install.js:389-396`):

1. `GSD_AGENTS_DIR` (SDK override)
2. `CLAUDE_CONFIG_DIR/agents`
3. `~/.claude/agents`

Both `query/init.ts` and `init-runner.ts` now share the resolver — eliminates the SDK-vs-installer drift that caused #2400 in the first place.

Added two regression tests covering `CLAUDE_CONFIG_DIR` and precedence. Full unit suite: 1247/1247.

Scope is intentionally limited to Claude Code — the SDK is Claude-first by design (`init-runner.ts`, `GSD_TEMPLATES_DIR`). A separate enhancement issue will cover making `checkAgentsInstalled` fully runtime-aware (honoring `CODEX_HOME`, `OPENCODE_CONFIG_DIR`, `GEMINI_CONFIG_DIR`, etc.).

PR #2455, opened 2026-04-20
fix(sdk): extractCurrentMilestone Backlog leak + state.begin-phase flag parsing

### LINKED ISSUE #2420: [sdk] state.begin-phase handler ignores --phase/--name/--plans flags and writes literal strings to STATE.md
## Summary

`gsd-sdk query state.begin-phase` invoked with flag-form args (which is how the shipped `execute-phase.md` workflow calls it) passes the flags through as literal positional values. The handler is positional-only and has no flag-parsing step, so STATE.md ends up with strings like `Phase: --phase (04) — EXECUTING` and `Plan: 1 of --name`.

## Environment

- `@gsd-build/sdk` v0.1.0 (installed via Homebrew)
- Node v23.6.1
- Darwin 25.2.0

## Reproduction

```bash
$ gsd-sdk query state.begin-phase --phase 99 --name "probe-test" --plans 1
{"phase":"--phase","name":"99","plan_count":"--name"}
```

STATE.md after the call:

```
Phase: --phase (99) — EXECUTING
Plan: 1 of --name
Status: Executing Phase --phase
Last activity: 2026-04-19 -- Phase --phase execution started
```

## Expected

The shipped workflow at `.claude/get-shit-done/workflows/execute-phase.md:228` invokes the command with flags:

```bash
gsd-sdk query state.begin-phase --phase "${PHASE_NUMBER}" --name "${PHASE_NAME}" --plans "${PLAN_COUNT}"
```

The handler should accept that form alongside the positional form (currently only positional works).

## Root cause

In `sdk/dist/query/state-mutation.js` (compiled from `sdk/src/query/state-mutation.ts`), `stateBeginPhase`:

```javascript
export const stateBeginPhase = async (args, projectDir) => {
    const phaseNumber = args[0];   // "--phase" when invoked with flags
    const phaseName = args[1] || '';  // e.g. "99"
    const planCount = args[2] || '?';  // "--name"
    ...
};
```

No flag-parsing step before positional assignment. The executor workflow (`execute-phase.md`) and the downstream STATE.md mutation are wired with the flag form in mind, so they silently produce broken output.

## Suggested fix

Accept both forms at the top of the handler, in a backward-compatible way:

```javascript
export const stateBeginPhase = async (args, projectDir) => {
    const flags = {};
    if (typeof args[0] === 'string' && args[0].startsWith('--')) {
        for (let i = 0; i < args.length - 1; i += 2) {
            if (typeof args[i] === 'string' && args[i].startsWith('--')) {
                flags[args[i].slice(2)] = args[i + 1];
            }
        }
    }
    const phaseNumber = flags.phase ?? args[0];
    const phaseName = flags.name ?? args[1] ?? '';
    const planCount = flags.plans ?? flags['plan-count'] ?? args[2] ?? '?';
    // ... rest unchanged
};
```

Positional calls still work; flag calls now parse correctly. Verified locally: `gsd-sdk query state.begin-phase 99 "probe-test" 1` returns `{"phase":"99","name":"probe-test","plan_count":"1"}` and writes a clean STATE.md.

## Impact

Any project driving `/gsd:execute-phase` has its STATE.md polluted with literal `--phase`/`--name` strings on the first begin-phase call of each run. Manual recovery (`git checkout .planning/STATE.md`) is required before the phase can advance cleanly. Discovered while running Phase 4 of a real project — the pollution survives into the next plan's SUMMARY commit if not caught.

## Local workaround

Until the handler is fixed, the workflow can call with positional args:

```bash
gsd-sdk query state.begin-phase "${PHASE_NUMBER}" "${PHASE_NAME}" "${PLAN_COUNT}"
```

which the current handler already accepts.

### LINKED ISSUE #2422: extractCurrentMilestone leaks ## Backlog and truncates at same-version sub-headings
## Summary

`extractCurrentMilestone()` in `@gsd-build/sdk@0.1.0` (`dist/query/roadmap.js:74-119`) has two bugs that compound: the current milestone slice loses its detail sections and gains the entire `## Backlog`. `init.manager` then reports 999.x backlog phases as if they were the current milestone's phases.

## Observed

With `STATE.md` set to `milestone: v2.0` and a ROADMAP containing a `### 🚧 v2.0 …` heading followed by a `## v2.0 Phase Details` sub-heading and `### Phase 999.x:` backlog items earlier in the file:

```bash
$ gsd-sdk query init.manager | jq '{phase_count, phases: [.phases[].number]}'
{
  "phase_count": 37,
  "phases": ["999.2", "999.3", "999.4", "…35 more 999.x backlog items…"]
}
```

Expected: the 4 v2.0 phases (1074–1077). Got: 37 backlog items. None of the actual v2.0 phases appear.

## Minimal repro

```md
# ROADMAP.md
## Backlog
### Phase 999.1: Parking lot item A
### Phase 999.2: Parking lot item B

### 🚧 v2.0 My Milestone (In Progress)
- [ ] **Phase 100: Real work**

## v2.0 Phase Details
### Phase 100: Real work
**Goal**: …
```

```yaml
# STATE.md frontmatter
milestone: v2.0
```

Calling `initManager` with the above produces `phase_count: 2` (the 999.x items), NOT `phase_count: 1` (Phase 100).

## Root cause

`dist/query/roadmap.js:100-118`:

```js
const sectionStart = sectionMatch.index;
const headingLevelMatch = sectionMatch[1].match(/^(#{1,3})\s/);
const headingLevel = headingLevelMatch ? headingLevelMatch[1].length : 2;
const restContent = content.slice(sectionStart + sectionMatch[0].length);
const nextMilestonePattern = new RegExp(`^#{1,${headingLevel}}\\s+(?:.*v\\d+\\.\\d+|✅|📋|🚧)`, 'mi');
const nextMatch = restContent.match(nextMilestonePattern);
// … computes sectionEnd from nextMatch …
const beforeMilestones = content.slice(0, sectionStart);
const currentSection = content.slice(sectionStart, sectionEnd);
const preamble = beforeMilestones.replace(/<details>[\s\S]*?<\/details>/gi, '');
return preamble + currentSection;
```

**Bug 1 — preamble leaks the Backlog section.**
`preamble` strips `<details>` blocks, but `## Backlog` and `### Phase 999.x:` are typically NOT inside `<details>`, so every backlog phase header ends up in `initManager`'s phase pattern scan.

**Bug 2 — `nextMilestonePattern` truncates at same-milestone sub-headings.**
The pattern `.*v\d+\.\d+` matches `## v2.0 Phase Details` (the milestone's own detail sub-section), so `sectionEnd` lands before any `### Phase 100:` detail heading. Combined with Bug 1, the output contains only preamble phases (backlog) and none of the current milestone's actual detail sections.

## Proposed fix

1. Return only `currentSection`, not `preamble + currentSection`. The preamble is never a reliable source of current-milestone data.
2. Extract the current version from `sectionMatch[1]` and exclude it from the next-milestone scan — only stop at a heading whose version is different, or one bearing a milestone emoji.

Patch sketch:

```js
const currentVersionMatch = sectionMatch[1].match(/v(\d+\.\d+)/);
const currentVersionStr = currentVersionMatch ? currentVersionMatch[1] : '';
const nextMilestoneRegex = new RegExp(
  `^#{1,${headingLevel}}\\s+(?:.*v(\\d+\\.\\d+)[^\\n]*|.*(?:✅|📋|🚧))`, 'gm'
);
let sectionEnd = content.length;
let m;
while ((m = nextMilestoneRegex.exec(restContent)) !== null) {
  const matchedVersion = m[1];
  if (matchedVersion && currentVersionStr && matchedVersion === currentVersionStr) continue;
  sectionEnd = sectionStart + sectionMatch[0].length + m.index;
  break;
}
return content.slice(sectionStart, sectionEnd);
```

Working implementation (verified against a real project with 37 backlog items → correctly collapses to the 4 in-progress phases):
https://github.com/shlee87/gsd-patches/blob/main/scripts/reapply-gsd-local-patches.cjs (search for `sdk-milestone-slice`).

## Environment

- `@gsd-build/sdk` 0.1.0
- macOS 14 (Darwin 25.2.0), Node 20+
- Called via `gsd-sdk query init.manager`

## Related

- #2409 (stale `current_milestone` when latest is shipped) — different root cause but same function family. Fixing either shouldn't block the other.

trek-e:
Closes #2422
Closes #2420

## Root cause

**#2422 — `extractCurrentMilestone`** had two compounding bugs:
1. **Backlog leak**: the function returned `preamble + currentSection`, where `preamble` is everything before the current milestone heading. Any `## Backlog` / `### Phase 999.x:` lines in that preamble were fed into the phase-pattern scanner, producing 37 backlog phases instead of the real milestone phases.
2. **Same-version truncation**: `nextMilestonePattern` matched `## v2.0 Phase Details` (a sub-heading belonging to the current milestone), cutting off the section before any `### Phase N:` detail headings could be scanned.

**#2420 — `stateBeginPhase`** read `args[0..2]` as positional values with no flag-parsing step. When called as `state.begin-phase --phase 04 --name "Title" --plans 3` (the form used by `execute-phase.md`), `phaseNumber` became `"--phase"`, `phaseName` became `"04"`, and `planCount` became `"--name"`, corrupting STATE.md on every phase start.

## Fix

**#2422** (`sdk/src/query/roadmap.ts` — `extractCurrentMilestone`):
- Removed the `preamble + currentSection` concatenation; returns only `currentSection`.
- Replaced the single-shot `nextMilestonePattern` with a `while` loop using a capturing-group regex that extracts the matched version. Same-version sub-headings (e.g. `## v2.0 Phase Details`) are skipped; only headings with a different version or a milestone emoji stop the scan.

**#2420** (`sdk/src/query/state-mutation.ts` — `stateBeginPhase`):
- Added flag-parsing at the top of the handler: if `args[0]` starts with `--`, iterate pairs and build a `flags` map; then resolve `phaseNumber`, `phaseName`, `planCount` from `flags` with positional fallbacks.
- Positional calls continue to work unchanged.

## Tests

New failing-first tests in `sdk/src/query/`:

- `roadmap.test.ts`: `bug-2422: does not include ## Backlog section before the current milestone`
- `roadmap.test.ts`: `bug-2422: does not truncate at same-version sub-heading (## v2.0 Phase Details)`
- `state-mutation.test.ts`: `bug-2420: parses --phase/--name/--plans flag-form args correctly`
- `state-mutation.test.ts`: `bug-2420: positional args still work after flag-parsing fix`

SDK test result: `PASS (47) FAIL (0)`. Full suite: exit 0, 86.99% line coverage (threshold 70%).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed milestone content extraction to exclude backlog sections and properly handle version-specific subsections
  * Enabled phase initialization to accept both flag-style (`--phase`, `--name`, `--plans`) and traditional positional arguments

* **Tests**
  * Added unit tests for milestone extraction scenarios
  * Added unit tests for phase initialization with multiple argument formats

### DIFF
--- a/sdk/src/query/roadmap.test.ts
+++ b/sdk/src/query/roadmap.test.ts
@@ -144,6 +144,57 @@ describe('extractCurrentMilestone', () => {
     const result = await extractCurrentMilestone(content, tmpDir);
     expect(result).toBe('current content');
   });
+
+  // ─── Bug #2422: preamble Backlog leak ─────────────────────────────────
+  it('bug-2422: does not include ## Backlog section before the current milestone', async () => {
+    const roadmapWithBacklog = `# ROADMAP
+
+## Backlog
+### Phase 999.1: Parking lot item A
+### Phase 999.2: Parking lot item B
+
+### 🚧 v2.0 My Milestone (In Progress)
+- [ ] **Phase 100: Real work**
+
+## v2.0 Phase Details
+### Phase 100: Real work
+**Goal**: Do stuff.
+`;
+    const state = `---\nmilestone: v2.0\n---\n# State\n`;
+    await writeFile(join(tmpDir, '.planning', 'STATE.md'), state);
+    await writeFile(join(tmpDir, '.planning', 'ROADMAP.md'), roadmapWithBacklog);
+
+    const result = await extractCurrentMilestone(roadmapWithBacklog, tmpDir);
+
+    // Must NOT include backlog phases
+    expect(result).not.toContain('Phase 999.1');
+    expect(result).not.toContain('Phase 999.2');
+    expect(result).not.toContain('Parking lot');
+    // Must include the actual v2.0 content
+    expect(result).toContain('Phase 100');
+  });
+
+  // ─── Bug #2422: same-version sub-heading truncation ───────────────────
+  it('bug-2422: does not truncate at same-version sub-heading (## v2.0 Phase Details)', async () => {
+    const roadmapWithDetails = `# ROADMAP
+
+### 🚧 v2.0 My Milestone (In Progress)
+- [ ] **Phase 100: Real work**
+
+## v2.0 Phase Details
+### Phase 100: Real work
+**Goal**: Do stuff.
+`;
+    const state = `---\nmilestone: v2.0\n---\n# State\n`;
+    await writeFile(join(tmpDir, '.planning', 'STATE.md'), state);
+    await writeFile(join(tmpDir, '.planning', 'ROADMAP.md'), roadmapWithDetails);
+
+    const result = await extractCurrentMilestone(roadmapWithDetails, tmpDir);
+
+    // The detail section must survive — not be cut off
+    expect(result).toContain('Phase 100');
+    expect(result).toContain('Phase Details');
+  });
 });
 
 // ─── roadmapGetPhase ──────────────────────────────────────────────────────
--- a/sdk/src/query/roadmap.ts
+++ b/sdk/src/query/roadmap.ts
@@ -110,7 +110,7 @@ export async function extractCurrentMilestone(content: string, projectDir: strin
 
   // Fallback: derive from ROADMAP in-progress marker
   if (!version) {
-    const inProgressMatch = content.match(/🚧\s*\*\*v(\d+\.\d+)\s/);
+    const inProgressMatch = content.match(/🚧\s*\*\*v(\d+(?:\.\d+)+)\s/);
     if (inProgressMatch) {
       version = 'v' + inProgressMatch[1];
     }
@@ -137,11 +137,12 @@ export async function extractCurrentMilestone(content: string, projectDir: strin
   const restContent = content.slice(sectionStart + sectionMatch[0].length);
 
   // Extract current version so same-version sub-headings are not treated as boundaries.
-  const currentVersionMatch = version ? version.match(/v(\d+\.\d+)/) : null;
+  // Capture full semver (major.minor.patch) so v2.0.1 is not collapsed to "2.0".
+  const currentVersionMatch = version ? version.match(/v(\d+(?:\.\d+)+)/i) : null;
   const currentVersionStr = currentVersionMatch ? currentVersionMatch[1] : '';
 
   const nextMilestoneRegex = new RegExp(
-    `^#{1,${headingLevel}}\\s+(?:.*v(\\d+\\.\\d+)[^\\n]*|.*(?:✅|📋|🚧))`,
+    `^#{1,${headingLevel}}\\s+(?:.*v(\\d+(?:\\.\\d+)+)[^\\n]*|.*(?:✅|📋|🚧))`,
     'gm'
   );
 
--- a/sdk/src/query/roadmap.ts
+++ b/sdk/src/query/roadmap.ts
@@ -130,30 +130,34 @@ export async function extractCurrentMilestone(content: string, projectDir: strin
 
   const sectionStart = sectionMatch.index;
 
-  // Find end: next milestone heading at same or higher level, or EOF
+  // Find end: next milestone heading at same or higher level, or EOF.
+  // Skip headings that belong to the SAME version (e.g. "## v2.0 Phase Details").
   const headingLevelMatch = sectionMatch[1].match(/^(#{1,3})\s/);
   const headingLevel = headingLevelMatch ? headingLevelMatch[1].length : 2;
   const restContent = content.slice(sectionStart + sectionMatch[0].length);
-  const nextMilestonePattern = new RegExp(
-    `^#{1,${headingLevel}}\\s+(?:.*v\\d+\\.\\d+|✅|📋|🚧)`,
-    'mi'
-  );
-  const nextMatch = restContent.match(nextMilestonePattern);
 
-  let sectionEnd: number;
-  if (nextMatch && nextMatch.index !== undefined) {
-    sectionEnd = sectionStart + sectionMatch[0].length + nextMatch.index;
-  } else {
-    sectionEnd = content.length;
+  // Extract current version so same-version sub-headings are not treated as boundaries.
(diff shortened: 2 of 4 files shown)


trek-e:
Self-review — two issues need to be fixed before this merges.

## 1. Flag parser silently misassigns when a value is missing (state-mutation.ts)

The new flag-parsing loop in `stateBeginPhase` steps `i` by 2 unconditionally:

```ts
for (let i = 0; i < args.length - 1; i += 2) {
  if (typeof args[i] === 'string' && args[i].startsWith('--')) {
    flags[args[i].slice(2)] = args[i + 1];
  }
}
```

If a flag is passed without a value — e.g. `['--phase', '--name', 'Title', '--plans', '1']` (omitted value for `--phase`) — the loop assigns `flags['phase'] = '--name'`, then skips `'Title'` entirely, and assigns `flags['name'] = '--plans'`. Result: `phaseNumber = '--name'`, `phaseName = '--plans'`. STATE.md is corrupted in a different way than the original bug.

Fix: before assigning, verify `args[i + 1]` is not itself a flag token:

```ts
for (let i = 0; i < args.length - 1; i += 2) {
  const key = args[i];
  const val = args[i + 1];
  if (typeof key === 'string' && key.startsWith('--')
      && typeof val === 'string' && !val.startsWith('--')) {
    flags[key.slice(2)] = val;
  }
}
```

Add a test for the missing-value case.

## 2. Version regex only captures major.minor — edge case for patch milestones (roadmap.ts)

Both `currentVersionMatch` and `nextMilestoneRegex` capture only `\d+\.\d+`. For a `v2.0.1` milestone, `currentVersionStr` becomes `'2.0'`. A heading `## v2.0.1 Phase Details` would match `nextMilestoneRegex`, capture `'2.0'` (not `'2.0.1'`), compare equal to `currentVersionStr = '2.0'`, and be incorrectly skipped. The section would then run past its boundary.

Fix: extend both patterns to `v(\d+\.\d+(?:\.\d+)?)`.

The core logic for the two reported bugs (#2422, #2420) is correct. These are two follow-on correctness issues that should land before merge.

Issue #2439, opened 2026-04-19
Command not found: gsd-sdk


ssouza:
### GSD Version

1.38.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v22.21.1

### Shell

/bin/zsh

### Installation Method

npx get-shit-done-cc@latest (fresh run)

### What happened?

When running /gsd:set-profile balanced in Claude Code, the workflow attempts to execute:

However, this fails with:
```
Error: Shell command failed for pattern "!gsd-sdk query config-set-model-profile balanced
      --raw": [stderr]                                                                        
     (eval):1: command not found: gsd-sdk
```

### What did you expect?

The gsd-sdk query command should be available and execute successfully.

### Steps to reproduce

1. In a Claude Code session with the gsd-build workflow configured
2. Run /gsd:set-profile balanced
3. Observe the bash error

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

### Runtime settings.json (if relevant)

```json

```

### How often does this happen?

Every time (100% reproducible)

### Impact

Minor — Cosmetic or edge case

### Privacy Checklist

- [x] I have reviewed all pasted output for PII (usernames, paths, API keys) and redacted where necessary

Issue #2451, opened 2026-04-19
Context monitor hook reports higher usage than actual (78% vs 65% real)


Pseudoataraxia:
## Summary

The GSD context monitor hook (`gsd-context-monitor.js`) consistently over-reports context usage compared to CC's native `/context` command. This is a recurring issue, not a one-off observation.

## Examples

| Hook says | CC `/context` says | Gap |
|---|---|---|
| 78% | 65% | +13 points |
| 80% | ~67% | +13 points |
| 90%+ | ~75% | +15 points |

The over-reporting is consistent throughout sessions, not just at high usage. The hook appears to systematically inflate the percentage.

## Impact

- Premature "context getting limited" warnings when 30%+ context remains
- Users stop work or prepare to pause unnecessarily
- The "Avoid starting new complex work" message triggers too early, disrupting flow

## Environment

- Claude Code: latest (2026-04-19)
- GSD: v1.38.0
- Model: claude-opus-4-6[1m] (1M context)
- Platform: macOS Darwin 25.3.0

## Expected behavior

Hook context percentage should be close to CC's native `/context` reporting (within ~5 points), or the warning thresholds should be calibrated to account for the known offset.

PR #2457, opened 2026-04-20
fix(sdk): decouple from build-from-source install, close #2441 #2453

### LINKED ISSUE #2441: installer: escalate or recover when gsd-sdk is installed but off-PATH after #2386
## Context
#2386 added `installSdkIfNeeded()` to `bin/install.js` which builds `@gsd-build/sdk` from the in-repo `sdk/` source and runs `npm install -g .`. It then verifies with \`which gsd-sdk\`.

When verification fails (npm's global bin is not on the user's PATH — common on macOS without explicit \`npm config prefix\` configuration), the installer currently only prints a yellow warning and returns without failing:

https://github.com/gsd-build/get-shit-done/blob/main/bin/install.js#L6711-L6722

The install proceeds successfully, and the user sees a \"✓ install complete\" summary. But every \`/gsd-*\` command will fail with \`command not found: gsd-sdk\` until they fix their PATH.

## Evidence
#2439 reported \`/gsd:set-profile balanced\` failing with \`command not found: gsd-sdk\` on v1.38.0 after a fresh \`npx get-shit-done-cc@latest\` on macOS. #2386 should have covered that flow.

Two plausible causes:
1. npm global bin off-PATH → installer warned, user missed it.
2. Install.js didn't run \`installSdkIfNeeded()\` at all in that flow (npx entry path, SDK source tree missing at install time, etc.).

We shipped PR #2440 as a defensive pre-flight on \`/gsd:set-profile\` (parity with \`/gsd-quick\`'s #2334 guard), but the root-cause issue is in the installer.

## Proposed investigation
- [ ] Ask #2439 reporter for their installer output (look for the yellow SDK warning, or its absence)
- [ ] Audit all \`/gsd-*\` command bodies — how many still lack a pre-flight and will crash the same way?
- [ ] Decide installer policy when SDK ends up off-PATH:
  - (a) Fail the install (exit non-zero) and print remediation
  - (b) Print a loud banner (red, boxed) at end-of-install with exact \`export PATH=…\` remediation for the user's shell
  - (c) Detect npm global bin dir and offer to add it to shell rc automatically
- [ ] Consider a post-install smoke: run \`gsd-sdk query --help\` and print a banner if it fails

## Related
- #2334 — /gsd-quick pre-flight
- #2386 — SDK auto-install in bin/install.js
- #2439 — /gsd:set-profile crash (user-facing symptom)
- #2440 — /gsd:set-profile defensive pre-flight (this issue is the root cause)

## Impact
Minor/Medium — users who hit the off-PATH case get opaque errors from every \`/gsd-*\` command until they manually remediate. Each affected command needs its own pre-flight guard otherwise.

### LINKED ISSUE #2453: installSdkIfNeeded: tsc-built dist/cli.js lands with mode 644 on fresh installs, leaving gsd-sdk non-executable
## Summary

On a fresh install (`npx get-shit-done-cc@latest --claude --global`), the installer's `installSdkIfNeeded()` step builds `sdk/dist/cli.js` via `tsc` and then runs `npm install -g .`. The resulting globally-installed `dist/cli.js` lands with mode **`644`** — the `gsd-sdk` bin symlink in the nvm bin dir points at a non-executable target, so `command -v gsd-sdk` fails and invocation returns `permission denied: gsd-sdk`.

The installer's own verifier (`resolveGsdSdk()`) then correctly reports failure and prints `✗ GSD SDK install failed — gsd-sdk not on your PATH` — but the error message pins the blame on PATH, not on file mode, so users end up editing `~/.zshrc` to add a directory that's already on PATH.

This broke every `/gsd-*` command on 1.38.1 on my machine until I manually chmod'd the CLI — and it will break for every other user on first install too.

## Environment

- GSD: 1.38.1 (released ~2026-04-19, fresh from `npx get-shit-done-cc@latest`)
- `@gsd-build/sdk`: 0.1.0 (built from in-repo `sdk/` source by the installer)
- OS: macOS 15.3 (Darwin 25.3.0)
- Node: v24.6.0 via nvm
- npm: 10.x (bundled with node)
- Shell: zsh

## Reproduction

```
$ npx -y get-shit-done-cc@latest --claude --global
...
  Building GSD SDK from source (…/get-shit-done-cc/sdk)…
added 123 packages, audited 124 packages in 841ms
> @gsd-build/sdk@0.1.0 build
> tsc
up to date in 118ms
  gsd-sdk was installed to:
    /Users/$USER/.nvm/versions/node/v24.6.0/bin
```

But the bin dir **IS** on PATH (`which node` resolves fine from the same directory). Checking the symlink target:

```
$ ls -la /Users/$USER/.nvm/versions/node/v24.6.0/lib/node_modules/@gsd-build/sdk/dist/cli.js
-rw-r--r--@ 1 $USER staff 19607 Apr 19 17:09 .../dist/cli.js

$ head -1 .../dist/cli.js
#!/usr/bin/env node

$ command -v gsd-sdk
(empty — zsh rejects non-executable symlink target)

$ /Users/$USER/.nvm/versions/node/v24.6.0/bin/gsd-sdk
zsh: permission denied: /Users/$USER/.nvm/versions/node/v24.6.0/bin/gsd-sdk
```

## Workaround

```
chmod +x /Users/$USER/.nvm/versions/node/v24.6.0/lib/node_modules/@gsd-build/sdk/dist/cli.js
```

After this, `gsd-sdk v0.1.0` resolves cleanly. Subsequent `/gsd-update` / `--sdk` runs preserve the fix (tsc's `O_TRUNC` overwrite keeps the existing mode), so the bug only bites **first installs** — but that's every new user.

## Root cause

- `sdk/package.json` has `"build": "tsc"` — `tsc` emits all outputs at the process default mode (`0o666 & ~umask` → `644`).
- The `bin` field (`"gsd-sdk": "./dist/cli.js"`) tells npm to create a symlink, but `npm install -g .` does **not** chmod the bin target when the source file was produced by a build script rather than extracted from a tarball. (npm chmods bin entries when installing from a packed tarball, but not always when installing from a local directory with an uncommitted build output — behavior varies by npm version.)
- The shebang `#!/usr/bin/env node` is present, so the only thing missing is the execute bit.

## Suggested fix (any one of these)

**Option A — SDK package.json (most portable):**
```json
"scripts": {
  "build": "tsc && node -e \"require('fs').chmodSync('dist/cli.js', 0o755)\""
}
```

**Option B — installer (most resilient, doesn't require SDK republish):**
In `bin/install.js` `installSdkIfNeeded()`, after step 4 (`resolveGsdSdk()` succeeds) or immediately after the `npm install -g .` step, add:
```js
const globalSdkDir = path.dirname(require.resolve('@gsd-build/sdk/package.json', { paths: [globalNodeModules] }));
const cliPath = path.join(globalSdkDir, 'dist', 'cli.js');
try { fs.chmodSync(cliPath, 0o755); } catch (e) { /* Windows / non-existent */ }
```

The installer already has this exact pattern four times for hook files (e.g. `bin/install.js:5838`, `:5846`, `:5959`, `:5965`). Mirroring it for the SDK CLI would be a one-line fix.

**Option C — SDK's `tsconfig.json`:** not viable; TypeScript has no `chmod` primitive.

## Impact

- Every new user on 1.38.1 hits this on first install.
- The error message ("not on your PATH") misdirects users toward a PATH fix that won't help.
- All `/gsd-*` commands silently degrade (fallback `|| echo ...` patterns swallow the `command not found` output, so workflows run but `init.quick`, `agent-skills`, etc. return garbage).

Happy to send a PR for Option B if the maintainers confirm it's the preferred path.

jeremymcs:
## Summary

Final, root-cause fix for the recurring `gsd-sdk` install fragility (#2441, #2453). Replaces the on-user-machine `tsc` + `npm install -g ./sdk` dance with a **prebuilt `sdk/dist/` shipped inside the parent `get-shit-done-cc` tarball**, exposed via a `bin/gsd-sdk.js` shim in the parent package.

Closes #2441. Closes #2453.

## Why a rethink instead of another patch

This is the 5th PR on the same surface:
- #2386 — introduced `installSdkIfNeeded()` (build-from-source + `npm install -g .`).
- #2440 — per-command `command -v gsd-sdk` guard.
- #2449 — made install failures fatal, added `install-smoke.yml` CI.
- #2443 — handler registry drift guard.
- #2453 (now) — `tsc` emits mode 644; `npm install -g <dir>` doesn't reliably chmod bin.

Each patch hardened a symptom. The shared root cause is "build-from-source on user machine + global-bin install" — every layer (PATH, chmod, npm version behavior, `NPM_CONFIG_PREFIX` RO, `ignore-scripts`, corepack) is hostile terrain. This PR removes that foundation.

Peer-reviewed by Codex (high confidence on architecture; flagged call-site audit + CI gate blind spot — both addressed below).

## Changes

**New: `bin/gsd-sdk.js` (32 lines)** — parent-package bin entry that resolves `sdk/dist/cli.js` and delegates via `node`. npm correctly chmods bin entries from tarballs, so the exec-bit problem in #2453 cannot occur here.

**`package.json`** — declares `bin.gsd-sdk`; adds `sdk/dist` to `files`; new `build:sdk` script; `prepublishOnly` now runs hooks + SDK builds; version `1.38.2`.

**`sdk/package.json`** — `prepublishOnly` is `rm -rf dist && tsc && chmod +x dist/cli.js` (the `rm -rf` guards against npm/npm#15147 stale-prepublishOnly bug).

**`bin/install.js` — 162 lines deleted** — `installSdkIfNeeded()`, `resolveGsdSdk()`, `detectShellRc()`, `emitSdkFatal()`. No more PATH fragility, no more chmod fragility, no more 3-step global install.

**`.github/workflows/release.yml`** — `npm run build:sdk` step before `npm publish` in both rc and finalize jobs, so every published tarball ships fresh dist.

**`.github/workflows/install-smoke.yml`** — new `smoke-unpacked` variant that installs from the unpacked dir with exec bit stripped. The original smoke used `npm pack` which normalizes file modes and therefore could not catch #2453-class bugs. This variant closes that gate.

**Tests** — new `tests/bug-2441-sdk-decouple.test.cjs` (9 tests) asserts: installer never invokes `npm install -g` for the SDK, `gsd-sdk` resolves to the shim, dist path is correctly resolved. Updated `tests/bugs-1656-1657.test.cjs` (stale assertions).

## Call-site audit

109 files reference `gsd-sdk` (17 slash commands, 12 agents, 72 workflow docs, 8 tests/infra). **Zero migration needed** — all are user-facing prose/shell snippets that invoke `gsd-sdk` on PATH at AI-runtime. The shim keeps `gsd-sdk` on PATH transparently.

## `--sdk` flag semantics

Previously forced a rebuild. Now verifies the bundled `sdk/dist/` is resolvable. Users who invoked `get-shit-done-cc --sdk` as "force rebuild" no longer need it — dist ships prebuilt.

## What this closes for good

- #2439-class errors (`command not found: gsd-sdk`) — no global install, no PATH dep.
- #2453-class errors (mode 644 on `dist/cli.js`) — `node` doesn't care about exec bit; parent shim is chmodded from the tarball.
- Follow-ups in #2441 peer review: `NPM_CONFIG_PREFIX` RO, `ignore-scripts=true`, corepack/pnpm-as-default, air-gapped envs — all eliminated because no `npm install` runs at user-install time.

## Test plan

- [x] `node scripts/run-tests.cjs` — 4170/4170 pass (baseline 4157, +13 new).
- [x] `node -c bin/install.js` / `node -c bin/gsd-sdk.js` — syntax OK.
- [ ] `install-smoke.yml` self-validates on this PR (smoke-packed + smoke-unpacked both green).
- [ ] Manual smoke: `npm pack && npm install -g <tarball>` → `gsd-sdk --version` resolves.
- [ ] Manual smoke: unpack tarball, strip exec bit on `sdk/dist/cli.js`, install — shim still works (it invokes via `node`).

## Summary by CodeRabbit

* **New Features**
  * Added a new gsd-sdk CLI command (shim delegating to the bundled SDK).

* **Bug Fixes**
  * Installer now relies on a bundled, prebuilt SDK distribution; SDK readiness is validated at install.
  * `--sdk` semantics updated to verify bundled SDK instead of forcing a source rebuild.

* **Chores**
  * Version bumped to 1.38.2.
  * CI/release pipelines updated to build and include the SDK distribution before publish.

* **Tests**
  * Added/updated regression tests validating SDK packaging and installer behavior.


trek-e (changes requested):
Adversarial review — CHANGES REQUESTED.

## Critical: PR 2460 conflict

PR 2460 (currently open against main) adds `chmodSync` inside `installSdkIfNeeded()`. This PR removes that entire function. If both target main, one of the following must happen before this merges:

1. PR 2460 is closed as superseded by this PR (the architectural fix eliminates the need for the chmod patch).
2. Or this PR merges first and 2460 is rebased/withdrawn.

The context-monitor fix in PR 2460 (`gsd-statusline.js` bridge `used_pct`) is independent and unaffected.

## Real bug: staleStateVersion can return a shipped milestone

The new `getMilestoneInfo` path reads `milestone:` from STATE.md and returns it immediately without checking whether that version is marked shipped (✅) in ROADMAP.md. If STATE.md has `milestone: v1.0` after v1.0 shipped (stale frontmatter — a common state during milestone transition), `getMilestoneInfo` returns v1.0 instead of falling through to detect the new active milestone.

The downstream `extractCurrentMilestone` and `getMilestonePhaseFilter` both depend on `getMilestoneInfo` returning the correct active version. A stale STATE.md `milestone:` field would cause phase filters to scope to the wrong milestone's phases.

CodeRabbit flagged this (PR comments, around line 1530-1549). The fix is to check the roadmap for a shipped marker before returning:

```js
if (stateVersion) {
  // Don't return a shipped version from STATE.md
  if (/[✅].*/.test(roadmap.split(stateVersion)[0]?.split('\n').pop() || '')) {
    // stateVersion is shipped — fall through to roadmap detection
  } else {
    // ... existing headingMatch / listMatch logic
  }
}
```

A simpler guard: check if the line containing `stateVersion` in ROADMAP.md is prefixed with ✅ or "shipped". Without this, a user who ships v1.0 and starts v2.0 but hasn't updated `STATE.md` frontmatter yet will have phase counting, archive guards, and plan routing all scoped to the wrong milestone.

## test:coverage without sdk/dist in clean environments

CodeRabbit flagged this across three review rounds. `test:coverage` does not depend on `build:sdk`, but the new tests (`tests/bug-2441-sdk-decouple.test.cjs`) assert on `sdk/dist/cli.js` and the installer checks for it. In a clean checkout (no pre-existing `sdk/dist/`), `test:coverage` will fail. CI works because the test.yml workflow runs `npm run build:sdk` before coverage, but local `npm run test:coverage` in a fresh clone will fail.

The fix is to add `"pretest:coverage": "npm run build:sdk"` to package.json scripts. This is a one-line addition and prevents confusing failures for contributors.

## Nitpick: test assertions on shim content are too loose

Tests `bin/gsd-sdk.js forwards to sdk/dist/cli.js` and `bin/gsd-sdk.js uses node to invoke cli.js` match substrings like `'sdk'`, `'dist'`, `'cli.js'`, `'spawnSync'`, `'node'` — these would pass from comments or the shebang line alone. The test adds no real regression protection. Using `assert.match` with the actual `path.resolve(__dirname, '..', 'sdk', 'dist', 'cli.js')` expression and the `spawnSync(process.execPath, [cliPath, ...process.argv.slice(2)])` call pattern would make these tests meaningful.

This is a non-blocking nitpick but worth fixing since the tests are supposed to prevent #2453-class regressions.

## What's good

The architectural direction is correct. Removing the build-from-source dance eliminates the entire class of failures described in #2441 and #2453. The shim approach is sound. CI gating via `install-smoke.yml smoke-unpacked` is the right safety net. The dep-lifting fix (commit 9276a34) correctly identifies that `sdk/node_modules` is not shipped in the tarball.

The `test:coverage` issue and stale STATE.md bug must be fixed before this merges. The conflict with PR 2460 must be resolved (either close 2460's install.js changes or merge this first).

jeremymcs:
Thanks for the adversarial review. Responses per item:

**#1 PR 2460 conflict — acknowledged, coordination plan:**
This PR is the architectural fix that makes 2460's `chmodSync` patch obsolete — 2460 adds a chmod inside `installSdkIfNeeded()`, but this PR deletes that function entirely. Proposing: **merge this PR first**, then 2460 rebases and drops its `bin/install.js` + `tests/bug-2453-sdk-cli-chmod.test.cjs` changes (both obviated), keeping only the independent `gsd-statusline.js` context-monitor fix for #2451. Will coordinate on 2460.

**#2 staleStateVersion / getMilestoneInfo — out of scope:**
This PR does not touch milestone detection, STATE.md parsing, ROADMAP.md parsing, `getMilestoneInfo`, `extractCurrentMilestone`, or `getMilestonePhaseFilter`. Full diff scope:

\`\`\`
.github/workflows/install-smoke.yml
.github/workflows/release.yml
.github/workflows/test.yml
CHANGELOG.md
bin/gsd-sdk.js          (new)
bin/install.js          (162 lines deleted, no milestone code)
package.json
package-lock.json
sdk/package.json
tests/bug-2441-sdk-decouple.test.cjs  (new)
tests/bugs-1656-1657.test.cjs
\`\`\`

Possibly conflated with another open PR. Happy to open a separate issue for the stale-milestone case if it's a real bug elsewhere — but fixing it here would be scope creep on a targeted hotfix.

**#3 test:coverage without sdk/dist — fixed in 6d42b69:**
Added both \`pretest\` and \`pretest:coverage\` hooks running \`npm run build:sdk\`, so a clean clone can \`npm run test\` or \`npm run test:coverage\` without extra steps.

**#4 shim test assertions — tightened in 6d42b69:**
Replaced loose \`.includes('sdk')\` substring checks with strict regex assertions on:
- the exact \`path.resolve(__dirname, "..", "sdk", "dist", "cli.js")\` expression
- \`spawnSync(process.execPath, ...)\`
- \`process.argv.slice(2)\` arg forwarding
- \`process.exit(...)\` child-status propagation

These would now fail if the shim regressed to requiring the exec bit or lost arg/exit-code forwarding.

trek-e (approved):

trek-e:
@jeremymcs — the architecture is sound and the author responses in the thread addressed all prior concerns. The shim approach (bin/gsd-sdk.js delegates to sdk/dist/cli.js via `node`) correctly sidesteps the chmod and PATH fragility at the root cause level rather than patching symptoms.

One coordination note: PR #2536 (fix/2453-sdk-cli-executable-bit) also touches `installSdkIfNeeded()` in `bin/install.js` — it adds a `chmodSync` call after the tsc build. Since this PR deletes that function entirely, #2536's bin/install.js changes are superseded. If this PR lands first, #2536 will need a rebase that drops those install.js hunks (it also has an independent `gsd-statusline.js` fix for #2451 that should be preserved).

This PR needs a rebase to resolve the merge conflict. Once rebased, it should be ready to merge.

trek-e:
Rebased onto current `main` (22 commits ahead of prior tip). Conflicts resolved:

- `package.json` — kept main's `1.38.2` version bump (both paths converged on 1.38.2 via different routes).
- `bin/install.js` — kept the decoupled-install path from this PR; discarded the now-dead build-from-source block from main (CVE-2024-27980 shell fix, `formatSpawnFailure`, `chmod` after `npm install -g` — all for a code path this PR deletes).
- `.github/workflows/release.yml` — kept the `npm run build:sdk` script call from this PR; discarded main's inline `cd sdk && npm ci && npm run build` (same work, expressed via the new npm script).

Added two small commits on top of the original series:

1. `fix(sdk): double-cast WorkflowConfig to Record for strict tsc build` — unblocks `npm run build` in `sdk/`; same fix as [#2622](https://github.com/gsd-build/get-shit-done/pull/2622), drop if that lands first.
2. `test: remove bug-2598 test obsoleted by SDK decoupling` — removes three tests (`bug-2598-windows-npm-spawn-shell`, `bug-2453-sdk-cli-chmod`, `bug-2525-sdk-executable-bit`) that asserted implementation details of the build-from-source path this PR deletes. The underlying concerns (Windows .cmd spawning, chmod of cli.js) are structurally addressed by the shim architecture — the shim uses `child_process.spawn` with an explicit argv (no .cmd wrapper) and a chmod-if-missing check at install time. Worth a follow-up issue to add new tests asserting the shim-path invariants.

Verification: full root test suite **5410 pass / 0 fail**. `cd sdk && npm run build` succeeds. CI will re-run from the forced push.

PR #2458, opened 2026-04-20
fix: zero-padded phase numbers bypass archived-phase guard; stale current_milestone

### LINKED ISSUE #2391: init plan-phase with zero-padded phase number bypasses archived-phase guard (workstream regression)
## GSD Version
1.37.1

## Runtime
Claude Code

## OS / Node / Shell
Windows 11, Node v22, bash (MSYS)

## Installation Method
Local (`.claude/get-shit-done/`)

## Summary

`node gsd-tools.cjs init plan-phase 03 --ws <workstream>` returns an archived milestone phase instead of the current workstream phase. The archived-phase guard added for #2185 is present in `cmdInitPlanPhase` but never activates because `getRoadmapPhaseInternal` fails to match the padded phase number against the current ROADMAP.md, which uses unpadded `### Phase 3:` headings. The zero-padded input `"03"` is spliced verbatim into a regex that requires `Phase 03:` — no match, so `roadmapPhase.found` is false and the archived fallback leaks through.

All `/gsd-*` slash commands canonically pass padded phase numbers (`03`, `07`, etc.), so this reproduces for every workstream user hitting a phase number the archived `v*-phases/` directory also has.

## Reproduction

Directory layout (project on the household workstream where v1.0 was archived):

```
.planning/
  ROADMAP.md                                        # legacy v1.0 (not relevant here)
  phases/                                           # (not used — workstream takes over)
  milestones/
    v1.0-phases/
      03-plant-collection-and-rooms/                # SHIPPED v1.0 phase 3
        03-CONTEXT.md
        03-RESEARCH.md
        03-01-PLAN.md … 03-05-PLAN.md
  workstreams/
    household/
      ROADMAP.md                                    # current — "### Phase 3: Rotation Engine + Availability"
      STATE.md
      phases/
        01-schema-foundation-data-migration/        # done
        02-query-action-layer-update/               # in progress
        # no 03-rotation-engine-availability yet
```

Run:

```
node .claude/get-shit-done/bin/gsd-tools.cjs init plan-phase 03 --ws household
```

returns:

```json
{
  "phase_found": true,
  "phase_dir": ".planning/milestones/v1.0-phases/03-plant-collection-and-rooms",
  "phase_name": "plant-collection-and-rooms",
  "phase_slug": "plant-collection-and-rooms",
  "has_context": true,
  "has_research": true,
  "has_plans": true,
  "plan_count": 6,
  "context_path": ".planning/milestones/v1.0-phases/03-plant-collection-and-rooms/03-CONTEXT.md",
  "research_path": ".planning/milestones/v1.0-phases/03-plant-collection-and-rooms/03-RESEARCH.md",
  "roadmap_path": ".planning/workstreams/household/ROADMAP.md",
  "phase_req_ids": null
}
```

Workaround — passing the unpadded number resolves correctly:

```
node .claude/get-shit-done/bin/gsd-tools.cjs init plan-phase 3 --ws household
```

```json
{
  "phase_found": true,
  "phase_dir": null,
  "phase_name": "Rotation Engine + Availability",
  "phase_slug": "rotation-engine-availability",
  "has_context": false,
  "has_research": false,
  "has_plans": false,
  "phase_req_ids": "ROTA-02, ROTA-03, ROTA-04, ROTA-05, ROTA-06, ROTA-07, AVLB-01, AVLB-02, AVLB-03, AVLB-04, AVLB-05"
}
```

## Expected

`init plan-phase 03 --ws household` should behave identically to `init plan-phase 3 --ws household` — same phase identity, same req IDs, same (clean-slate) paths. Downstream, the `/gsd-plan-phase` workflow should steer gsd-planner at the household Phase 3 context, not shipped v1.0 artifacts.

## Root Cause

`getRoadmapPhaseInternal` in `get-shit-done/bin/lib/core.cjs` (around line 1327) builds the ROADMAP lookup regex by splicing the caller-supplied `phaseNum` verbatim:

```js
const escapedPhase = escapeRegex(phaseNum.toString());
const phasePattern = new RegExp(`#{2,4}\s*Phase\s+${escapedPhase}:\s*([^\n]+)`, 'i');
```

When `phaseNum === "03"` the pattern becomes `/#{2,4}\s*Phase\s+03:/i` and cannot match `### Phase 3: Rotation Engine + Availability`. The regex is only tolerant of padding if the ROADMAP itself uses the padded form — and the canonical ROADMAP format, and the form emitted by `/gsd-add-phase` / `/gsd-new-project`, is unpadded (`Phase 1:`, `Phase 2:`, `Phase 3:` …).

Because the match returns null, the current-milestone guard at `init.cjs:216`:

```js
if (phaseInfo?.archived && roadmapPhase?.found) {
  phaseInfo = null;
}
```

never fires, and the archived v1.0 match from `findPhaseInternal` leaks through. The mismatch between `phase_req_ids: null` (couldn't find the phase in ROADMAP) and `phase_found: true / phase_dir: .planning/milestones/v1.0-phases/...` (found it in archive) is the telltale fingerprint.

This is the same class of bug as #2185 (archived milestone leaks through), exposed by a new path — and closely adjacent to #2195 (padded vs unpadded phase-number handling in stats). #2185's fix assumed the guard predicate was reliable; it is not.

## Impact

For any project that uses workstreams (`--ws <name>`) **and** has at least one archived milestone, every `/gsd-plan-phase N` invocation where `N` is single-digit is silently routed at stale artifacts:

- `gsd-planner` reads the archived phase's CONTEXT.md / RESEARCH.md
- Generated plans live in the wrong phase directory (`.planning/milestones/v*-phases/…` not the workstream's `phases/`)
- `phase_req_ids` drops to null so checker's requirements-coverage gate can't run

Hit in real use on a plant-tracking app: `/gsd-plan-phase 03 --ws household` for the household workstream's Phase 3 (Rotation Engine + Availability) resolved to the shipped v1.0 phase 3 (plant-collection-and-rooms), a totally unrelated subject. Had to manually detect the drift and work around by passing `3` instead of `03`.

## Suggested Fix

Normalize `phaseNum` in `getRoadmapPhaseInternal` before regex construction so the lookup accepts both padded and unpadded forms. Strip a single leading zero when the phase is purely numeric and longer than one digit:

```js
const normalized = /^0\d+$/.test(String(phaseNum)) ? String(phaseNum).replace(/^0+/, '') : String(phaseNum);
const escapedPhase = escapeRegex(normalized);
```

Or, more permissively, build the regex to accept optional leading zeros on the roadmap side:

```js
const escapedPhase = escapeRegex(String(phaseNum).replace(/^0+(?=\d)/, ''));
const phasePattern = new RegExp(`#{2,4}\s*Phase\s+0*${escapedPhase}:\s*([^\n]+)`, 'i');
```

Either fix unblocks padded `/gsd-*` invocations; the second also future-proofs against roadmaps that adopt padded headings.

## Related

- #2185 — archived phase leak (fix-released). Same failure class; this is a path the guard doesn't cover.
- #2195 — duplicate phase rows in stats from padded/unpadded keying.
- #112 — older thread on phase-number padding inconsistency across the codebase.

### LINKED ISSUE #2409: `init` commands return stale `current_milestone` when latest is shipped
## Observed

`gsd-tools init new-milestone` (and other `init` variants via the JSON workflow context) returns a stale `current_milestone` that matches a historical milestone rather than the actual in-progress or shipped latest.

```json
$ gsd-tools init new-milestone
{
  ...
  "current_milestone": "v2.2",
  "current_milestone_name": "— Ruflo Advanced Features)",
  "latest_completed_milestone": "v2.9",
  "latest_completed_milestone_name": "Full-Pass Verification & Bug Fixing",
  ...
}
```

My project's latest is v2.9 (shipped 2026-04-18) — `latest_completed_milestone` correctly reflects this. But `current_milestone` returns `v2.2` which was shipped months ago on 2026-04-03.

## Root cause (my guess)

Looking at `extractCurrentMilestone()` in `~/.claude/get-shit-done/bin/lib/core.cjs`:

1. Reads `milestone:` frontmatter from `STATE.md`. In my case this is `v2.9`.
2. Looks for section matching that version in ROADMAP.md's content.

Then in `getMilestoneInfo()`:
1. Looks for `🚧\s*\*\*v(N\.N)\s+...\*\*` in-progress marker. My ROADMAP.md has `🚧 v2.9 Full-Pass...` but inside a `<summary>🚧 v2.9 ...</summary>` tag without the `**bold**` the regex requires — no match.
2. Falls through to `stripShippedMilestones()` + regex for first `## .*vN.N...` heading.
3. After strip, first matching heading is `## Phase Details (v1.5–v2.1)` or similar — seems to parse v2.2 from one of the later "Phase Details (v2.X — ...)" headings.

Either way, the result is wrong.

## Expected

When the `milestone:` frontmatter in STATE.md is set to the latest shipped milestone (v2.9 in my case), the CLI should recognize it as the current milestone rather than falling through to a regex that matches old phase-details headings.

Alternatively: when the current milestone is marked as shipped (via `✅` in the top milestone list), `current_milestone` should return `null` / empty to signal "between milestones" so `/gsd-new-milestone` can proceed without ambiguity.

## Impact

Any workflow that relies on `current_milestone` gets wrong data. I saw this surface in my existing memory from a previous session (labeled `project_gsd_tools_milestone_drift`): the init tool creates wrong-named branches during autonomous runs because it thinks we're in an old milestone.

## Environment

- GSD: 1.37.0
- OS: macOS 25.3.0 (darwin)
- Project has `milestone: v2.9` in STATE.md frontmatter
- ROADMAP.md top list shows `✅ v2.9 ... shipped 2026-04-18`
- ROADMAP.md phase details section still has `<summary>🚧 v2.9 ... IN PROGRESS</summary>` — inconsistent with the top marker, but this is the accumulated doc state after verifying milestones where phases don't all flip to ✅

## Proposed resolution

Option A — strict: if `milestone:` in STATE matches a `✅ ... shipped` entry in the top milestone list, treat it as completed and return `null` / empty.

Option B — separate field: add `current_active_milestone` (nullable) alongside `current_milestone` so workflows can distinguish "most recent we were in" vs "actively in-progress."

Option C — fix the in-progress regex: accept 🚧 with or without `**bold**` inside `<summary>` tags.

Happy to contribute a PR if you point me at the preferred direction.

trek-e:
## Summary

- **#2391** — `getRoadmapPhaseInternal` now strips leading zeros from purely numeric phase numbers before building the ROADMAP heading regex. `"03"` now correctly matches `### Phase 3:` instead of failing to match and falling through to the archived milestone.
- **#2409** — `getMilestoneInfo` now reads `milestone:` frontmatter from `STATE.md` first and looks up that version in `ROADMAP.md`. This prevents stale heading matches when the active milestone's `🚧` marker is inside a `<summary>` tag without `**bold**` formatting.

## Test plan

- [ ] `tests/core.test.cjs` — 3 new unit tests: zero-padded `"03"` matches Phase 3, double-zero `"007"` matches Phase 7, regression check for unpadded `"3"`
- [ ] `tests/core.test.cjs` — 2 new unit tests: `getMilestoneInfo` reads STATE.md milestone field over ROADMAP regex match
- [ ] `tests/init.test.cjs` — 2 new integration tests: `init plan-phase 03` returns current ROADMAP phase not archived, `"3"` and `"03"` return identical identity
- [ ] Full suite: 4168 tests, 0 failures

Closes #2391
Closes #2409

## Summary by CodeRabbit

* **Bug Fixes**
  * Phase lookup now treats zero-padded and unpadded phase numbers as equivalent.
  * Milestone detection now respects state-configured milestone versions, and avoids selecting already shipped/complete milestones.

* **Tests**
  * Added regression tests covering phase number normalization, milestone selection precedence, shipped-milestone exclusion, and init plan-phase equivalence between padded and unpadded inputs.

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -1331,7 +1331,12 @@ function getRoadmapPhaseInternal(cwd, phaseNum) {
 
   try {
     const content = extractCurrentMilestone(fs.readFileSync(roadmapPath, 'utf-8'), cwd);
-    const escapedPhase = escapeRegex(phaseNum.toString());
+    // Strip leading zeros from purely numeric phase numbers so "03" matches "Phase 3:"
+    // in canonical ROADMAP headings. Non-numeric IDs (e.g. "PROJ-42") are kept as-is.
+    const normalized = /^\d+$/.test(String(phaseNum))
+      ? String(phaseNum).replace(/^0+(?=\d)/, '')
+      : String(phaseNum);
+    const escapedPhase = escapeRegex(normalized);
     // Match both numeric (Phase 1:) and custom (Phase PROJ-42:) headers
     const phasePattern = new RegExp(`#{2,4}\\s*Phase\\s+${escapedPhase}:\\s*([^\\n]+)`, 'i');
     const headerMatch = content.match(phasePattern);
@@ -1505,6 +1510,43 @@ function generateSlugInternal(text) {
 function getMilestoneInfo(cwd) {
   try {
     const roadmap = fs.readFileSync(path.join(planningDir(cwd), 'ROADMAP.md'), 'utf-8');
+
+    // 0. Prefer STATE.md milestone: frontmatter as the authoritative source.
+    // This prevents falling through to a regex that may match an old heading
+    // when the active milestone's 🚧 marker is inside a <summary> tag without
+    // **bold** formatting (bug #2409).
+    let stateVersion = null;
+    if (cwd) {
+      try {
+        const statePath = path.join(planningDir(cwd), 'STATE.md');
+        if (fs.existsSync(statePath)) {
+          const stateRaw = fs.readFileSync(statePath, 'utf-8');
+          const m = stateRaw.match(/^milestone:\s*(.+)/m);
+          if (m) stateVersion = m[1].trim();
+        }
+      } catch { /* intentionally empty */ }
+    }
+
+    if (stateVersion) {
+      // Look up the name for this version in ROADMAP.md
+      const escapedVer = escapeRegex(stateVersion);
+      // Match heading-format: ## Roadmap v2.9: Name  or  ## v2.9 Name
+      const headingMatch = roadmap.match(
+        new RegExp(`##[^\\n]*${escapedVer}[:\\s]+([^\\n(]+)`, 'i')
+      );
+      if (headingMatch) {
+        return { version: stateVersion, name: headingMatch[1].trim() };
+      }
+      // Match list-format: 🚧 **v2.9 Name** or 🚧 v2.9 Name
+      const listMatch = roadmap.match(
+        new RegExp(`🚧\\s*\\*?\\*?${escapedVer}\\s+([^*\\n]+)`, 'i')
+      );
+      if (listMatch) {
+        return { version: stateVersion, name: listMatch[1].trim() };
+      }
+      // Version found in STATE.md but no name match — return with bare version
+      return { version: stateVersion, name: 'milestone' };
+    }
 
     // First: check for list-format roadmaps using 🚧 (in-progress) marker
     // e.g. "- 🚧 **v2.1 Belgium** — Phases 24-28 (in progress)"
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -1535,17 +1535,24 @@ function getMilestoneInfo(cwd) {
         new RegExp(`##[^\\n]*${escapedVer}[:\\s]+([^\\n(]+)`, 'i')
       );
       if (headingMatch) {
-        return { version: stateVersion, name: headingMatch[1].trim() };
+        // If the heading line contains ✅ the milestone is already shipped.
+        // Fall through to normal detection so the NEW active milestone is returned
+        // instead of the stale shipped one still recorded in STATE.md.
+        if (!headingMatch[0].includes('✅')) {
+          return { version: stateVersion, name: headingMatch[1].trim() };
+        }
+        // Shipped milestone — do not early-return; fall through to normal detection below.
+      } else {
+        // Match list-format: 🚧 **v2.9 Name** or 🚧 v2.9 Name
+        const listMatch = roadmap.match(
+          new RegExp(`🚧\\s*\\*?\\*?${escapedVer}\\s+([^*\\n]+)`, 'i')
+        );
+        if (listMatch) {
+          return { version: stateVersion, name: listMatch[1].trim() };
+        }
+        // Version found in STATE.md but no name match in ROADMAP — return bare version
+        return { version: stateVersion, name: 'milestone' };
       }
-      // Match list-format: 🚧 **v2.9 Name** or 🚧 v2.9 Name
-      const listMatch = roadmap.match(
-        new RegExp(`🚧\\s*\\*?\\*?${escapedVer}\\s+([^*\\n]+)`, 'i')
-      );
-      if (listMatch) {
-        return { version: stateVersion, name: listMatch[1].trim() };
-      }
-      // Version found in STATE.md but no name match — return with bare version
-      return { version: stateVersion, name: 'milestone' };
     }
 
     // First: check for list-format roadmaps using 🚧 (in-progress) marker
@@ -1559,11 +1566,14 @@ function getMilestoneInfo(cwd) {
       };
     }
 
-    // Second: heading-format roadmaps — strip shipped milestones in <details> blocks
+    // Second: heading-format roadmaps — strip shipped milestones.
+    // <details> blocks are stripped by stripShippedMilestones; heading-format ✅ markers
+    // are excluded by the negative lookahead below so a stale STATE.md version (or any
+    // shipped ✅ heading) never wins over the first non-shipped milestone heading.
     const cleaned = stripShippedMilestones(roadmap);
-    // Extract version and name from the same ## heading for consistency
+    // Negative lookahead skips headings that contain ✅ (shipped milestone marker).
     // Supports 2+ segment versions: v1.2, v1.2.1, v2.0.1, etc.
-    const headingMatch = cleaned.match(/## .*v(\d+(?:\.\d+)+)[:\s]+([^\n(]+)/);
+    const headingMatch = cleaned.match(/## (?!.*✅).*v(\d+(?:\.\d+)+)[:\s]+([^\n(]+)/);
     if (headingMatch) {
       return {
         version: 'v' + headingMatch[1],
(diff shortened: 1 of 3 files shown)


trek-e:
Adversarial review — CHANGES REQUESTED (self-authored: findings posted as comment).

## Real bug: stateVersion is returned for shipped milestones

The new `getMilestoneInfo` code reads `milestone:` from STATE.md and returns that version without checking if ROADMAP.md marks it as shipped. Scenario:

1. User ships v1.0. ROADMAP updates (✅ v1.0 marker or `<details>` wrapping).
2. New milestone v2.0 starts. STATE.md still has `milestone: v1.0` (stale — happens during transition before `init new-milestone` is run).
3. `getMilestoneInfo` reads `milestone: v1.0` from STATE.md and returns it immediately.
4. `extractCurrentMilestone`, `getMilestonePhaseFilter`, `getRoadmapPhaseInternal` all scope to v1.0's phases — wrong milestone, stale context.

CodeRabbit flagged this (inline comment around lines 1530-1549 in core.cjs). The fix is to check whether the version found in STATE.md is marked shipped before returning early. If it is shipped, fall through to the existing ROADMAP-based detection.

Without this guard, the fix for #2409 creates a new failure mode: users whose STATE.md lags behind milestone transition get stale phase scoping. The stale-STATE.md case is real — it's the window between `init new-milestone` and the previous milestone being archived.

## What's correct

**Zero-padded phase numbers (#2391):** The `replace(/^0+(?=\d)/, '')` on purely numeric inputs is correct. The `(?=\d)` lookahead prevents stripping zeros from `"007"` down to `""` — `"007".replace(/^0+(?=\d)/, '')` gives `"7"`. Non-numeric IDs pass through untouched.

**Tests for #2391:** Three tests cover `"03"` → Phase 3, `"007"` → Phase 7, and unpadded `"3"` as regression check. The integration test in `init.test.cjs` correctly verifies that `"03"` returns current ROADMAP phase (not archived v1.0), and that `"3"` and `"03"` return identical identity.

The `stateVersion` shipped-milestone guard must be added before this merges. Suggest adding a helper check like:

```js
// Do not return a shipped milestone from STATE.md stale frontmatter.
// Check if this version appears on a line with a ✅ marker.
const isShipped = roadmap.split('\n').some(line => 
  line.includes(stateVersion) && /✅|shipped/i.test(line)
);
if (isShipped) {
  // fall through to roadmap-based detection
} else {
  // ... existing headingMatch/listMatch return logic
}
```

trek-e:
The CodeRabbit finding on `stateVersion` returning a shipped milestone (core.cjs ~line 1530-1549) is already addressed in the current code. The `headingMatch` branch already contains an explicit ✅ guard:

```js
if (!headingMatch[0].includes('✅')) {
  return { version: stateVersion, name: headingMatch[1].trim() };
}
// Shipped milestone — do not early-return; fall through to normal detection below.
```

If `stateVersion` appears on a heading line that contains ✅, the early-return is skipped and detection falls through to the ROADMAP-based path. The `listMatch` branch also only matches 🚧-prefixed lines, so a shipped version not preceded by 🚧 will not match there either.

The finding is accurate in concept (a stale STATE.md milestone could be returned) but the fix is already in the branch. No further change needed here.

PR #2460, opened 2026-04-20
fix(install): chmod sdk dist/cli.js executable; fix context monitor over-reporting

### LINKED ISSUE #2451: Context monitor hook reports higher usage than actual (78% vs 65% real)

### LINKED ISSUE #2453: installSdkIfNeeded: tsc-built dist/cli.js lands with mode 644 on fresh installs, leaving gsd-sdk non-executable

trek-e:
## Summary

- **#2453**: `installSdkIfNeeded()` now explicitly `chmodSync`s `sdk/dist/cli.js` to `0o755` after `npm install -g .`. TypeScript emits files at process umask (typically `0o644`), and `npm install -g` from a local directory does not chmod bin-script targets the way tarball extraction does. Without this, the `gsd-sdk` symlink target is non-executable on every first install, causing `command -v gsd-sdk` to fail and misdirecting users to check their PATH.
- **#2451**: `gsd-statusline.js` now writes `used_pct` to the bridge file as the raw CC value (`Math.round(100 - remaining_percentage)`) rather than the buffer-normalized value. The normalization (which accounts for the 16.5% autocompact reserve) was inflating context monitor warning messages by ~13 percentage points compared to CC's native `/context` command. The statusline progress bar continues to display the normalized value; only the bridge field fed to `gsd-context-monitor.js` changes.

## Test plan

- [ ] `node --test tests/bug-2453-sdk-cli-chmod.test.cjs` — 4 pass
- [ ] `node --test tests/bug-2451-context-monitor-over-report.test.cjs` — 7 pass
- [ ] `node --test tests/gsd-statusline.test.cjs` — 27 pass (including updated #2219 tests)
- [ ] `npm run test:coverage` — 4169 pass, 0 fail

Closes #2453
Closes #2451

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -6786,6 +6786,26 @@ function installSdkIfNeeded() {
     emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
   }
 
+  // 3a. Explicitly chmod dist/cli.js to 0o755 in the global install location.
+  // `tsc` emits files at process umask (typically 0o644 — non-executable), and
+  // `npm install -g` from a local directory does NOT chmod bin-script targets the
+  // way tarball extraction does. Without this, the `gsd-sdk` bin symlink points at
+  // a non-executable file and `command -v gsd-sdk` fails on every first install
+  // (root cause of #2453). Mirrors the pattern used for hook files in this installer.
+  try {
+    const prefixRes = spawnSync(npmCmd, ['config', 'get', 'prefix'], { encoding: 'utf-8' });
+    if (prefixRes.status === 0) {
+      const npmPrefix = (prefixRes.stdout || '').trim();
+      const sdkPkg = JSON.parse(fs.readFileSync(path.join(sdkDir, 'package.json'), 'utf-8'));
+      const sdkName = sdkPkg.name; // '@gsd-build/sdk'
+      const globalModulesDir = process.platform === 'win32'
+        ? path.join(npmPrefix, 'node_modules')
+        : path.join(npmPrefix, 'lib', 'node_modules');
+      const cliPath = path.join(globalModulesDir, sdkName, 'dist', 'cli.js');
+      try { fs.chmodSync(cliPath, 0o755); } catch (e) { /* Windows / path not found */ }
+    }
+  } catch (e) { /* Non-fatal: PATH verification in step 4 will catch any real failure */ }
+
   // 4. Verify gsd-sdk is actually resolvable on PATH. npm's global bin dir is
   //    not always on the current shell's PATH (Homebrew prefixes, nvm setups,
   //    unconfigured npm prefix), so a zero exit status from `npm install -g`
--- a/hooks/gsd-statusline.js
+++ b/hooks/gsd-statusline.js
@@ -147,10 +147,15 @@ function runStatusline() {
       if (sessionSafe) {
         try {
           const bridgePath = path.join(os.tmpdir(), `claude-ctx-${session}.json`);
+          // used_pct written to the bridge must match CC's native /context reporting:
+          // raw used = 100 - remaining_percentage (no buffer normalization applied).
+          // The normalized `used` value is correct for the statusline progress bar but
+          // inflates the context monitor warning messages by ~13 points (#2451).
+          const rawUsedPct = Math.round(100 - remaining);
           const bridgeData = JSON.stringify({
             session_id: session,
             remaining_percentage: remaining,
-            used_pct: used,
+            used_pct: rawUsedPct,
             timestamp: Math.floor(Date.now() / 1000)
           });
           fs.writeFileSync(bridgePath, bridgeData);
--- a/tests/bug-2451-context-monitor-over-report.test.cjs
+++ b/tests/bug-2451-context-monitor-over-report.test.cjs
@@ -1,0 +1,185 @@
+/**
+ * Regression test for bug #2451
+ *
+ * The GSD context monitor hook over-reports usage by ~13 percentage points
+ * compared to Claude Code's native /context command. The root cause:
+ *
+ * gsd-statusline.js writes two values to the bridge file:
+ *   - remaining_percentage: raw remaining from CC (e.g. 35%)
+ *   - used_pct: normalized "usable" percentage (e.g. 78%) — accounts for
+ *     the 16.5% autocompact buffer by scaling: (100 - remaining - buffer) /
+ *     (100 - buffer) * 100
+ *
+ * gsd-context-monitor.js displays used_pct (78%) in warning messages.
+ * But CC's native /context shows raw used = 100 - remaining = 65%.
+ * The 13-point gap is exactly the buffer normalization overhead.
+ *
+ * Fix: the bridge must write used_pct as the raw value (Math.round(100 -
+ * remaining)), not the buffer-normalized value. The statusline progress bar
+ * continues to use the normalized value for its own display; only the bridge
+ * value that feeds the context monitor needs to be raw/CC-consistent.
+ */
+
+'use strict';
+
+const { test, describe } = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const { execFileSync } = require('node:child_process');
+
+const HOOK_PATH = path.join(__dirname, '..', 'hooks', 'gsd-statusline.js');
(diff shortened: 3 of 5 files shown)


trek-e:
Adversarial review — CONDITIONAL APPROVAL (self-authored: findings posted as comment).

## Conflict with PR 2457

PR 2457 (fix/2441-sdk-decouple) removes `installSdkIfNeeded()` entirely from `bin/install.js`. This PR adds `chmodSync` inside that function. Both target `main`. If 2457 merges first, this PR's `bin/install.js` changes conflict and the chmod block becomes dead code in a function that no longer exists.

**Resolution path:** The context monitor fix (`gsd-statusline.js` + its tests) in this PR is independent and valid regardless. The chmod fix is superseded by 2457's architectural approach (shim invokes cli.js via `node`, so the exec bit is irrelevant). Before merging:

- If 2457 merges first: this PR needs a rebase with the `bin/install.js` hunk dropped (keep only the `gsd-statusline.js` changes and the two test files).
- If this PR merges first: 2457's install.js rewrite will cleanly remove the chmodSync block along with the whole function.

## chmod fix itself (bin/install.js)

The fix is correct in isolation: after `npm install -g .`, resolve the prefix, read `sdk/package.json` for the package name, construct the global modules path platform-specifically, then `chmodSync(cliPath, 0o755)`. Error swallowed silently with a comment. The Windows branch (`node_modules` directly under prefix) and POSIX branch (`lib/node_modules`) are correct.

One subtle gap: the path uses `sdkName` from `sdk/package.json` (the in-repo copy), not from the globally-installed copy. These should be the same, but if a globally-installed version has a different package name than the in-repo source, the path would be wrong. This is an extremely unlikely edge case and the silent catch handles it gracefully.

## Context monitor fix (gsd-statusline.js)

The fix is correct. `used_pct` in the bridge file was `used` (buffer-normalized), now it's `rawUsedPct = Math.round(100 - remaining)`. The `used` variable (normalized, displayed in the statusline bar) is unchanged. The bridge field is now consistent with CC's `/context` native reporting.

The separation of concerns is correct: the statusline UI shows normalized values (accounting for autocompact buffer), while the bridge communicates raw values so `gsd-context-monitor.js` can make threshold comparisons consistent with what the user sees in CC's own `/context`.

Tests in `bug-2451-context-monitor-over-report.test.cjs` are thorough: bridge `used_pct` for three different remaining values, monitor message text at WARNING and CRITICAL thresholds, and gap-from-raw-CC assertion.

Updated `gsd-statusline.test.cjs` correctly shifts the `#2219` tests from asserting on `bridge.used_pct` (now raw) to asserting on the normalized display value parsed from ANSI stdout. The test for bridge `used_pct` raw value is added as a new assertion. This is the right way to split the concerns.

## Summary

The context monitor fix is clean and should merge. The chmod fix is superseded by PR 2457. Recommend rebasing this PR to drop the `bin/install.js` hunk after 2457 merges, or merging this PR first and letting 2457 clean up install.js entirely.

PR #2463, opened 2026-04-20
test: add --no-sdk to copilot-install E2E runners + static guard

### LINKED ISSUE #2461: test: copilot-install E2E calls missing --no-sdk cause test.yml CI failures
## Bug

`tests/copilot-install.test.cjs` has E2E sections (lines 1116–1136 and 1376–1396) that explicitly `delete env.GSD_TEST_MODE` to run the real installer. None of these subprocess calls pass `--no-sdk`, so they trigger `installSdkIfNeeded()`.

Since `af66cd8` made `installSdkIfNeeded()` fatal on failure, these tests crash in `test.yml` CI where:
- the npm global bin directory is not on PATH
- `resolveGsdSdk()` returns null → `emitSdkFatal()` → exit 1

## Root cause

Commit `e213ce0` added `--no-sdk` to 3 hook-deployment tests but missed the 4 installer runner functions in `copilot-install.test.cjs`:

- `runCopilotInstall` — `--copilot --local`
- `runCopilotUninstall` — `--copilot --local --uninstall`
- `runClaudeInstall` — `--claude --local`
- `runClaudeUninstall` — `--claude --local --uninstall`

These tests cover **copilot skill deployment and user-file preservation** — they have no interest in SDK install. The SDK install E2E path is already covered by `install-smoke.yml`.

## Fix

Add `--no-sdk` to all 4 runner functions + add a static guard test (`tests/sdk-no-sdk-guard.test.cjs`) that permanently catches this class of regression going forward.

trek-e:
## Summary

- Adds `--no-sdk` to 4 installer runner functions in `tests/copilot-install.test.cjs` that were missing it, causing fatal `installSdkIfNeeded()` failures in `test.yml` CI
- Adds `tests/sdk-no-sdk-guard.test.cjs`: static analysis guard that permanently catches any future installer subprocess call missing `--no-sdk`

## Root cause

`e213ce0` fixed 3 hook-deployment tests but missed the E2E section of `copilot-install.test.cjs`, which explicitly deletes `GSD_TEST_MODE` and runs the real installer. Since `af66cd8` made `installSdkIfNeeded()` fatal, these calls crashed in CI where npm global bin isn't on PATH → `resolveGsdSdk()` returns null → `emitSdkFatal()` → exit 1.

Affected functions: `runCopilotInstall`, `runCopilotUninstall`, `runClaudeInstall`, `runClaudeUninstall`.

## Test plan

- [x] `node --test tests/sdk-no-sdk-guard.test.cjs` — fails before fix, passes after
- [x] `node --test tests/copilot-install.test.cjs` — all 111 tests pass
- [x] SDK install E2E path is unchanged; dedicated coverage remains in `install-smoke.yml`

Closes #2461

### DIFF
--- a/tests/copilot-install.test.cjs
+++ b/tests/copilot-install.test.cjs
@@ -1116,7 +1116,7 @@ const EXPECTED_AGENTS = fs.readdirSync(path.join(__dirname, '..', 'agents'))
 function runCopilotInstall(cwd) {
   const env = { ...process.env };
   delete env.GSD_TEST_MODE;
-  return execFileSync(process.execPath, [INSTALL_PATH, '--copilot', '--local'], {
+  return execFileSync(process.execPath, [INSTALL_PATH, '--copilot', '--local', '--no-sdk'], {
     cwd,
     encoding: 'utf-8',
     stdio: ['pipe', 'pipe', 'pipe'],
@@ -1127,7 +1127,7 @@ function runCopilotInstall(cwd) {
 function runCopilotUninstall(cwd) {
   const env = { ...process.env };
   delete env.GSD_TEST_MODE;
-  return execFileSync(process.execPath, [INSTALL_PATH, '--copilot', '--local', '--uninstall'], {
+  return execFileSync(process.execPath, [INSTALL_PATH, '--copilot', '--local', '--uninstall', '--no-sdk'], {
     cwd,
     encoding: 'utf-8',
     stdio: ['pipe', 'pipe', 'pipe'],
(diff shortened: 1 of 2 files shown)


trek-e:

**#2461 — Add --no-sdk to copilot-install E2E runners + static guard**

All four functions (`runCopilotInstall`, `runCopilotUninstall`, `runClaudeInstall`, `runClaudeUninstall`) now include `--no-sdk`. The static guard is well-designed: regex built at runtime (avoids tripping itself), scoped only to files that delete `GSD_TEST_MODE`, correctly self-excluding since the guard file does not delete `GSD_TEST_MODE`, and the `INSTALL`/`install.js` args filter correctly targets installer invocations. Completes the partial fix from `e213ce0`. No issues.

PR #2466, opened 2026-04-20
fix(phase): guard backlog dirs and YYYY-MM dates in integer phase removal

### LINKED ISSUE #2434: bug(phase): renameIntegerPhases missing 999.x backlog guard — integer phase remove renames backlog directory
## Summary

When `phase remove <N>` is called with an **integer** phase number and the
project contains a backlog phase (`999.x`), the backlog directory is
incorrectly renumbered to `998.x`.

Example: removing phase 4 in a project that has `999.1-…` renames the backlog
directory to `998.1-…`, leaving it out of sync with the still-correct
`999.1` heading in ROADMAP.md.

The **decimal** path is unaffected — `renameDecimalPhases` is anchored to
`baseInt` and leaves `999.x` alone. This is likely why the bug has gone
unnoticed: most removals in active projects are decimal.

## Expected behavior

`999.x` backlog directories are parked outside the active phase sequence and
must remain untouched by any renumbering operation, consistent with the
existing treatment in `phase-add`, `phase-complete`, `init manager`, etc.

## Actual behavior

`renameIntegerPhases` in `get-shit-done/bin/lib/phase.cjs` (lines ~620–653)
scans all phase directories and filters only on `dirInt > removedInt`:

```js
.map(dir => {
  const m = dir.match(/^(\d+)([A-Z])?(?:\.(\d+))?-(.+)$/i);
  if (!m) return null;
  const dirInt = parseInt(m[1], 10);
  return dirInt > removedInt ? { dir, oldInt: dirInt, ... } : null;
})
```

Because `999 > removedInt` for any small `removedInt`, the backlog directory
is captured and decremented to `998`.

The same module already carries backlog guards at four other locations
(integer check `num >= 999` at lines 352, 366, 430, 440; regex check
`/^999(?:\.|$)/` at line 917). The guard was simply missed in
`renameIntegerPhases`.

## Reproduction

1. Project with an active integer phase 4 and a backlog phase
   `999.1-something`.
2. Run `node bin/gsd-tools.cjs phase remove 4`.
3. Observe: `.planning/phases/999.1-something/` is renamed to
   `.planning/phases/998.1-something/`, while `ROADMAP.md` still says
   `### Phase 999.1: …`.
4. Confirmed NOT triggered by decimal remove (e.g. `phase remove 4.1` uses
   `renameDecimalPhases`, which is anchored to `baseInt` and leaves `999.x`
   untouched).

## Proposed fix

Add the same guard already used elsewhere in the file:

```js
return (dirInt > removedInt && dirInt < 999) ? { ... } : null;
```

Happy to open a PR.

## Environment

- GSD version on main: 1.37.1 (phase.cjs SHA1 `20833faad2c6c2d035a349aa006783838571c31b`)
- Confirmed present as far back as v1.30.0 (`renameIntegerPhases` at line 509,
  no `< 999` guard) — bug has been latent for weeks.
- Node 20, macOS 25.4.0

## Related

- #1950 (fix(phase): skip 999.x backlog phases in phase-add numbering)
- #2129 / PR #2135 (exclude 999.x from phase-complete + init-manager)

Both previous fixes established the `num >= 999` pattern that is missing
in `renameIntegerPhases`.

### LINKED ISSUE #2435: bug(phase): updateRoadmapAfterPhaseRemoval regex corrupts YYYY-MM dates in ROADMAP.md on integer phase remove
## Summary

When `phase remove <N>` is called with an **integer** phase number, the
renumbering loop in `updateRoadmapAfterPhaseRemoval` destructively rewrites
`YYYY-MM-DD` date strings in `ROADMAP.md`.

Concrete symptom observed on a real project: removing phase 4 in a roadmap
containing `2026-04-14` flipped every such date to `2004-04-14` (delta of
22 years). The bug is fully deterministic.

Related to but separate from #2434 (same command, different function,
different root cause).

## Root cause

`get-shit-done/bin/lib/phase.cjs` in `updateRoadmapAfterPhaseRemoval`,
around line 676:

```js
for (let oldNum = MAX_PHASE; oldNum > removedInt; oldNum--) {
  const newNum = oldNum - 1;
  const oldPad = String(oldNum).padStart(2, '0');
  const newPad = String(newNum).padStart(2, '0');
  // ...
  content = content.replace(
    new RegExp(`${oldPad}-(\\d{2})`, 'g'),     // <-- too broad
    `${newPad}-$1`
  );
  // ...
}
```

The regex `NN-\d{2}` is intended to renumber plan IDs like `05-01` → `04-01`,
but it also matches the `YY-MM` fragment of any ISO date `20YY-MM-DD`.

Because the loop iterates `oldNum` from 99 down to `removedInt + 1`, a date
whose two-digit year falls inside that range is decremented **once per
iteration** after it first matches. For `removedInt = 4`:

- `oldNum = 26`: `2026-04-14` matches `26-04` → `2025-04-14`
- `oldNum = 25`: `2025-04-14` matches `25-04` → `2024-04-14`
- … (22 total iterations) …
- `oldNum = 5`:  `2005-04-14` → `2004-04-14`

Net: `2026 → 2004` (delta = `YY − removedInt`).

## Expected behavior

`phase remove` must renumber phase/plan IDs but must not touch date strings.

## Reproduction

Minimal Node repro against any ROADMAP.md containing a `2026-04-14`-style
date:

```js
const fs = require('fs');
let content = fs.readFileSync('ROADMAP.md', 'utf-8');
const removedInt = 4;
for (let oldNum = 99; oldNum > removedInt; oldNum--) {
  const oldPad = String(oldNum).padStart(2, '0');
  const newPad = String(oldNum - 1).padStart(2, '0');
  content = content.replace(new RegExp(`${oldPad}-(\\d{2})`, 'g'), `${newPad}-$1`);
}
// Every 2026-MM-DD is now 2004-MM-DD
```

End-to-end: run `node bin/gsd-tools.cjs phase remove 4` on a project whose
ROADMAP.md has any `2026-MM-DD` line (e.g. `completed 2026-04-14`) — the
committed ROADMAP.md shows `2004-04-14` afterwards.

## Why this has likely gone unnoticed

Three conditions must coincide:

1. `phase remove` is invoked (relatively rare — deferred/backlog workflows
   are more common).
2. Integer phase (decimal removals go through `renameDecimalPhases`, which
   is anchored to `baseInt` and is unaffected).
3. `ROADMAP.md` contains date strings whose two-digit year is **greater than
   `removedInt`**. For a 2026-era project, removing phases 1–25 triggers;
   removing phase ≥ 26 does not (the loop condition `oldNum > removedInt`
   skips `oldNum = 26`, leaving year `26` untouched). The `phase complete`
   workflow writes such dates automatically when a phase is shipped, so
   mature projects with completed phases are at highest risk.

## Proposed fix

Two options, maintainer's call:

**Option A — narrow the regex.** Require the plan-ID context (trailing dash
+ ascii letter, or `.md`, or end-of-token):

```js
content = content.replace(
  new RegExp(`\\b${oldPad}-(\\d{2})(?=[-.:\\s]|$)(?!\\d)`, 'g'),
  `${newPad}-$1`
);
```

(exact anchor pattern depends on the full set of valid plan-ID contexts in
ROADMAP.md — a review of the tests around plan-ID rewriting is warranted)

**Option B — drive renumbering from the known plan-ID set**, not from a
generic regex. Collect plan IDs by scanning the filesystem/headings first,
then do a list of literal replacements. Eliminates the whole class of
"regex matches something it shouldn't" issues.

Happy to open a PR for either once a direction is chosen.

## Environment

- GSD version on main: 1.37.1 (phase.cjs SHA1 `20833faad2c6c2d035a349aa006783838571c31b`)
- Confirmed present in v1.30.0 at the equivalent location (no change to
  the offending regex since then).
- Node 20, macOS 25.4.0

## Related

- #2434 (sibling bug in the same `phase remove` workflow — backlog
  999.x directory gets renumbered). Both are symptoms of `phase remove`
  integer path being under-tested.

trek-e:
Closes #2435
Closes #2434

## Root cause

Two bugs in `renameIntegerPhases` and `updateRoadmapAfterPhaseRemoval` in `bin/lib/phase.cjs`:

**#2434 – 999.x backlog directories get decremented:** `renameIntegerPhases` collected every directory with an integer prefix greater than the removed phase number, including `999.x` backlog dirs. On removal of e.g. phase 4, the `999` directory was included and renamed to `998`.

**#2435 – YYYY-MM-DD dates corrupted in ROADMAP.md:** The regex used to rewrite phase number references in the roadmap (`oldPad-(\\d{2})`) matched date substrings like `04-14` inside `2026-04-14`, corrupting dates when e.g. phase 4 was removed and `04` appeared in a date.

## Fix

- **#2434:** Added `&& dirInt < 999` guard in `renameIntegerPhases` so backlog directories (`999.x`) are never included in the rename sweep.
- **#2435:** Changed the roadmap rewrite regex from `` `${oldPad}-(\\d{2})` `` to `` `(?<!\\d)${oldPad}-(\\d{2})(?!\\d)` `` — negative lookbehind/ahead prevent matching inside longer digit sequences like dates.

## Tests

- `bug-2434: integer phase remove does not rename 999.x backlog directory`
- `bug-2435: integer phase remove does not corrupt YYYY-MM-DD dates in ROADMAP.md`

Full suite: 419 pass, 0 fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Backlog phase directories in the 999.x range are preserved and no longer get renumbered.
  * Date strings in roadmap files are no longer altered during phase updates.
  * Numeric matching tightened to avoid unintended text replacements when renumbering.

* **Tests**
  * Added regression tests covering phase removal edge cases to ensure the above behaviors.

### DIFF
--- a/get-shit-done/bin/lib/phase.cjs
+++ b/get-shit-done/bin/lib/phase.cjs
@@ -673,7 +673,7 @@ function updateRoadmapAfterPhaseRemoval(roadmapPath, targetPhase, isDecimal, rem
         const oldPad = oldStr.padStart(2, '0'), newPad = newStr.padStart(2, '0');
         content = content.replace(new RegExp(`(#{2,4}\\s*Phase\\s+)${oldStr}(\\s*:)`, 'gi'), `$1${newStr}$2`);
         content = content.replace(new RegExp(`(Phase\\s+)${oldStr}([:\\s])`, 'g'), `$1${newStr}$2`);
-        content = content.replace(new RegExp(`(?<!\\d)${oldPad}-(\\d{2})(?!\\d)`, 'g'), `${newPad}-$1`);
+        content = content.replace(new RegExp(`(?<![0-9-])${oldPad}-(\\d{2})(?![0-9-])`, 'g'), `${newPad}-$1`);
         content = content.replace(new RegExp(`(\\|\\s*)${oldStr}\\.\\s`, 'g'), `$1${newStr}. `);
         content = content.replace(new RegExp(`(Depends on:\\*\\*\\s*Phase\\s+)${oldStr}\\b`, 'gi'), `$1${newStr}`);
       }
--- a/get-shit-done/bin/lib/phase.cjs
+++ b/get-shit-done/bin/lib/phase.cjs
@@ -625,7 +625,7 @@ function renameIntegerPhases(phasesDir, removedInt) {
       const m = dir.match(/^(\d+)([A-Z])?(?:\.(\d+))?-(.+)$/i);
       if (!m) return null;
       const dirInt = parseInt(m[1], 10);
-      return dirInt > removedInt ? { dir, oldInt: dirInt, letter: m[2] ? m[2].toUpperCase() : '', decimal: m[3] ? parseInt(m[3], 10) : null, slug: m[4] } : null;
+      return (dirInt > removedInt && dirInt < 999) ? { dir, oldInt: dirInt, letter: m[2] ? m[2].toUpperCase() : '', decimal: m[3] ? parseInt(m[3], 10) : null, slug: m[4] } : null;
     })
     .filter(Boolean)
     .sort((a, b) => a.oldInt !== b.oldInt ? b.oldInt - a.oldInt : (b.decimal || 0) - (a.decimal || 0));
@@ -673,7 +673,7 @@ function updateRoadmapAfterPhaseRemoval(roadmapPath, targetPhase, isDecimal, rem
         const oldPad = oldStr.padStart(2, '0'), newPad = newStr.padStart(2, '0');
         content = content.replace(new RegExp(`(#{2,4}\\s*Phase\\s+)${oldStr}(\\s*:)`, 'gi'), `$1${newStr}$2`);
         content = content.replace(new RegExp(`(Phase\\s+)${oldStr}([:\\s])`, 'g'), `$1${newStr}$2`);
-        content = content.replace(new RegExp(`${oldPad}-(\\d{2})`, 'g'), `${newPad}-$1`);
+        content = content.replace(new RegExp(`(?<!\\d)${oldPad}-(\\d{2})(?!\\d)`, 'g'), `${newPad}-$1`);
         content = content.replace(new RegExp(`(\\|\\s*)${oldStr}\\.\\s`, 'g'), `$1${newStr}. `);
         content = content.replace(new RegExp(`(Depends on:\\*\\*\\s*Phase\\s+)${oldStr}\\b`, 'gi'), `$1${newStr}`);
       }
--- a/tests/phase.test.cjs
+++ b/tests/phase.test.cjs
@@ -1357,6 +1357,53 @@ describe('phase remove command', () => {
     // Phase 5 should be renumbered to 4
     assert.ok(roadmap.includes('Phase 4: Final'), 'Phase 5 should be renumbered to Phase 4');
   });
+
+  test('bug-2435: integer phase remove does not corrupt date whose month matches removed phase number', () => {
+    // Setup: removing phase 4 from a roadmap containing 2026-05-14
+    // When renumbering phase 5→4, the regex must not replace "05-14" in the date "2026-05-14"
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'ROADMAP.md'),
+      `# Roadmap
+
+### Phase 1: Foundation
+**Goal:** Setup
+**Completed:** 2026-01-15
+
+### Phase 2: Auth
+**Goal:** Authentication
+**Completed:** 2026-02-20
+
+### Phase 3: Features
+**Goal:** Core features
+**Completed:** 2026-03-10
+
+### Phase 4: Extras
+**Goal:** Extra stuff
+
+### Phase 5: Final
+**Goal:** Final phase
+**Due:** 2026-05-14
+`
+    );
+
+    fs.mkdirSync(path.join(tmpDir, '.planning', 'phases', '01-foundation'), { recursive: true });
+    fs.mkdirSync(path.join(tmpDir, '.planning', 'phases', '02-auth'), { recursive: true });
+    fs.mkdirSync(path.join(tmpDir, '.planning', 'phases', '03-features'), { recursive: true });
+    fs.mkdirSync(path.join(tmpDir, '.planning', 'phases', '04-extras'), { recursive: true });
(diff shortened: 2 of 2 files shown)


trek-e:
**Adversarial review — CHANGES REQUESTED (cannot self-review via formal mechanism)**

**#2434 — 999.x backlog directory guard**

The `dirInt < 999` guard is correct and consistent with the existing `>= 999` skip pattern throughout `phase.cjs`. No issues.

**#2435 — YYYY-MM-DD date corruption — fix is INCOMPLETE**

The regex change from:
```js
new RegExp(`${oldPad}-(\\d{2})`, 'g')
```
to:
```js
new RegExp(`(?<!\\d)${oldPad}-(\\d{2})(?!\\d)`, 'g')
```

does NOT protect YYYY-MM-DD dates in the general case. The lookbehind `(?<!\d)` only prevents a match when the digit immediately before `oldPad` is another digit. In a date like `2026-05-14`, the character before `05` is `-` (a non-digit), so the lookbehind does not prevent the match. When removing phase 4, the loop renames phase 5 → 4 using `oldPad='05'`, and the regex `/(?<!\d)05-(\d{2})(?!\d)/` **does match** `05-14` inside `2026-05-14`, corrupting it to `2026-04-14`.

Verified:
```js
const re = new RegExp('(?<!\\d)05-(\\d{2})(?!\\d)', 'g');
'**Completed:** 2026-05-14'.replace(re, '04-$1');
// => '**Completed:** 2026-04-14'  ← corrupted
```

**The test passes by accident.** The test uses `2026-04-14` when removing phase 4. The loop for that case checks `oldPad='05'` (renaming phase 5 → 4), and `2026-04-14` does not contain `05-14`, so the test passes but does not actually validate protection against the real corruption scenario.

**Required fix:** The roadmap rewrite regex must be scoped to match only phase directory prefixes in a path context, not bare `NN-XX` substrings inside prose. Options:

1. Match only at start-of-line or after `phases/` path separator:
   ```js
   new RegExp(`(^|phases/)${oldPad}-(\\d{2})`, 'gm')
   ```
2. Use a full-date negative lookbehind that accounts for the `YYYY-` prefix:
   ```js
   new RegExp(`(?<!\\d{4}-)${oldPad}-(\\d{2})(?!\\d)`, 'g')
   ```
3. Scope replacements to lines that start with `###`, `##`, or contain `Phase`:
   Only rewrite lines where a phase number is semantically present.

The test must also be updated to use a date that matches the pattern being replaced (e.g., use `2026-05-14` when removing phase 4, so the test actually covers the `05-14` replacement).

PR #2467, opened 2026-04-20
fix(bugs): plan-phase branch rename, worktree teardown errors, Makefile test priority, OpenCode Windows $HOME

### LINKED ISSUE #2376: @$HOME无法被正确映射
### GSD Version

1.36.0

### Runtime

OpenCode

### Operating System

Windows

### Node.js Version

v24.12.0

### Shell

PowerShell7

### Installation Method

### What happened?

在opencode中使用任何gsd命令时@$HOME都无法被正确映射为用户目录
<auto-slash-command>
# /gsd-stats Command

**Description**: Display project statistics — phases, plans, requirements, git metrics, and timeline

**Scope**: opencode

---

## Command Instructions

<objective>
Display comprehensive project statistics including phase progress, plan execution metrics, requirements completion, git history stats, and project timeline.
</objective>

<execution_context>
[file not found: C:\Users\user\.config\opencode\command\$HOME\.config\opencode\get-shit-done\workflows\stats.md]
</execution_context>

<process>
Execute the stats workflow from [file not found: C:\Users\user\.config\opencode\command\$HOME\.config\opencode\get-shit-done\workflows\stats.md] end-to-end.
</process>
</auto-slash-command><objective>
Display comprehensive project statistics including phase progress, plan execution metrics, requirements completion, git history stats, and project timeline.
</objective>

<execution_context>
@$HOME/.config/opencode/get-shit-done/workflows/stats.md
</execution_context>

<process>
Execute the stats workflow from @$HOME/.config/opencode/get-shit-done/workflows/stats.md end-to-end.
</process>

### What did you expect?

@&HOME被正确映射为用户目录

### Steps to reproduce

1.install GSD with npx
2.Select runtime:Opencode
3.run /gsd-plan
4.error appears

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

Major — Core feature is broken, no workaround

### Workaround (if any)

_No response_

### Additional context

_No response_

### Privacy Checklist

### LINKED ISSUE #2388: plan-phase init silently renames feature branch to match updated phase slug
## Summary

Running `/gsd-plan-phase` when the phase has been renamed since the feature branch was created silently creates a new local branch matching the updated phase slug and switches to it — without user confirmation or notification. The original feature branch (with remote tracking) is left behind, and subsequent commits land on the new local-only branch. On push, if the user pushes the original branch name, nothing goes to origin because the head hasn't moved on that branch; if the user pushes the new branch name, they've effectively split history across two branches by accident.

## Environment

- GSD version: **1.36.0**
- Harness: Claude Code
- OS: macOS
- Project uses the GSD v1 (`.planning/`) layout

## Reproduction

1. Start a phase via `/gsd-discuss-phase N` (feature branch `feature/gsd--phase-N-{original-slug}` is created from `develop`, tracks `origin/feature/gsd--phase-N-{original-slug}`).
2. During discuss-phase, rename the phase in `ROADMAP.md`; commit and push the rename on the original-named branch.
3. End session / `/clear` / `/gsd-resume-work`.
4. Invoke `/gsd-plan-phase N`.

**Observed:** During the `init plan-phase` call (or a subsequent `commit` call that feeds from init state), the shell session silently runs `git checkout -b feature/gsd--phase-N-{new-slug}` and switches to it. The git reflog shows `checkout: moving from feature/gsd--phase-N-{original-slug} to feature/gsd--phase-N-{new-slug}` with no visible command in the turn transcript.

**Expected:** Either (a) do nothing — branches are git identity, not phase identity, and a phase rename should not mutate branch names silently; or (b) explicitly prompt the user before renaming, with clear before/after and rollback instructions; or (c) perform a proper rename (`git branch -m` + re-push with tracking re-established + remote deletion of old branch) rather than creating a parallel new branch.

## Impact

- **Silent workflow state mutation.** The user did not ask for a branch change. The harness performed one without audit trail in the active turn.
- **Divergent local/remote state.** The new branch is local-only; origin still has the original branch at the pre-plan-phase state. Pushing the original name becomes a no-op; pushing the new name requires `-u` setup and leaves the old one orphaned.
- **Handoff artefacts become stale.** If `/gsd-pause-work` writes `branch: feature/gsd--phase-N-{original-slug}` (reading from STATE.md or HANDOFF.json from a prior session), and the actual current branch is now `feature/gsd--phase-N-{new-slug}`, the two fall out of sync and a resume-work pass has to reconcile manually.
- **Violates git identity conventions.** The original branch is the canonical git object; phase rename is a plan-level change, not a git-level change. A common convention is `feature/gsd--phase-N-{slug}` named at discuss-phase and left alone thereafter.

## Trigger mechanism

The most likely trigger is `gsd-tools init plan-phase` or a chained action from an `init` JSON result that sets `phase_slug` to the current slug and somewhere in the tool chain reconciles the branch name against the slug. The git reflog shows a checkout with no user-issued command between the previous commit and the next tool-originated commit.

## Suggested fixes

1. **Do not mutate git identity from plan-phase init.** Branches are created at discuss-phase and owned by the user's git workflow. Leave them alone.
2. If there's a strong reason to keep the branch name aligned with the slug, **offer the rename as an explicit action** with a visible prompt (e.g. "Phase N was renamed. Current branch is `{old}`; rename to `{new}`? [y/N]") and — if accepted — execute `git branch -m`, push with `--force-with-lease` under user awareness, delete the old remote branch with confirmation. Never silent-create a parallel branch.
3. Add a `workflow.branch_rename_on_phase_rename` config toggle (default `false`) for users who want the new behaviour as opt-in.

## Workaround

Fast-forward the original-named branch to the new-named branch's tip (linear history, no conflicts), delete the local-only new branch, push the original-named branch:

```bash
git checkout feature/gsd--phase-N-{original-slug}
git merge --ff-only feature/gsd--phase-N-{new-slug}
git branch -d feature/gsd--phase-N-{new-slug}
git push origin feature/gsd--phase-N-{original-slug}
```

This works cleanly when the two branches have linear history. A project that commits to the silent new branch without noticing, then returns to the old branch name, would land in a more painful state (`git pull` on old branch looks like it succeeded, commits appear lost, user has to rescue via reflog).

### LINKED ISSUE #2396: Hardcoded host-level test commands bypass container-only project Makefiles
**Affected:** `workflows/execute-phase.md` (post-merge gate ~L693, regression gate ~L1114), `workflows/verify-phase.md` (~L200), `workflows/audit-fix.md` (~L106).

**Severity:** High for container-only projects; cosmetic for host-runtime projects.

---

## Symptom

Workflows hardcode language-sniffed test commands (`npm test` / `cargo test` / `python -m pytest`) and run them on the host. Container-only projects (Python/Node tools live only inside Podman/Docker images, never on the host) hit silent false positives or false negatives — host `python` is missing or wrong, and the gate result is meaningless.

Secondary symptom: orchestrators that try to "fix" this by improvising `make shell <<< 'uv run pytest ...'` end up doing one cold-start container per command. 8 plans × 3 gates × ~30–60s = 12–24 min of container startup, often hitting token limits.

## Root cause

No mechanism for "use the project's canonical gate command." Workflows always sniff language artifacts and invoke the host runtime, even when the project ships a `make test` wrapper that already routes through the right container.

## Proposed fix

Resolve `TEST_CMD` in priority order, then run it:

1. `.planning/config.json` key `workflow.test_command` (per-project override)
2. `Makefile` with `test:` target → `make test` (also `Justfile`, `bun test`/`npm test`, etc.)
3. Language sniff (current behavior — fallback only)

Same pattern applies to `lint_command` / `typecheck_command` / `format_command`.

### Patch sketch — `execute-phase.md` step 5.6

```bash
TEST_CMD=$(node "$HOME/.claude/get-shit-done/bin/gsd-tools.cjs" config-get workflow.test_command --default "" 2>/dev/null)
if [ -z "$TEST_CMD" ]; then
  if   [ -f Makefile ] && grep -q "^test:" Makefile;     then TEST_CMD="make test"
  elif [ -f Justfile ] || [ -f justfile ];               then TEST_CMD="just test"
  elif [ -f package.json ] && command -v bun >/dev/null; then TEST_CMD="bun test"
  elif [ -f package.json ];                              then TEST_CMD="npm test"
  elif [ -f Cargo.toml ];                                then TEST_CMD="cargo test"
  elif [ -f go.mod ];                                    then TEST_CMD="go test ./..."
  elif [ -f pyproject.toml ] || [ -f requirements.txt ]; then TEST_CMD="python -m pytest -x -q --tb=short"
  else echo "⚠ No test runner detected"; TEST_CMD="true"  # no-op preserves TEST_EXIT=0 contract
  fi
fi
timeout 300 bash -c "$TEST_CMD" 2>&1
TEST_EXIT=$?
```

For `verify-phase.md`'s stricter "no runner = failure" semantic (currently `exit 1`), substitute `TEST_CMD="false"` in the no-runner branch.

**Non-breaking:** projects without config and without wrapper targets keep current behavior.

---

## Validated locally

Patched all four occurrences on a container-only Python project (`make test` routes through a long-lived Podman container via `make dev-up` + `podman exec`):

- **`gsd-tools.cjs config-get <key.path> --default <value>` is already implemented** — no new bin code, just the workflow change.
- **All four locations took the pattern cleanly.** `audit-fix.md` only had `npm test 2>&1 | tail -20`; replaced with the resolver, then piped to `tail -20`.
- **Subshell gotcha worth flagging** in the upstream patch: the original code wraps language-sniff in `timeout 300 bash -c '...'`, so `exit 0`/`exit 1` in the no-runner branch only exits the subshell. Once you pull the `if/elif/else` out to recipe top level, those `exit`s terminate the whole recipe and skip the downstream `TEST_EXIT` branching. `TEST_CMD="true"` / `"false"` is the drop-in fix (already incorporated above).
- **Side effect:** documenting `make test` as canonical also stops orchestrators from improvising the cold-start `make shell <<< 'cmd'` pattern that triggers the secondary symptom.

## Reproduction

Run `/gsd-execute-phase` on a container-only Python project (no host `python`/`uv`) with >1 plan per wave. Step 5.6 invokes `python -m pytest` against a missing host interpreter, returning `command not found` (or worse, succeeding against the wrong venv). The orchestrator then typically improvises 9–12 `podman run --rm ...` calls for follow-up debugging — each a cold container start.

### LINKED ISSUE #2431: quick.md: worktree teardown silently accumulates locked worktrees when CC session holds the lock
## Summary

`quick.md` Step 8's worktree teardown runs `git worktree remove --force 2>/dev/null || true`. The `2>/dev/null` silently swallows failure output. In Claude Code sessions, an increasingly-common failure mode goes undetected: CC's session-pid-keyed worktree lock survives subagent termination, so `git worktree remove --force` refuses the removal until the outer CC session exits — but the error is swallowed, the orchestrator proceeds as if cleanup succeeded, and the stale locked worktree accumulates on disk across sessions.

## Reproduction

1. Claude Code 2.1.112 on WSL2 (or any recent CC version on any OS — the lock mechanism is CC-side)
2. Run `/gsd-quick <task>` in a repo with `workflow.use_worktrees: true`
3. The executor subagent runs with `isolation: "worktree"`
4. CC writes `.git/worktrees/<agent-id>/locked` containing `claude agent agent-<id> (pid <outer-cc-pid>)` where `<outer-cc-pid>` is the live CC session PID
5. At `quick.md` Step 8 teardown, `git worktree remove "$WT" --force` exits non-zero with an error like:
   ```
   fatal: cannot remove working tree — the worktree is locked
   (use --reason to see the lock reason, or --lock for ...)
   ```
6. The `2>/dev/null || true` pattern suppresses the error and marks the cleanup step as succeeded
7. The worktree persists on disk, visible via `git worktree list` as `locked`
8. The lock only releases when the outer CC session exits — by which time nothing in the workflow is still trying to clean up

Third occurrence of this pattern in my 30-hour window as of 2026-04-19. Full forensic write-up in [reports/2026-04-19-worktree-leak-gh-36182-third-incident.md](https://github.com/anthropics/claude-code/issues/36182#issuecomment-4275489607) (referenced from CC issue #36182).

## Root cause

Two layers compound:

**Layer 1 (GSD, actionable here):** `quick.md:693` (and the parallel pattern in `execute-phase.md` / `execute-plan.md` step 5.5):
```bash
git worktree remove "$WT" --force 2>/dev/null || true
```
`2>/dev/null` hides the failure; `|| true` makes the exit non-breaking. Combined, the silent-fail pattern prevents downstream detection.

**Layer 2 (CC-side, not fixable here):** Claude Code installs a `locked` file containing the outer CC session's PID when dispatching `Agent(isolation: "worktree")`. The lock only releases on outer-session exit. Mechanism referenced but not documented: see `anthropics/claude-code#36182` thread for observational evidence.

Layer 2 is upstream's to fix. Layer 1 is the one that turns a surmountable CC quirk into silently-accumulating disk state.

## Proposed fix — `quick.md` Step 8

Replace the silent-fail pattern with a lock-aware retry block that surfaces residual failures:

```bash
# Before:
git worktree remove "$WT" --force 2>/dev/null || true

# After:
if ! git worktree remove "$WT" --force; then
  if [ -f ".git/worktrees/$(basename "$WT")/locked" ]; then
    echo "⚠ Worktree $WT is locked — attempting to unlock and retry"
    git worktree unlock "$WT" 2>/dev/null || true
    if ! git worktree remove "$WT" --force; then
      echo "⚠ Residual worktree at $WT — manual cleanup required after CC session exits:"
      echo "    git worktree unlock \"$WT\" && git worktree remove \"$WT\" --force && git branch -D \"$WT_BRANCH\""
    fi
  else
    echo "⚠ Residual worktree at $WT (not locked but remove failed) — investigate manually"
  fi
fi
```

The same pattern should apply to `execute-phase.md` and `execute-plan.md` cleanup steps — all three workflows currently share the silent-fail pattern (see `forks-worktree-safety.md` in the DHX skill repo for enumerated file list).

## Impact

- **Silent disk accumulation**: stale worktrees persist across sessions, consuming disk space (each worktree is a full working-tree copy).
- **Confusion during `/gsd-quick` reuse**: the next invocation can't detect or clean the previous session's residue.
- **Amplifies CC `#36182`**: residual worktrees from a leak incident stay visible but uncleanup-able, obscuring the leak-recovery audit trail.

## Alternatives considered

1. **Leave silent-fail in place, handle via external hook**: we're shipping a DHX SessionStart hook (`dhx-stale-worktree-sweep.sh`) that sweeps dead-PID locked worktrees with safety gates (clean tree + merged base + dead PID). Works, but fixes this at the wrong layer — any GSD user without the DHX skills repo still accumulates silent residue.
2. **Remove `--force`**: surfaces the error but then the workflow can't proceed at all when the lock is held. Worse UX.
3. **Strip `2>/dev/null` only, keep `|| true`**: exposes the error without changing control flow. Viable minimal fix, but the lock-aware retry provides actual recovery guidance. Either works.

## Environment

- GSD 1.37.1
- Claude Code 2.1.112
- WSL2 (Linux 6.6.87.2-microsoft-standard-WSL2)
- Shell: bash
- `workflow.use_worktrees: true`

Happy to submit the patch as a PR if maintainers agree on the shape. Pattern applies identically to `execute-phase.md` and `execute-plan.md` so either a shared helper or mirrored edits across all three files.

trek-e:
## Summary

- **#2388** — `plan-phase.md` silently renamed feature branches when `phase_slug` changed. Added a "Git Branch Invariant" section that explicitly forbids branch creation, renaming, or switching during plan-phase. Phase slug changes are plan-level only; the git branch is owned by the user's git workflow and must not be mutated.
- **#2431** — `quick.md` and `execute-phase.md` worktree teardown silently swallowed locked-worktree errors via `2>/dev/null || true`. Replaced with a lock-aware block: detects `.git/worktrees/*/locked`, attempts `git worktree unlock` + retry, and prints a user-visible recovery command when removal still fails.
- **#2396** — Hardcoded `npm test` / `cargo test` / `go test ./...` in `execute-phase.md`, `verify-phase.md`, and `audit-fix.md` bypassed container Makefiles. Added a three-tier resolver (project config → Makefile/Justfile → language sniff) so `make test` is preferred when a `test:` target exists.
- **#2376** — OpenCode on Windows does not expand `$HOME` in `@file` references, causing `MODULE_NOT_FOUND` at runtime. Added a platform guard in `bin/install.js` so OpenCode+win32 uses the resolved absolute path instead of the `$HOME`-prefixed form.

## Test plan

- [ ] `node --test tests/bug-2388-plan-phase-no-branch-rename.test.cjs` — 4/4 pass
- [ ] `node --test tests/bug-2431-worktree-locked-surfacing.test.cjs` — 10/10 pass
- [ ] `node --test tests/bug-2396-makefile-test-priority.test.cjs` — 9/9 pass
- [ ] `node --test tests/bug-2376-opencode-windows-home-path.test.cjs` — 6/6 pass
- [ ] Full `npm test` passes (all 29 new assertions plus existing suite)

Closes #2388
Closes #2431
Closes #2396
Closes #2376

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -5459,9 +5459,12 @@ function install(isGlobal, runtime = 'claude') {
   // For global installs: use $HOME/ so paths expand correctly inside double-quoted
   // shell commands (~ does NOT expand inside double quotes, causing MODULE_NOT_FOUND).
   // For local installs: use resolved absolute path (may be outside $HOME).
+  // Exception: OpenCode on Windows does not expand $HOME in @file references —
+  // use the absolute path instead so @$HOME/... references resolve correctly (#2376).
   const resolvedTarget = path.resolve(targetDir).replace(/\\/g, '/');
   const homeDir = os.homedir().replace(/\\/g, '/');
-  const pathPrefix = isGlobal && resolvedTarget.startsWith(homeDir)
+  const isWindowsHost = process.platform === 'win32';
+  const pathPrefix = isGlobal && resolvedTarget.startsWith(homeDir) && !(isOpencode && isWindowsHost)
     ? '$HOME' + resolvedTarget.slice(homeDir.length) + '/'
     : `${resolvedTarget}/`;
 
--- a/get-shit-done/workflows/audit-fix.md
+++ b/get-shit-done/workflows/audit-fix.md
@@ -103,7 +103,25 @@ Task(
 
 **b. Run tests:**
 ```bash
-npm test 2>&1 | tail -20
+AUDIT_TEST_CMD=$(gsd-sdk query config-get workflow.test_command --default "" 2>/dev/null || true)
+if [ -z "$AUDIT_TEST_CMD" ]; then
+  if [ -f "Makefile" ] && grep -q "^test:" Makefile; then
+    AUDIT_TEST_CMD="make test"
+  elif [ -f "Justfile" ] || [ -f "justfile" ]; then
+    AUDIT_TEST_CMD="just test"
+  elif [ -f "package.json" ]; then
+    AUDIT_TEST_CMD="npm test"
+  elif [ -f "Cargo.toml" ]; then
+    AUDIT_TEST_CMD="cargo test"
+  elif [ -f "go.mod" ]; then
+    AUDIT_TEST_CMD="go test ./..."
+  elif [ -f "pyproject.toml" ] || [ -f "requirements.txt" ]; then
+    AUDIT_TEST_CMD="python -m pytest -x -q --tb=short"
+  else
+    AUDIT_TEST_CMD="true"
+  fi
+fi
+eval "$AUDIT_TEST_CMD" 2>&1 | tail -20
 ```
 
 **c. If tests pass** — commit atomically:
--- a/get-shit-done/workflows/execute-phase.md
+++ b/get-shit-done/workflows/execute-phase.md
@@ -665,7 +665,19 @@ Execute each selected wave in sequence. Within a wave: parallel if `PARALLELIZAT
        fi
 
        # Remove the worktree
-       git worktree remove "$WT" --force 2>/dev/null || true
+       if ! git worktree remove "$WT" --force; then
+         WT_NAME=$(basename "$WT")
+         if [ -f ".git/worktrees/${WT_NAME}/locked" ]; then
+           echo "⚠ Worktree $WT is locked — attempting to unlock and retry"
+           git worktree unlock "$WT" 2>/dev/null || true
+           if ! git worktree remove "$WT" --force; then
+             echo "⚠ Residual worktree at $WT — manual cleanup required after session exits:"
+             echo "    git worktree unlock \"$WT\" && git worktree remove \"$WT\" --force && git branch -D \"$WT_BRANCH\""
+           fi
+         else
+           echo "⚠ Residual worktree at $WT (remove failed) — investigate manually"
+         fi
+       fi
 
        # Delete the temporary branch
        git branch -D "$WT_BRANCH" 2>/dev/null || true
@@ -688,22 +700,29 @@ Execute each selected wave in sequence. Within a wave: parallel if `PARALLELIZAT
    merging their work creates failures.
 
    ```bash
+   # Resolve test command: project config > Makefile > language sniff
+   TEST_CMD=$(gsd-sdk query config-get workflow.test_command --default "" 2>/dev/null || true)
+   if [ -z "$TEST_CMD" ]; then
+     if [ -f "Makefile" ] && grep -q "^test:" Makefile; then
+       TEST_CMD="make test"
+     elif [ -f "Justfile" ] || [ -f "justfile" ]; then
+       TEST_CMD="just test"
+     elif [ -f "package.json" ]; then
+       TEST_CMD="npm test"
+     elif [ -f "Cargo.toml" ]; then
+       TEST_CMD="cargo test"
+     elif [ -f "go.mod" ]; then
+       TEST_CMD="go test ./..."
+     elif [ -f "pyproject.toml" ] || [ -f "requirements.txt" ]; then
+       TEST_CMD="python -m pytest -x -q --tb=short 2>&1 || uv run python -m pytest -x -q --tb=short"
+     else
+       TEST_CMD="true"
+       echo "⚠ No test runner detected — skipping post-merge test gate"
+     fi
+   fi
    # Detect test runner and run quick smoke test (timeout: 5 minutes)
    TEST_EXIT=0
-   timeout 300 bash -c '
-   if [ -f "package.json" ]; then
-     npm test 2>&1
-   elif [ -f "Cargo.toml" ]; then
-     cargo test 2>&1
-   elif [ -f "go.mod" ]; then
-     go test ./... 2>&1
-   elif [ -f "pyproject.toml" ] || [ -f "requirements.txt" ]; then
-     python -m pytest -x -q --tb=short 2>&1 || uv run python -m pytest -x -q --tb=short 2>&1
-   else
-     echo "⚠ No test runner detected — skipping post-merge test gate"
-     exit 0
-   fi
-   '
+   timeout 300 bash -c "$TEST_CMD" 2>&1
    TEST_EXIT=$?
    if [ "${TEST_EXIT}" -eq 0 ]; then
      echo "✓ Post-merge test gate passed — no cross-plan conflicts"
@@ -1111,16 +1130,27 @@ Collect all unique test file paths into `REGRESSION_FILES`.
 **Step 3: Run regression tests (if any found)**
 
 ```bash
-# Detect test runner and run prior phase tests
-if [ -f "package.json" ]; then
-  npm test 2>&1
-elif [ -f "Cargo.toml" ]; then
-  cargo test 2>&1
-elif [ -f "go.mod" ]; then
-  go test ./... 2>&1
-elif [ -f "requirements.txt" ] || [ -f "pyproject.toml" ]; then
-  python -m pytest ${REGRESSION_FILES} -q --tb=short 2>&1
+# Resolve test command: project config > Makefile > language sniff
+REG_TEST_CMD=$(gsd-sdk query config-get workflow.test_command --default "" 2>/dev/null || true)
+if [ -z "$REG_TEST_CMD" ]; then
+  if [ -f "Makefile" ] && grep -q "^test:" Makefile; then
+    REG_TEST_CMD="make test"
+  elif [ -f "Justfile" ] || [ -f "justfile" ]; then
+    REG_TEST_CMD="just test"
+  elif [ -f "package.json" ]; then
+    REG_TEST_CMD="npm test"
+  elif [ -f "Cargo.toml" ]; then
+    REG_TEST_CMD="cargo test"
+  elif [ -f "go.mod" ]; then
+    REG_TEST_CMD="go test ./..."
+  elif [ -f "requirements.txt" ] || [ -f "pyproject.toml" ]; then
+    REG_TEST_CMD="python -m pytest ${REGRESSION_FILES} -q --tb=short"
+  else
+    REG_TEST_CMD="true"
+  fi
 fi
+# Detect test runner and run prior phase tests
+eval "$REG_TEST_CMD" 2>&1
 ```
 
 **Step 4: Report results**
--- a/get-shit-done/workflows/plan-phase.md
+++ b/get-shit-done/workflows/plan-phase.md
@@ -22,6 +22,10 @@ Valid GSD subagent types (use exact names — do not fall back to 'general-purpo
 
 <process>
 
+## 0. Git Branch Invariant
+
+**Do not create, rename, or switch git branches during plan-phase.** Branch identity is established at discuss-phase and is owned by the user's git workflow. A phase rename in ROADMAP.md is a plan-level change only — it does not mutate git branch names. If `phase_slug` in the init JSON differs from the current branch name, that is expected and correct; leave the branch unchanged.
+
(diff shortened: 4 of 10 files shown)


trek-e:

**#2388 — plan-phase branch rename guard**

Git Branch Invariant section is correctly placed at step 0, clearly prohibits branch creation/renaming/switching, and explains phase_slug divergence is expected. Tests verify prohibition text, phase-slug clarification, and absence of `git checkout -b`. No issues.

**#2431 — Worktree teardown error surfacing**

Lock-aware block is present in both `quick.md` and `execute-phase.md`. Checks `.git/worktrees/${WT_NAME}/locked`, attempts `git worktree unlock`, retries, and prints a user-visible recovery command. The old silent `2>/dev/null || true` pattern is removed from both files. No issues.

**#2396 — Makefile test priority**

Three-tier resolver applied consistently to execute-phase.md, verify-phase.md, and audit-fix.md. `eval "$TEST_CMD"` and `timeout 300 bash -c "$TEST_CMD"` work correctly for single commands. If a user sets `workflow.test_command` to a compound command with pipes, the `bash -c` form handles it correctly since the whole string is executed as a shell invocation. No issues.

**#2376 — OpenCode Windows $HOME expansion**

`!(isOpencode && isWindowsHost)` guard correctly scopes the exemption to OpenCode on Windows only. Claude Code on Windows and non-Windows OpenCode are unaffected. No issues.

**Cross-PR note:** PRs #2467 and #2469 both touch `quick.md` and `plan-phase.md` from different base commits. When #2467 and #2469 are merged, the second merge will need to verify that `plan-phase.md` contains both the Git Branch Invariant (from #2467) and the step 13c commit step (from #2469).

trek-e:
**Merge order note:** PR #2469 also modifies `quick.md` and `plan-phase.md`. Whoever merges second should verify the combined `plan-phase.md` contains both the Git Branch Invariant (from this PR) and step 13c commit-docs gate (from #2469).

Issue #1756, opened 2026-04-05
Worktree merge resurrects archived milestone state when branch outlives milestone transition


TheNimbleNeuron:
### GSD Version

1.32.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v22

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

A worktree executor branch was created during v2.0.0 milestone (Phase 21 execution). While the executor ran, the main branch archived v2.0.0 (`/gsd:complete-milestone`) and created a new v3.0.0 milestone with a fresh ROADMAP.md, STATE.md, and phases 21-24.

When the worktree branch merged back to main, git's 3-way merge silently:

1. **Overwrote ROADMAP.md** — replaced the v3.0.0 roadmap with stale v2.0.0 content (phases 9-16 instead of 21-24)
2. **Overwrote STATE.md** — replaced v3.0.0 state (milestone: v3.0.0, phase 21) with v2.0.0 state (milestone: v2.0.0, phase 14)
3. **Resurrected 150+ deleted files** — archived phase directories (01-14) that had been intentionally removed during milestone archival were re-added by the merge

This required multiple manual fix commits:
- `91b15e4` — "resolve merge conflicts (orchestrator owns STATE/ROADMAP)"
- `099d32a` — "fix worktree merge artifacts — restore phase 21 planning files, remove re-added old phases"
- `bb88dea` — "restore Plan 21-01 and 21-02 artifacts lost in worktree reset"

Despite these fixes, ROADMAP.md and STATE.md still ended up with stale v2.0.0 content on the current branch.

### How this differs from #1571

Issue #1571 (fixed in v1.31.0 via PR #1561) addresses **parallel agents racing on shared files within a wave** — solved by `withPlanningLock` and `readModifyWriteStateMd` making the orchestrator the single writer.

This bug is different: there's no write race. It's a **stale-snapshot problem**. The worktree branch was created when `.planning/` contained v2.0.0 artifacts. By the time it merged back, main had moved to v3.0.0. The v1.31 locking fix doesn't apply because:
- There's only one merge happening (not parallel writes)
- The problem is that the worktree branch carries an entire outdated `.planning/` tree
- Git's 3-way merge treats the worktree's version as a legitimate change, not a stale snapshot

### What did you expect?

[...]

PR #2503, opened 2026-04-21
fix(workflow): worktree merge must not delete legitimate new files (#2501)

### LINKED ISSUE #2501: Worktree merge step deletes legitimate SUMMARY.md files — resurrection-detection loop has inverted logic
### GSD Version

1.38.1

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.4.1

### Shell

/bin/zsh

### Installation Method

### What happened?

The worktree-cleanup block in `workflows/execute-phase.md` (introduced by PR #1764 for issue #1756, first shipped in v1.34.0) contains a resurrection-detection loop whose logic is the opposite of its stated intent. It stages a `git rm` against **every** newly-added file under `.planning/` — including legitimate `SUMMARY.md` files that executor agents just created.

The offending block in `workflows/execute-phase.md` (step 5.5, "Worktree cleanup"):

```bash
# Snapshot list of files on main BEFORE merge to detect resurrections
PRE_MERGE_FILES=$(git ls-files .planning/)
...
# Detect files deleted on main but re-added by worktree merge
# (e.g., archived phase directories that were intentionally removed)
DELETED_FILES=$(git diff --diff-filter=A --name-only HEAD~1 -- .planning/ 2>/dev/null || true)
for RESURRECTED in $DELETED_FILES; do
  # Check if this file was NOT in main's pre-merge tree
  if ! echo "$PRE_MERGE_FILES" | grep -qxF "$RESURRECTED"; then
    git rm -f "$RESURRECTED" 2>/dev/null || true
  fi
done
```

The comment says "files deleted on main but re-added by worktree merge". The logic does: "any file added by the merge AND not in main's pre-merge tree → delete".

But every legitimate new file added via worktree merge satisfies both conditions:
- Added by the merge (executor wrote it in the worktree and committed): ✓
- Not in main's pre-merge tree (it is a newly created file): ✓

So the loop stages `git rm` against every fresh `SUMMARY.md`, every new planning artifact, etc.

### Observed impact

During `/gsd-execute-phase` on a phase with parallel Wave 1 containing 3 plans (worktree mode enabled):

1. Merge of worktree 1 succeeded — `SUMMARY.md` landed in the merge commit, log showed `create mode 100644 .../07.9.1-01-SUMMARY.md`
2. Post-merge, the resurrection loop staged `rm '.planning/phases/.../07.9.1-01-SUMMARY.md'`
3. Amend-check condition only looked at `STATE.md`/`ROADMAP.md` diffs, so the staged deletion was **not** amended into the merge commit — it stayed in the index
4. Next `git merge worktree-2` aborted with `error: Your local changes to the following files would be overwritten by merge: .../07.9.1-01-SUMMARY.md`

This blocks parallel-mode entirely as soon as more than one plan per wave produces a new artifact. Silent data loss is also possible: if the amend path ever did trigger, the legitimate `SUMMARY.md` would be silently removed from the merge commit.

### Steps to reproduce

1. Configure a phase with 2+ parallel plans in Wave 1
2. Each plan must create a new file under `.planning/phases/<phase>/` (every GSD plan creates `SUMMARY.md`, so this is the default case)
3. Run `/gsd-execute-phase <phase>` with `workflow.use_worktrees: true` (default)
4. Observe: first worktree merges and stages `rm` on its own `SUMMARY.md`; second worktree's merge aborts

### Expected behavior

The loop should only `git rm` files that were **previously committed on main and subsequently deleted** (archived phase directories — the originally-motivating case from #1756). Legitimate new files produced by executor agents must remain in the merge commit.

### Minimal fix

Replace the condition with a history check:

```bash
ADDED_BY_MERGE=$(git diff --diff-filter=A --name-only HEAD~1..HEAD -- .planning/ 2>/dev/null || true)
for CANDIDATE in $ADDED_BY_MERGE; do
  # Resurrection = path was ever committed on main AND later deleted.
  # `git log HEAD~1 --diff-filter=D` returns paths deleted in main's history (pre-merge).
  if git log HEAD~1 --diff-filter=D --format= --name-only -- "$CANDIDATE" 2>/dev/null \
       | grep -qxF "$CANDIDATE"; then
    git rm -f "$CANDIDATE" 2>/dev/null || true
  fi
  # Else: brand-new file from worktree (SUMMARY.md etc.) — keep.
done
```

The snapshot `PRE_MERGE_FILES=$(git ls-files .planning/)` can be removed — it never actually served the resurrection check.

Happy to open a fix PR with a test that covers both positive (archived-directory resurrection) and negative (new `SUMMARY.md`) paths — waiting for `confirmed-bug` label per CONTRIBUTING.md.

### Introduced in

PR #1764 (merged 2026-04-05) shipped in v1.34.0. Logic unchanged through v1.38.1.

Histameany:
## Fix PR

> Opened as **draft** pending the `confirmed-bug` label on #2501. Happy to undraft the moment a maintainer confirms — the fix is small, tests pass locally, and reproducing the bug takes under a minute (see #2501).

---

## Linked Issue

Fixes #2501

## What was broken

The worktree-merge resurrection-detection block in `execute-phase.md` and `quick.md` staged `git rm` against every path that the merge commit added under `.planning/` — including legitimate new artifacts (e.g. `SUMMARY.md`). In parallel mode the next worktree merge then aborted with `Your local changes to the following files would be overwritten by merge`.

## What this fix does

Replaces the check with a history-based lookup: a path is a true resurrection only if it was previously committed on main **and** later deleted. Brand-new files from worktrees (which never existed on main) are kept.

Removes the now-unused `PRE_MERGE_FILES=$(git ls-files .planning/)` snapshot in both files, and renames the downstream variable from `DELETED_FILES` (which never actually held deletions) to `RESURRECTED_FILES` so the amend guard reads correctly.

## Root cause

The block introduced by #1764 used `git diff --diff-filter=A --name-only HEAD~1` (paths **added** by the merge) but named the variable `DELETED_FILES` and then checked `! echo "\$PRE_MERGE_FILES" | grep -qxF "\$RESURRECTED"`. Every new file trivially satisfies both conditions — nothing prevents fresh `SUMMARY.md` output from being flagged as a resurrection.

The intended semantic from #1756 was: detect paths main already removed (archived phase directories) that a stale worktree branch reintroduces. That requires checking history, not the current tree.

### How I verified the fix

1. Reproduced the bug against 1.38.1 on a real project: parallel-wave execute-phase with 3 plans → first merge staged `rm` on its own SUMMARY.md → second merge aborted. Traced to this exact block.
2. Applied the patch to the local install → same phase executed cleanly, no SUMMARY.md deletions, all merges completed.
3. Ran the full repo test suite locally: `npm test` → 4881 pass / 0 fail.

### Regression test added?

- [x] Yes — `tests/bug-2501-worktree-resurrection-false-positive.test.cjs`

The test asserts three invariants against both `execute-phase.md` and `quick.md`:
- The broken `PRE_MERGE_FILES` snapshot is gone.
- The broken `! echo "\$PRE_MERGE_FILES" | grep -qxF` condition is gone.
- The replacement uses `git log HEAD~1 --diff-filter=D` to detect true resurrections.
- The amend guard references the new `RESURRECTED_FILES` variable.

### Manual verification steps for reviewers

```bash
# Minimal repro of the original bug against unpatched workflow:
git init /tmp/repro && cd /tmp/repro
git commit --allow-empty -m "initial"
git worktree add ../wt-a
cd ../wt-a && mkdir -p .planning/phases/01 && echo "x" > .planning/phases/01/SUMMARY.md
git add . && git commit -m "add summary"
cd ../repro
git merge --no-ff wt-a -m "merge"
# Run the OLD resurrection loop → SUMMARY.md is staged for deletion.
# Run the NEW resurrection loop → SUMMARY.md is left alone (correct).
```

---

**Note on CONTRIBUTING compliance:** I'm aware CONTRIBUTING.md asks contributors to wait for `confirmed-bug` before the PR. I've filed #2501 first; this PR is intentionally a draft until that happens. The bug is reproducible in one minute and the fix is 14 lines + a test, so I'm hoping this speeds up triage rather than short-circuiting it — please redirect if you'd prefer me to close this until #2501 is confirmed.

### DIFF
--- a/get-shit-done/workflows/execute-phase.md
+++ b/get-shit-done/workflows/execute-phase.md
@@ -602,9 +602,6 @@ Execute each selected wave in sequence. Within a wave: parallel if `PARALLELIZAT
        [ -f .planning/STATE.md ] && cp .planning/STATE.md "$STATE_BACKUP" || true
        [ -f .planning/ROADMAP.md ] && cp .planning/ROADMAP.md "$ROADMAP_BACKUP" || true
 
-       # Snapshot list of files on main BEFORE merge to detect resurrections
-       PRE_MERGE_FILES=$(git ls-files .planning/)
-
        # Pre-merge deletion check: warn if the worktree branch deletes tracked files
        DELETIONS=$(git diff --diff-filter=D --name-only HEAD..."$WT_BRANCH" 2>/dev/null || true)
        if [ -n "$DELETIONS" ]; then
@@ -647,19 +644,25 @@ Execute each selected wave in sequence. Within a wave: parallel if `PARALLELIZAT
        fi
        rm -f "$STATE_BACKUP" "$ROADMAP_BACKUP"
 
-       # Detect files deleted on main but re-added by worktree merge
-       # (e.g., archived phase directories that were intentionally removed)
-       DELETED_FILES=$(git diff --diff-filter=A --name-only HEAD~1 -- .planning/ 2>/dev/null || true)
-       for RESURRECTED in $DELETED_FILES; do
-         # Check if this file was NOT in main's pre-merge tree
-         if ! echo "$PRE_MERGE_FILES" | grep -qxF "$RESURRECTED"; then
-           git rm -f "$RESURRECTED" 2>/dev/null || true
+       # Detect true resurrections: files added by the merge that were previously
+       # committed on main AND subsequently deleted (e.g., archived phase directories
+       # removed during milestone transition). Legitimate new files from worktrees
+       # (SUMMARY.md etc.) never existed on main and must be kept.
+       # Fix for #2501 — the previous check deleted every newly-added file.
+       ADDED_BY_MERGE=$(git diff --diff-filter=A --name-only HEAD~1..HEAD -- .planning/ 2>/dev/null || true)
+       RESURRECTED_FILES=""
+       for CANDIDATE in $ADDED_BY_MERGE; do
+         # Ask history (pre-merge main tip) whether this path was ever deleted there.
+         if git log HEAD~1 --diff-filter=D --format= --name-only -- "$CANDIDATE" 2>/dev/null \
+              | grep -qxF "$CANDIDATE"; then
+           git rm -f "$CANDIDATE" 2>/dev/null || true
+           RESURRECTED_FILES="$RESURRECTED_FILES $CANDIDATE"
          fi
        done
 
        # Amend merge commit with restored files if any changed
        if ! git diff --quiet .planning/STATE.md .planning/ROADMAP.md 2>/dev/null || \
-          [ -n "$DELETED_FILES" ]; then
+          [ -n "$RESURRECTED_FILES" ]; then
          # Only amend the commit with .planning/ files if commit_docs is enabled (#1783)
          COMMIT_DOCS=$(gsd-sdk query config-get commit_docs 2>/dev/null || echo "true")
          if [ "$COMMIT_DOCS" != "false" ]; then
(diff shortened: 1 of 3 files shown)


trek-e:
draft pre are not accepted, you can only submit when ready for review, please do not pollute the pull requests with draft PRs

Issue #2370, opened 2026-04-17
Bug: tests/read-guard.test.cjs fails inside Claude Code — test helper does not sanitize $CLAUDECODE


trek-e:
## Summary

Four tests in `tests/read-guard.test.cjs` fail with `SyntaxError: Unexpected end of JSON input` when `npm test` runs **inside a Claude Code session**. The failures are deterministic and reproduce on clean `origin/main` — they are not caused by any recent change.

## Failing tests

All four failures come from `tests/read-guard.test.cjs`:

1. `injects read-first guidance when Write targets an existing file` (line 64)
2. `injects read-first guidance when Edit targets an existing file` (line 85)
3. `guidance message includes the filename` (line 164)
4. `guidance message instructs to use Read tool before editing` (line 180)

All four fail at a `JSON.parse(result.stdout)` call (e.g. `tests/read-guard.test.cjs:173`) with:

```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at TestContext.<anonymous> (…/tests/read-guard.test.cjs:173:25)
```

## Root cause

The hook under test — `hooks/gsd-read-guard.js` — intentionally exits early and emits no stdout when it detects it is running inside Claude Code, because Claude Code already enforces read-before-edit natively (see #1984, #2344):

```js
// hooks/gsd-read-guard.js:40
if (process.env.CLAUDE_SESSION_ID || process.env.CLAUDECODE) {
  process.exit(0);
}
```

The test helper at `tests/read-guard.test.cjs:29–49` sanitizes `CLAUDE_SESSION_ID` via env overrides but **does not sanitize `CLAUDECODE`**:

```js
// tests/read-guard.test.cjs:32
const env = { ...process.env, CLAUDE_SESSION_ID: '', ...envOverrides };
```

Inside a Claude Code session, `CLAUDECODE=1` is exported automatically. It is inherited by `execFileSync` via `...process.env`, so the hook takes the early-exit branch, produces no output, and `JSON.parse('')` throws.

On CI (and outside Claude Code), neither `CLAUDE_SESSION_ID` nor `CLAUDECODE` is set, so the tests pass.

## Reproduction

Inside a Claude Code session:

```sh
node --test tests/read-guard.test.cjs
# → 4 failures
```

Outside a Claude Code session (or with `CLAUDECODE=` explicitly cleared):

```sh
CLAUDECODE= node --test tests/read-guard.test.cjs
# → all pass
```

[...]

Issue #2344, opened 2026-04-17
gsd-read-guard.js: CLAUDE_SESSION_ID check never fires in Claude Code (advisory spams every edit)


MSDNAndi:
## Summary

`~/.claude/hooks/gsd-read-guard.js` checks for `process.env.CLAUDE_SESSION_ID` to skip its advisory reminder when running inside Claude Code. However, Claude Code CLI actually exports `CLAUDECODE=1`, **not** `CLAUDE_SESSION_ID`. The skip never fires, so the "READ-BEFORE-EDIT REMINDER" injects on every Edit/Write call even though Claude Code natively enforces read-before-edit.

## Symptom

Every Edit/Write tool call inside Claude Code sees:
```
PreToolUse:Edit hook additional context: READ-BEFORE-EDIT REMINDER: You are about to modify "..." which already exists. If you have not already used the Read tool to read this file in the current session, you MUST Read it first before editing...
```

This reminder piles up noisily in long-running sessions — I saw it fire hundreds of times in a recent /batch run. The hook is advisory-only (`exit 0`), so it doesn't break anything, just makes the output noisy and can confuse sub-agent delegates.

## Root cause

At `hooks/gsd-read-guard.js:40`:

```js
// Claude Code natively enforces read-before-edit — skip the advisory (#1984)
if (process.env.CLAUDE_SESSION_ID) {
  process.exit(0);
}
```

Verified in my env (Claude Code CLI 2.1.112):
```
$ env | grep -i claude
CLAUDECODE=1
CLAUDE_CODE_ENTRYPOINT=cli
CLAUDE_CODE_EXECPATH=/home/erben/.local/share/claude/versions/2.1.112
```

No `CLAUDE_SESSION_ID` — so the check is always falsy and the advisory fires.

## Fix

One-line change — check both env vars:

```js
if (process.env.CLAUDE_SESSION_ID || process.env.CLAUDECODE) {
  process.exit(0);
}
```

Happy to send a PR if you prefer. GSD version: 1.36.0.

## Why this matters

Users with heavy Edit/Write workflows see dozens of these reminders per response. Worse, the reminders don't help — they fire after Read has already satisfied the native runtime check, so the message is actively misleading ("you MUST Read it first" — already done). Fixing this cleans up the noise without weakening the guard for non-Claude-Code hosts that the hook was designed for (OpenCode, MiniMax, etc.).

PR #2521, opened 2026-04-21
fix(hooks): detect Claude Code via stdin session_id (closes #2520)

### LINKED ISSUE #2520: hooks/gsd-read-guard.js: CLAUDECODE env check never fires inside Claude Code hook subprocesses (regression of #2344)
## Summary

The Claude Code exemption in `hooks/gsd-read-guard.js:40` does not skip the advisory on Claude Code v2.1.116. Every Write/Edit on an existing file produces a `READ-BEFORE-EDIT REMINDER` system-reminder even though Claude Code's native read-enforcement is active. Advisory-only — not blocking — but it pollutes context on every Edit.

Regression of the fix landed for #2344.

## Reproduction

1. Clean install GSD 1.38.3 (`npx get-shit-done-cc@latest --claude --global`).
2. Open Claude Code CLI 2.1.116.
3. Read + Edit any existing file.
4. Observe: every Edit produces a `<system-reminder>` containing `READ-BEFORE-EDIT REMINDER: ...`.

## Root cause — Claude Code's hook-subprocess env filter

#2344's fix added `|| process.env.CLAUDECODE` to the skip check. `CLAUDECODE=1` IS set by Claude Code and IS propagated to **Bash tool** subprocesses — but it is NOT propagated to **hook** subprocesses. Claude Code applies a separate env filter when spawning `PreToolUse` hook commands; that filter drops bare `CLAUDECODE` and `CLAUDE_SESSION_ID` and keeps only `CLAUDE_CODE_*`-prefixed vars plus `CLAUDE_PROJECT_DIR`.

Empirical dump from inside `gsd-read-guard.js` on Claude Code 2.1.116 (instrumented the hook, triggered an Edit):

```json
{
  "tool_name": "Edit",
  "session_id": "<UUID present via stdin>",
  "claude_env_keys": [
    "CLAUDE_CODE_EFFORT_LEVEL",
    "CLAUDE_CODE_ENTRYPOINT",
    "CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS",
    "CLAUDE_CODE_SSE_PORT",
    "CLAUDE_PROJECT_DIR"
  ],
  "CLAUDECODE": undefined,
  "CLAUDE_SESSION_ID": undefined,
  "CLAUDE_CODE_ENTRYPOINT": "cli",
  "CLAUDE_CODE_SSE_PORT": "51291"
}
```

Same Claude Code session, a Bash tool subprocess shows `CLAUDECODE=1` — so Bash and hook subprocesses see different env. The #2344 reporter observed `CLAUDECODE=1` in their shell and reasonably assumed hooks would inherit it; on 2.1.116 that assumption is false.

## Proposed fix

Use `data.session_id` from the hook's stdin JSON as the primary Claude Code signal. It is part of Claude Code's documented `PreToolUse` hook-input schema and is reliably present. Keep the env-var checks as fallbacks:

```js
// Claude Code natively enforces read-before-edit — skip the advisory (#1984, #2344)
const isClaudeCode =
  (typeof data.session_id === 'string' && data.session_id.length > 0) ||
  process.env.CLAUDE_CODE_ENTRYPOINT ||
  process.env.CLAUDE_CODE_SSE_PORT ||
  process.env.CLAUDE_SESSION_ID ||
  process.env.CLAUDECODE;
if (isClaudeCode) {
  process.exit(0);
}
```

Non-Claude-host behavior (OpenCode / MiniMax) is unchanged: if `data.session_id` is absent AND no `CLAUDE_CODE_*` env var is set, the advisory fires as before.

## Test-suite caveat

`tests/bug-2344-read-guard-claudecode-env.test.cjs` passes because its harness sets `CLAUDECODE=1` in the hook env directly. That matches the pre-filter shell env but not the post-filter hook env that Claude Code actually delivers. A regression test that mirrors the real filter (drop `CLAUDECODE` + `CLAUDE_SESSION_ID`; keep `CLAUDE_CODE_ENTRYPOINT` / `CLAUDE_CODE_SSE_PORT`; pass `session_id` via stdin) would have caught this. I'll include one in the PR.

## Environment

- GSD: 1.38.3
- Claude Code CLI: 2.1.116
- Node: 25.2.1
- OS: Linux

## Impact

Non-blocking, but every Edit on an existing file pushes a ~60-word `<system-reminder>` into the transcript. In sessions with many edits the cumulative attention-budget and context-window cost is meaningful.

## PR

Happy to send a PR with the fix + the regression test.

forfrossen:
Closes #2520.

## Problem

The fix landed for #2344 assumed `CLAUDECODE=1` would propagate from the Claude Code CLI into hook subprocesses. Empirically it doesn't — Claude Code v2.1.116 applies a separate env filter when spawning `PreToolUse` hook commands that drops bare `CLAUDECODE` and `CLAUDE_SESSION_ID` and keeps only `CLAUDE_CODE_*`-prefixed vars plus `CLAUDE_PROJECT_DIR`. Bash tool subprocesses DO still receive `CLAUDECODE=1`, which is why the #2344 reporter saw it and reasonably assumed hooks would too.

Result: every Write/Edit on an existing file inside Claude Code produces a redundant `READ-BEFORE-EDIT REMINDER` `<system-reminder>` block.

Empirical dump from inside `gsd-read-guard.js` on Claude Code CLI 2.1.116 (instrumented the hook, triggered an Edit):

```json
{
  "tool_name": "Edit",
  "session_id": "<UUID present via stdin>",
  "claude_env_keys": [
    "CLAUDE_CODE_EFFORT_LEVEL",
    "CLAUDE_CODE_ENTRYPOINT",
    "CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS",
    "CLAUDE_CODE_SSE_PORT",
    "CLAUDE_PROJECT_DIR"
  ],
  "CLAUDECODE": undefined,
  "CLAUDE_SESSION_ID": undefined
}
```

## Fix

Use `data.session_id` from the hook's stdin JSON as the primary Claude Code signal — it's part of Claude Code's documented `PreToolUse` hook-input schema and is reliably present. Keep `CLAUDE_CODE_ENTRYPOINT` / `CLAUDE_CODE_SSE_PORT` as env fallbacks (those DO reach hook subprocesses), and keep the legacy `CLAUDE_SESSION_ID` / `CLAUDECODE` checks for back-compat and in case future Claude Code versions restore them to the hook env.

```js
const isClaudeCode =
  (typeof data.session_id === 'string' && data.session_id.length > 0) ||
  process.env.CLAUDE_CODE_ENTRYPOINT ||
  process.env.CLAUDE_CODE_SSE_PORT ||
  process.env.CLAUDE_SESSION_ID ||
  process.env.CLAUDECODE;
if (isClaudeCode) {
  process.exit(0);
}
```

Non-Claude-host behavior (OpenCode / MiniMax) is unchanged: with no `session_id` on stdin and no `CLAUDE_CODE_*` env var, the advisory still fires.

## Tests

- **New:** `tests/bug-2520-read-guard-hook-subprocess-env.test.cjs` — spawns the hook with an env that mirrors the actual Claude Code hook-subprocess filter (drops `CLAUDECODE` / `CLAUDE_SESSION_ID`, keeps `CLAUDE_CODE_ENTRYPOINT` / `CLAUDE_CODE_SSE_PORT`, passes `session_id` via stdin). Covers: session_id signal alone, env-var fallback alone, and negative case (non-Claude host).
- **Extended:** the `runHook` helpers in `tests/read-guard.test.cjs` and `tests/bug-2344-read-guard-claudecode-env.test.cjs` now also strip `CLAUDE_CODE_ENTRYPOINT` / `CLAUDE_CODE_SSE_PORT` / `CLAUDE_PROJECT_DIR` from the subprocess env. Same class of leak as #2370 / PR #2375, but applied to the new detection signals so positive-path tests stay green when the test suite itself runs inside a Claude Code session.

Full suite: `npm test` → **4918 pass, 0 fail.**

## Not changed in this PR

Sibling guard hooks (`gsd-prompt-guard.js`, `gsd-workflow-guard.js`) do not currently check `CLAUDECODE` / `CLAUDE_SESSION_ID`, so the same regression does not apply to them. Happy to follow up if they gain similar detection in the future.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of external code-runner sessions by expanding checks to additional runtime signals and fallbacks, ensuring the read-before-edit advisory is consistently skipped when appropriate.

* **Tests**
  * Added regression tests and updated test helpers to validate session detection across varied runtime and subprocess environments.

### DIFF
--- a/hooks/gsd-read-guard.js
+++ b/hooks/gsd-read-guard.js
@@ -36,8 +36,27 @@ process.stdin.on('end', () => {
       process.exit(0);
     }
 
-    // Claude Code natively enforces read-before-edit — skip the advisory (#1984, #2344)
-    if (process.env.CLAUDE_SESSION_ID || process.env.CLAUDECODE) {
+    // Claude Code natively enforces read-before-edit — skip the advisory (#1984, #2344, #2520).
+    //
+    // Detection signals, in priority order:
+    //   1. `data.session_id` on the hook's stdin payload — part of Claude
+    //      Code's documented PreToolUse hook-input schema, always present.
+    //      Reliable across Claude Code versions because it's schema, not env.
+    //   2. `CLAUDE_CODE_ENTRYPOINT` / `CLAUDE_CODE_SSE_PORT` — env vars that
+    //      Claude Code does propagate to hook subprocesses (verified on
+    //      Claude Code CLI 2.1.116).
+    //   3. `CLAUDE_SESSION_ID` / `CLAUDECODE` — kept for back-compat and in
+    //      case future Claude Code versions propagate them to hook
+    //      subprocesses. On 2.1.116 they reach Bash tool subprocesses but
+    //      not hook subprocesses, which is why checking them alone is
+    //      insufficient (regression of #2344 fixed here as #2520).
+    const isClaudeCode =
+      (typeof data.session_id === 'string' && data.session_id.length > 0) ||
+      process.env.CLAUDE_CODE_ENTRYPOINT ||
+      process.env.CLAUDE_CODE_SSE_PORT ||
+      process.env.CLAUDE_SESSION_ID ||
+      process.env.CLAUDECODE;
+    if (isClaudeCode) {
       process.exit(0);
     }
 
--- a/tests/bug-2344-read-guard-claudecode-env.test.cjs
+++ b/tests/bug-2344-read-guard-claudecode-env.test.cjs
@@ -29,6 +29,9 @@ function runHook(payload, envOverrides = {}) {
     ...process.env,
     CLAUDE_SESSION_ID: '',
     CLAUDECODE: '',
+    CLAUDE_CODE_ENTRYPOINT: '',
+    CLAUDE_CODE_SSE_PORT: '',
+    CLAUDE_PROJECT_DIR: '',
     ...envOverrides,
   };
   try {
--- a/tests/bug-2520-read-guard-hook-subprocess-env.test.cjs
+++ b/tests/bug-2520-read-guard-hook-subprocess-env.test.cjs
@@ -1,0 +1,122 @@
+/**
+ * Regression test for bug #2520
+ *
+ * The fix for #2344 added `|| process.env.CLAUDECODE` to the Claude Code
+ * skip check. That works in principle — CLAUDECODE=1 is propagated to Bash
+ * tool subprocesses — but it does NOT reach hook subprocesses on Claude Code
+ * v2.1.116. Claude Code applies a separate env filter when spawning
+ * PreToolUse hook commands; that filter drops bare CLAUDECODE and
+ * CLAUDE_SESSION_ID and keeps only CLAUDE_CODE_*-prefixed vars plus
+ * CLAUDE_PROJECT_DIR. `data.session_id` is, however, reliably delivered via
+ * the hook's stdin JSON payload (documented part of Claude Code's hook
+ * input schema).
+ *
+ * Fix: use `data.session_id` as the primary Claude Code signal, with
+ * CLAUDE_CODE_ENTRYPOINT / CLAUDE_CODE_SSE_PORT as env-var fallbacks, and
+ * keep legacy CLAUDECODE / CLAUDE_SESSION_ID for back-compat and
+ * future-proofing.
+ */
+
+process.env.GSD_TEST_MODE = '1';
+
+const { test, describe, beforeEach, afterEach } = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const path = require('node:path');
+const { execFileSync } = require('node:child_process');
+
+const { createTempDir, cleanup } = require('./helpers.cjs');
+
+const HOOK_PATH = path.join(__dirname, '..', 'hooks', 'gsd-read-guard.js');
+
+/**
+ * Spawn the hook with an env that mirrors the actual Claude Code hook
+ * subprocess env: CLAUDECODE and CLAUDE_SESSION_ID are stripped, only
+ * CLAUDE_CODE_*-prefixed vars (plus CLAUDE_PROJECT_DIR) remain. Extra env
+ * overrides can be supplied via `envOverrides`.
+ */
+function runHookInClaudeCodeSubprocess(payload, envOverrides = {}) {
+  const input = JSON.stringify(payload);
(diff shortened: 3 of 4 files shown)


forfrossen:
Thanks for the review — legitimate finding. The session_id-isolation test was passing via the env fallback rather than exercising the primary signal. Fixed in e13d527: the case now passes `{ CLAUDE_CODE_ENTRYPOINT: '', CLAUDE_CODE_SSE_PORT: '', CLAUDE_PROJECT_DIR: '' }` to the helper, so only the stdin `session_id` can trigger the skip. Verified that the test would fail against the pre-fix hook (env-only check) since both legacy vars are stripped by the helper and the propagation-verified fallbacks are now cleared explicitly. Full suite still green (4918 pass).

trek-e (approved):
The fix is correct and complete.

**Signal priority design is right**:  from stdin is the proper primary signal — it's part of Claude Code's documented hook-input schema so it won't drift with env-var propagation behavior. The env-var fallbacks (, ) give a belt-and-suspenders fallback, and the legacy / are preserved for back-compat without being relied upon as the primary path.

**Test isolation is solid**: the  test helper correctly strips  and  and the session_id-isolation test explicitly clears  and  so each signal is tested independently. The three test cases cover all meaningful branches.

**Test update to  is correct**: clearing the two new env vars in the base env prevents false passes when the test runner is itself running inside Claude Code.

Approving.

trek-e (approved):
The fix is correct and complete.

Signal priority design is right: `data.session_id` from stdin is the proper primary signal — it is part of Claude Code's documented hook-input schema so it won't drift with env-var propagation behavior. The env-var fallbacks (`CLAUDE_CODE_ENTRYPOINT`, `CLAUDE_CODE_SSE_PORT`) give a belt-and-suspenders fallback, and the legacy `CLAUDECODE`/`CLAUDE_SESSION_ID` are preserved for back-compat without being relied upon as the primary path.

Test isolation is solid: the bug-2520 test helper correctly strips `CLAUDECODE` and `CLAUDE_SESSION_ID` and the session_id-isolation test explicitly clears `CLAUDE_CODE_ENTRYPOINT` and `CLAUDE_CODE_SSE_PORT` so each signal is tested independently. The three test cases cover all meaningful branches.

Test update to bug-2344 is correct: clearing the two new env vars in the base env prevents false passes when the test runner is itself running inside Claude Code.

PR #2576, opened 2026-04-22
fix(#2544): config-get exits 1 on missing key (UNIX convention)

### LINKED ISSUE #2544: gsd-sdk query config-get: writes 'Error: Key not found' to stderr but exits 0 — callers can't detect missing keys
### Summary

`gsd-sdk query config-get <key>` on a missing key:

- writes `Error: Key not found: <key>` to **stderr**
- writes nothing to stdout
- **exits 0** (success)

This is a broken contract. Callers cannot distinguish \"key absent\" from \"key present with empty value\" without parsing the stderr string, and the `Error:` prefix makes every missing-key probe look like a failure in interactive logs. It also diverges from the established UNIX convention for config lookups (`git config --get` → exit 1, silent).

### Evidence

```
$ gsd-sdk query config-get workflow.discuss_mode 2>&1
Error: Key not found: workflow.discuss_mode
$ echo $?
0

$ git config --get nonexistent.key 2>&1
$ echo $?
1
```

This surfaces during normal GSD workflows. `workflows/plan-phase.md` probes several config keys including `workflow.discuss_mode`, `workflow.ai_integration_phase`, etc. In a fresh project where these keys were never set, every probe logs an `Error:` line, creating visible noise during interactive sessions:

```
⏺ Bash(gsd-sdk query config-get workflow.discuss_mode 2>/dev/null; echo \"---\"; ...)
  ⎿  Error: Exit code 10
     ---
     ---
     ---
```

### Impact

1. **Log noise in interactive CC sessions.** Every unset-key probe writes `Error: ...` to the workflow transcript, alarming to users and hiding real errors.
2. **Callers cannot rely on exit code.** A `gsd-sdk query config-get k || fallback` pattern is broken — the `||` branch never fires.
3. **Stdout is empty on both missing and present-empty-string cases**, with no way to disambiguate.

### Expected behavior (any of these)

**Option A — match `git config --get` convention (preferred, most compatible):**
- stdout: empty
- stderr: empty
- exit code: 1

**Option B — add a `--default <value>` flag:**
```
$ gsd-sdk query config-get workflow.discuss_mode --default interactive
interactive         # stdout; exit 0
```
When the flag is given, missing key returns the default cleanly and exits 0.

**Option C — both of the above.** A is the base contract; B is a convenience on top.

### Suggested fix

In the `config-get` query handler (per `sdk/src/query/QUERY-HANDLERS.md` registry — file not listed from where I could check but the handler source should be easy to find), change:

```js
// Current (approximate)
if (!(key in config)) {
  console.error(`Error: Key not found: ${key}`);
  return; // exit 0
}
```

To:

```js
if (!(key in config)) {
  if (opts.default !== undefined) {
    process.stdout.write(opts.default + '\\n');
    return;
  }
  process.exit(1); // UNIX convention; no stderr noise for expected-absent case
}
```

### Related

Separate from #2525 (gsd-sdk chmod bug) and #2543 (legacy `/gsd-<cmd>` syntax in 102 places) but surfaced by the same upgrade audit.

trek-e:
Closes #2544

## Root cause
`gsd-sdk query config-get` was writing `Error: Key not found` to stderr but exiting with code 0 on a missing key. Callers using `if ! gsd-sdk query config-get key; then` could not detect the missing key, causing silent config dropout.

## Fix
Exit with code 1 (standard UNIX error convention) when a requested config key is not found, so callers can detect missing keys with standard shell exit-code checks.

## Test
`node --test tests/bug-2544-config-get-exit-code.test.cjs`

## Summary by CodeRabbit

* **New Features**
  * Added requirement traceability validation to detect and warn about unregistered requirement IDs in REQUIREMENTS.md.

* **Bug Fixes**
  * Fixed CLI entrypoint permissions on macOS/Unix.
  * Infrastructure/foundation phases now correctly auto-pass UAT verification.
  * Improved version parsing for milestone milestones to support optional third version segment.
  * Corrected error classification for missing configuration keys.
  * Enhanced executor model inheritance handling in workflows.
  * Extended artifact staging to include deferred items.

* **Tests**
  * Added regression test suites for recent bug fixes and requirement discovery.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -6780,18 +6780,20 @@ function installSdkIfNeeded() {
     emitSdkFatal('Failed to `npm run build` in sdk/.', { globalBin: null, exitCode: 1 });
   }
 
-  // 3. Install the built package globally so `gsd-sdk` lands on PATH.
-  const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
-  if (globalResult.status !== 0) {
-    emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
-  }
-
-  // Ensure tsc-emitted cli.js is executable after npm install -g creates the bin symlink.
+  // Ensure tsc-emitted cli.js is executable before npm install -g creates the bin symlink.
   // tsc emits .js files as 644; npm install -g creates the symlink but does not chmod the
   // target, so the kernel refuses to exec the file on macOS (issue #2525).
   const cliPath = path.join(sdkDir, 'dist', 'cli.js');
   if (fs.existsSync(cliPath)) {
-    try { fs.chmodSync(cliPath, 0o755); } catch (e) { /* Windows */ }
+    try { fs.chmodSync(cliPath, 0o755); } catch (e) {
+      if (process.platform !== 'win32') throw e;
+    }
+  }
+
+  // 3. Install the built package globally so `gsd-sdk` lands on PATH.
+  const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
+  if (globalResult.status !== 0) {
+    emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
   }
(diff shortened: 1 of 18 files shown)


trek-e:

PR #2579, opened 2026-04-22
fix(#2554): preserve leading zero before decimal in getMilestonePhaseFilter

### LINKED ISSUE #2554: state disk-scan excludes decimal phase dirs (e.g. "00.1") from progress counts
## Summary

`getMilestonePhaseFilter` in `bin/lib/core.cjs` builds a normalized set of ROADMAP phase IDs using `n.replace(/^0+/, '')`, which over-strips for decimal phase IDs. `"00.1"` collapses to `".1"`, while the disk-side extractor `/^0*(\d+[A-Za-z]?(?:\.\d+)*)/` applied to a directory like `"00.1-<slug>"` yields `"0.1"`. Set membership fails and the directory is excluded from every disk scan inside `buildStateFrontmatter` → `state json` / `state update` / `state patch`.

## Impact

Any project that uses `phase insert` / inserted decimal phases (e.g. `### Phase 00.1:` in ROADMAP.md + `phases/00.1-<slug>/` on disk) gets progress counters silently under-reported on every state-mutating SDK call. In our project:

- Before any `state update`: `completed_phases: 2, total_plans: 15, completed_plans: 15` (2 phases × 8+7 plans).
- After `gsd-sdk query state.update "Last Activity" "2026-04-21"`: counters get rewritten to `completed_phases: 1, total_plans: 8, completed_plans: 8`.

So simply touching `Last Activity` silently rewinds progress as if Phase 00.1 never happened. `stopped_at` also gets lost. This regression is triggered by `readModifyWriteStateMd` → `syncStateFrontmatter` → `buildStateFrontmatter`, which runs a fresh disk scan on every write.

## Reproduction

```js
// Exactly what getMilestonePhaseFilter does today:
const phaseIds = ["0", "00.1", "1"];
const normalized = new Set(
  phaseIds.map(n => (n.replace(/^0+/, '') || '0').toLowerCase())
);
// => Set { '0', '.1', '1' }

// What isDirInMilestone extracts from the directory:
"00.1-app-namespace-rename".match(/^0*(\d+[A-Za-z]?(?:\.\d+)*)/)[1];
// => "0.1"   ← NOT in the Set above
```

The asymmetry between the two regexes is the whole bug.

## Fix

Strip leading zeros only when followed by a digit:

```diff
 const normalized = new Set(
-  [...milestonePhaseNums].map(n => (n.replace(/^0+/, '') || '0').toLowerCase())
+  [...milestonePhaseNums].map(n => (n.replace(/^0+(?=\d)/, '') || '0').toLowerCase())
 );
```

Post-fix behavior:

| Input    | Before  | After   | `isDirInMilestone` extraction | Match? |
|----------|---------|---------|-------------------------------|--------|
| `"0"`    | `"0"`   | `"0"`   | `"0"` (from `"0-foo"`)        | ✓ (unchanged) |
| `"01"`   | `"1"`   | `"1"`   | `"1"` (from `"01-foo"`)       | ✓ (unchanged) |
| `"00.1"` | `".1"`  | `"0.1"` | `"0.1"` (from `"00.1-foo"`)   | ✓ (fixed) |
| `"10"`   | `"10"`  | `"10"`  | `"10"` (from `"10-foo"`)      | ✓ (unchanged) |

Verified locally: after the one-line patch, `gsd-sdk state json` returns the correct counters and repeated `state update "Last Activity" ...` calls no longer regress `completed_phases` / `total_plans` / `completed_plans`.

## Version

- GSD **1.38.3** (npm `get-shit-done-cc@1.38.3`)
- File: `bin/lib/core.cjs:1566` (current master may differ by a few lines)
- Trigger: projects created with `/gsd-insert-phase` that produce decimal phase IDs

Happy to open a PR if the fix direction looks right.

trek-e:
Closes #2554

## Root cause
`getMilestonePhaseFilter` in the SDK stripped leading zeros from phase directory names, causing `00.1` to be treated as `0.1` which didn't match the actual directory. This excluded decimal phase dirs (e.g. `00.1`) from progress counts.

## Fix
Preserve the leading zero in the phase filter regex so decimal phase dirs with zero-padded prefixes are correctly included in progress scans.

## Test
`node --test tests/bug-2554-decimal-phase-filter.test.cjs`

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -6780,18 +6780,20 @@ function installSdkIfNeeded() {
     emitSdkFatal('Failed to `npm run build` in sdk/.', { globalBin: null, exitCode: 1 });
   }
 
-  // 3. Install the built package globally so `gsd-sdk` lands on PATH.
-  const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
-  if (globalResult.status !== 0) {
-    emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
-  }
-
-  // Ensure tsc-emitted cli.js is executable after npm install -g creates the bin symlink.
+  // Ensure tsc-emitted cli.js is executable before npm install -g creates the bin symlink.
   // tsc emits .js files as 644; npm install -g creates the symlink but does not chmod the
   // target, so the kernel refuses to exec the file on macOS (issue #2525).
   const cliPath = path.join(sdkDir, 'dist', 'cli.js');
   if (fs.existsSync(cliPath)) {
-    try { fs.chmodSync(cliPath, 0o755); } catch (e) { /* Windows */ }
+    try { fs.chmodSync(cliPath, 0o755); } catch (e) {
+      if (process.platform !== 'win32') throw e;
+    }
+  }
+
+  // 3. Install the built package globally so `gsd-sdk` lands on PATH.
+  const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
+  if (globalResult.status !== 0) {
+    emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
   }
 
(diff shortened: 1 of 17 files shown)


trek-e:

PR #2580, opened 2026-04-22
fix(#2555): SDK agent-skills reads config.agent_skills and returns <agent_skills> block

### LINKED ISSUE #2555: SDK `agent-skills` query handler ignores `config.agent_skills[agentType]` — documented injection is a silent no-op for SDK-backed workflows
### Issue type

Wrong behavior (no longer matches docs)

### Where is the issue?

`sdk/src/query/skills.ts` (HEAD of main @ `7397f580a555491eb2ba0d4e51d8dafbd489a1db`), plus workflow call sites that now route through the SDK path instead of the legacy path.

### Summary

The SDK query handler `agent-skills` ignores `config.agent_skills[agentType]` entirely. It scans skill directories from the filesystem and returns **every** discovered skill as a flat JSON list, regardless of what the project has configured. The legacy `gsd-tools.cjs agent-skills <type>` path still works correctly and returns the documented `<agent_skills>` block. Since current workflows have been migrated to call the SDK path (e.g. `workflows/plan-phase.md:32-34`), the documented `agent_skills` injection feature is effectively a silent no-op for every SDK-backed workflow.

Net effect: users following `docs/CONFIGURATION.md#agent-skills-injection` verbatim — configuring `agent_skills.<agent-type>` in `.planning/config.json` and expecting the referenced skill to be injected into the subagent Task() prompt — get no injection. The config is accepted, stored, validated, and then discarded at read time.

This is a close cousin of #2397 (docs/validator/consumer drift) but is a distinct bug: here both the validator and the docs agree, and only the SDK consumer is incomplete.

### Reproduction

Minimal repro (in any project root with `.planning/`):

```bash
# 1. Create a project skill
mkdir -p .claude/skills/demo-skill
cat > .claude/skills/demo-skill/SKILL.md <<'SKILL'
---
name: demo-skill
description: demo
---
demo content
SKILL

# 2. Wire it to gsd-planner in .planning/config.json
#    "agent_skills": { "gsd-planner": [".claude/skills/demo-skill"] }

# 3. Compare the two code paths
echo "=== SDK path (used by current workflows) ==="
gsd-sdk query agent-skills gsd-planner

echo "=== Legacy path ==="
node ~/.claude/get-shit-done/bin/gsd-tools.cjs agent-skills gsd-planner
```

Observed output:

```
=== SDK path (used by current workflows) ===
{
  "agent_type": "gsd-planner",
  "skills": [ "demo-skill", "architect-preflight", "ask-architect-opencode",
              ...every skill found on the filesystem, ~90 entries in a populated project... ],
  "skill_count": 90
}

=== Legacy path ===
<agent_skills>
- @.claude/skills/demo-skill/SKILL.md
</agent_skills>
```

Expected (per `docs/CONFIGURATION.md#agent-skills-injection`):

Both paths should return the `<agent_skills>` block with **only** skills configured under `config.agent_skills[agentType]`, suitable for direct `${AGENT_SKILLS_PLANNER}`-style interpolation into Task() prompts. The SDK output is not a drop-in replacement — a JSON object with all skills cannot be embedded into the prompt without additional parsing the workflow does not perform.

### Evidence in code

`sdk/src/query/skills.ts` (and its compiled `dist/query/skills.js`) — the handler never references `config.agent_skills`:

```typescript
  const agentType = (args[0] || '').trim();
  if (!agentType) return { data: '' };
  const skillDirs = [
    join(projectDir, '.claude', 'skills'),
    join(projectDir, '.agents', 'skills'),
    // ...
  ];
  const skills: string[] = [];
  for (const dir of skillDirs) {
    // readdirSync and collect every sub-dir with a SKILL.md
  }
  return { data: { agent_type: agentType, skills: [...new Set(skills)], skill_count: ... } };
};
```

`sdk/src/config.ts:54,93,157-159` — the key is declared, merged, and stored.  
`sdk/src/query/config-mutation.ts:117-118` — the `agent_skills.<type>` write path is accepted via dynamic regex.  
But there is no code anywhere in `sdk/` that reads `config.agent_skills[agentType]` and uses it to filter or format the skills response. I ran `grep -rn "agent_skills\|agentSkills" sdk/src/` — every hit is either type declaration, merge/default, or the config-mutation validator. No consumer.

Meanwhile the legacy `bin/lib/init.cjs` `buildAgentSkillsBlock` function reads the config correctly, filters to configured paths only, validates path safety + SKILL.md existence, and formats the exact `<agent_skills>\nRead these user-configured skills:\n- @<path>/SKILL.md\n</agent_skills>` block the docs promise. The legacy `cmdAgentSkills` (`bin/lib/init.cjs:1611`) is the one the docs actually describe.

### Workflow call sites that hit the broken path

At least `workflows/plan-phase.md:32-34`:

```bash
AGENT_SKILLS_RESEARCHER=$(gsd-sdk query agent-skills gsd-researcher 2>/dev/null)
AGENT_SKILLS_PLANNER=$(gsd-sdk query agent-skills gsd-planner 2>/dev/null)
AGENT_SKILLS_CHECKER=$(gsd-sdk query agent-skills gsd-checker 2>/dev/null)
```

These values are then interpolated into Task() prompts via `${AGENT_SKILLS_PLANNER}` etc. (e.g. `plan-phase.md:724, :987`). Since the SDK output is a JSON object rather than the expected block, the subagent prompt ends up containing a full JSON dump of every filesystem-discovered skill, with no `agent_skills` filtering applied.

Other workflows invoking `agent-skills` (ripgrep on `workflows/`): `audit-milestone.md`, `diagnose-issues.md`, `discuss-phase-assumptions.md`, `discuss-phase.md`, `docs-update.md`, `execute-phase.md`, `map-codebase.md`, `new-milestone.md`, `new-project.md`. Spot checks show the same `$(gsd-sdk query agent-skills ...)` pattern.

### Docs vs reality

`docs/CONFIGURATION.md#agent-skills-injection` (on main at the commit above) says:

> At spawn time, workflows call `gsd-sdk query agent-skills <type>` (or legacy `node gsd-tools.cjs agent-skills <type>`) to load configured skills. If skills exist for the agent type, they are injected as an `<agent_skills>` block in the Task() prompt

The "or" phrasing implies both paths are equivalent. They are not. Only the legacy path honors `config.agent_skills`. The SDK path doesn't.

### Versions

- `@gsd-build/sdk`: `0.1.0`
- GSD plugin: `1.38.3`
- Verified against `main` @ `7397f580a555491eb2ba0d4e51d8dafbd489a1db` (the current HEAD of `main` at time of filing).

### Suggested fix

Port the `buildAgentSkillsBlock` semantics from `bin/lib/init.cjs` into `sdk/src/query/skills.ts`:

1. Read `config.agent_skills[agentType]` via the existing config loader.
2. If the entry is undefined/empty, return `{ data: '' }` (matches `cmdAgentSkills` when no agent type is configured).
3. For each configured path, apply the same validation the legacy function does (path traversal check, `global:` prefix, SKILL.md existence).
4. Return the formatted `<agent_skills>\nRead these user-configured skills:\n- @<path>\n</agent_skills>` block as `{ data: <string> }`, not as a structured object — workflows interpolate the value directly into prompts.
5. Keep backwards-compat for callers that expected the current (undocumented) "scan everything" behavior by exposing it under a separate query name, or drop it entirely if there are no consumers — a quick grep on `workflows/` suggests the only consumer is the to-be-fixed prompt interpolation path.

### Related

- #2397 — Config schema: docs vs validator vs consumer drift across 6+ documented keys (silent config drops). Same disease, different key. Section B of #2397 lists `agent_skills` as "Works at runtime, partially settable" — but does not call out that the runtime consumer on the SDK side is broken.
- #1992 — prior work to support `global:<name>` prefix for `agent_skills`. The legacy path implements this; the SDK path, by virtue of not reading the config at all, also skips `global:` handling.
- #1355 — original feature request for user skills/MCP injection into GSD subagents.

trek-e:
Closes #2555

## Root cause
The `agent-skills` query handler in the SDK ignored `config.agent_skills[agentType]` entirely — it returned an empty block regardless of what was configured. The documented per-agent skill injection was a silent no-op for SDK-backed workflows.

## Fix
The handler now reads `config.agent_skills[agentType]` from the project config and returns the configured skills list in the `<agent_skills>` XML block format expected by agents.

## Test
`node --test tests/bug-2555-agent-skills-config.test.cjs`

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -6780,18 +6780,20 @@ function installSdkIfNeeded() {
     emitSdkFatal('Failed to `npm run build` in sdk/.', { globalBin: null, exitCode: 1 });
   }
 
-  // 3. Install the built package globally so `gsd-sdk` lands on PATH.
-  const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
-  if (globalResult.status !== 0) {
-    emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
-  }
-
-  // Ensure tsc-emitted cli.js is executable after npm install -g creates the bin symlink.
+  // Ensure tsc-emitted cli.js is executable before npm install -g creates the bin symlink.
   // tsc emits .js files as 644; npm install -g creates the symlink but does not chmod the
   // target, so the kernel refuses to exec the file on macOS (issue #2525).
   const cliPath = path.join(sdkDir, 'dist', 'cli.js');
   if (fs.existsSync(cliPath)) {
-    try { fs.chmodSync(cliPath, 0o755); } catch (e) { /* Windows */ }
+    try { fs.chmodSync(cliPath, 0o755); } catch (e) {
+      if (process.platform !== 'win32') throw e;
+    }
+  }
+
+  // 3. Install the built package globally so `gsd-sdk` lands on PATH.
+  const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
+  if (globalResult.status !== 0) {
+    emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
   }
 
   // 4. Verify gsd-sdk is actually resolvable on PATH. npm's global bin dir is
--- a/bin/install.js
+++ b/bin/install.js
@@ -6780,6 +6780,14 @@ function installSdkIfNeeded() {
     emitSdkFatal('Failed to `npm run build` in sdk/.', { globalBin: null, exitCode: 1 });
   }
 
+  // Ensure tsc-emitted cli.js is executable before npm install -g creates the bin symlink.
+  // tsc emits .js files as 644; npm install -g creates the symlink but does not chmod the
+  // target, so the kernel refuses to exec the file on macOS (issue #2525).
+  const cliPath = path.join(sdkDir, 'dist', 'cli.js');
+  if (fs.existsSync(cliPath)) {
+    try { fs.chmodSync(cliPath, 0o755); } catch (e) { /* Windows */ }
+  }
+
   // 3. Install the built package globally so `gsd-sdk` lands on PATH.
   const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
   if (globalResult.status !== 0) {
--- a/bin/install.js
+++ b/bin/install.js
@@ -6780,18 +6780,18 @@ function installSdkIfNeeded() {
     emitSdkFatal('Failed to `npm run build` in sdk/.', { globalBin: null, exitCode: 1 });
   }
 
-  // Ensure tsc-emitted cli.js is executable before npm install -g creates the bin symlink.
+  // 3. Install the built package globally so `gsd-sdk` lands on PATH.
+  const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
+  if (globalResult.status !== 0) {
+    emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
+  }
+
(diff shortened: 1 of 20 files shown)


trek-e:

PR #2593, opened 2026-04-22
fix(#2544): config-get exits 1 silently for missing keys, adds --default flag

### LINKED ISSUE #2544: gsd-sdk query config-get: writes 'Error: Key not found' to stderr but exits 0 — callers can't detect missing keys

pleasedodisturb:
## Summary
- **Bug:** `gsd-sdk query config-get` threw `GSDError(Validation)` for missing keys — noisy `Error:` on stderr, exit code 10, breaking UNIX convention where absent config keys are expected
- **Fix A:** New `ErrorClassification.NotFound` (exit 1, silent stderr) — matches `git config --get` behavior
- **Fix B:** `--default <value>` flag returns a fallback on absent keys (exit 0). Values are JSON-parsed for type preservation (`"true"` → `true`, `"42"` → `42`) with string fallback
- **Security:** Reviewed — 5/5 threats closed (no injection, no masking, no path traversal)

## Test plan
- [x] 19/19 vitest tests pass (8 new: NotFound classification, --default variants, JSON type parsing, bare --default validation)
- [x] Existing `utils.test.ts` passes (10/10)
- [ ] Verify `gsd-sdk query config-get missing.key` exits 1 with no stderr
- [ ] Verify `gsd-sdk query config-get missing.key --default interactive` exits 0 with `"interactive"` on stdout
- [ ] Verify `gsd-sdk query config-get existing.key --default ignored` returns existing value

Closes #2544

### DIFF
--- a/sdk/src/cli.ts
+++ b/sdk/src/cli.ts
@@ -419,7 +419,11 @@ export async function main(argv: string[] = process.argv.slice(2)): Promise<void
       }
     } catch (err) {
       if (err instanceof GSDError) {
-        console.error(`Error: ${err.message}`);
+        // NotFound is an expected condition (e.g. config key absent) — exit 1
+        // silently, matching `git config --get` convention. No stderr noise.
+        if (err.classification !== ErrorClassification.NotFound) {
+          console.error(`Error: ${err.message}`);
+        }
         process.exitCode = exitCodeFor(err.classification);
       } else if (err instanceof GSDToolsError) {
         console.error(`Error: ${err.message}`);
--- a/sdk/src/errors.ts
+++ b/sdk/src/errors.ts
@@ -30,6 +30,9 @@ export enum ErrorClassification {
 
   /** Timeout, signal, user cancel. Exit code 1. */
   Interruption = 'interruption',
+
+  /** Lookup target does not exist (e.g. config key absent). Exit code 1, silent stderr. */
+  NotFound = 'not_found',
 }
 
 // ─── GSDError ───────────────────────────────────────────────────────────────
@@ -64,6 +67,8 @@ export function exitCodeFor(classification: ErrorClassification): number {
       return 10;
     case ErrorClassification.Blocked:
       return 11;
+    case ErrorClassification.NotFound:
+      return 1;
     case ErrorClassification.Execution:
     case ErrorClassification.Interruption:
     default:
--- a/sdk/src/query/config-query.test.ts
+++ b/sdk/src/query/config-query.test.ts
@@ -49,24 +49,106 @@ describe('configGet', () => {
     await expect(configGet([], tmpDir)).rejects.toThrow(GSDError);
   });
 
-  it('throws GSDError for nonexistent key', async () => {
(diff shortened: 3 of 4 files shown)


trek-e:
already addressed by maintainer

pleasedodisturb:
@trek-e how do I pick bugs which are worth doing? I'm new here (on the internet)

PR #2600, opened 2026-04-22
fix(#2598): pass shell: true to npm spawnSync on Windows

### LINKED ISSUE #2598: Installer: `spawnSync('npm.cmd', …)` without `shell: true` fails with EINVAL on Windows (CVE-2024-27980) — SDK build always fails on fresh install
## Summary

On Windows, `npx get-shit-done-cc@latest` **always fails at the SDK build step** on any Node.js version that includes the CVE-2024-27980 fix (Node ≥ 18.20.2 / ≥ 20.12.2 / ≥ 21.7.3, and every version since). The installer reports:

```
Building GSD SDK from source (…\node_modules\get-shit-done-cc\sdk)…

  Reason: Failed to `npm install` in sdk/.
```

Note there is **zero output from npm itself** — because `npm` never runs. The spawn call fails before it launches.

## Environment

| | |
|---|---|
| OS | Windows 11 Pro 10.0.29570 |
| Shell (reproducer) | Git Bash (MSYS) — also reproduces from cmd/PowerShell |
| Node | v25.9.0 |
| npm | 11.6.2 |
| get-shit-done-cc | 1.38.3 |
| Invocation | `npx get-shit-done-cc@latest --claude --global` |

## Root cause

`bin/install.js` has three `spawnSync` calls that launch `npm.cmd` on Windows, all **without `shell: true`**:

- `bin/install.js:6790` — `spawnSync(npmCmd, ['install'], { cwd: sdkDir, stdio: 'inherit' })`
- `bin/install.js:6796` — `spawnSync(npmCmd, ['run', 'build'], { cwd: sdkDir, stdio: 'inherit' })`
- `bin/install.js:6802` — `spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' })`

Since [CVE-2024-27980](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2), Node.js refuses to spawn `.cmd` / `.bat` files on Windows unless `shell: true` is passed. `spawnSync` with a `.cmd` target without shell returns:

```
{ status: null, signal: null, error: Error: spawnSync npm.cmd EINVAL }
```

The installer checks `if (installResult.status !== 0)` (line 6791) — `null !== 0` is true, so it trips the failure path. Because `stdio` is `'inherit'` and no child ever launched, the user sees a failure with no underlying diagnostic. The `installResult.error` is never surfaced.

## Minimal reproduction

```js
// repro.js — requires Node ≥ 18.20.2 / ≥ 20.12.2 / ≥ 21.7.3
const { spawnSync } = require('child_process');
const r = spawnSync('npm.cmd', ['--version'], { encoding: 'utf-8' });
console.log('status:', r.status);             // null
console.log('error:',  r.error && r.error.code); // EINVAL
```

Adding `shell: true` makes it work:

```js
const r = spawnSync('npm.cmd', ['--version'], { encoding: 'utf-8', shell: true });
console.log('status:', r.status);  // 0
console.log('stdout:', r.stdout);  // e.g. "11.6.2\n"
```

## Proposed fix

Add `shell: true` (or switch to a library that handles Windows `.cmd` resolution, e.g. `execa` / `cross-spawn`) to the three `spawnSync` calls at `bin/install.js:6790`, `:6796`, `:6802`. With `stdio: 'inherit'` and an argv array, `shell: true` on Windows is safe here because all arguments are static string literals — no user-controlled input.

A minimal patch:

```diff
- const installResult = spawnSync(npmCmd, ['install'], { cwd: sdkDir, stdio: 'inherit' });
+ const installResult = spawnSync(npmCmd, ['install'], { cwd: sdkDir, stdio: 'inherit', shell: true });
…
- const buildResult   = spawnSync(npmCmd, ['run', 'build'], { cwd: sdkDir, stdio: 'inherit' });
+ const buildResult   = spawnSync(npmCmd, ['run', 'build'], { cwd: sdkDir, stdio: 'inherit', shell: true });
…
- const globalResult  = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
+ const globalResult  = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit', shell: true });
```

Secondary improvement (not required, but recommended): when `status !== 0`, surface `installResult.error?.message` in `emitSdkFatal` so spawn-time failures aren't reported as "npm install failed" with no diagnostic. That would have saved several hours of debugging here.

Other `spawnSync` sites in the installer may deserve an audit — e.g. the `command -v` resolver at line 6687 uses `sh -c` (fine on POSIX), and the `where gsd-sdk` call at line 6681 spawns `where` (a built-in exe, not a `.cmd`, so unaffected). The three SDK-build spawns are the ones that break.

## Workaround (until fix lands)

This works and is persistent — once `gsd-sdk` is on PATH, the installer's `resolveGsdSdk()` check short-circuits the broken SDK build step on every future run:

```bash
# 1. Run the installer normally — it will install skills/agents/hooks successfully,
#    then fail at the SDK step. This is fine; all non-SDK state is already correct.
npx get-shit-done-cc@latest --claude --global

# 2. Manually complete the three commands the installer would have run. The SDK
#    source tree is sitting in the npx cache at the path the installer printed:
cd "$(ls -1d /c/Users/$USER/AppData/Local/npm-cache/_npx/*/node_modules/get-shit-done-cc/sdk | head -1)"
npm install && npm run build && npm install -g .

# 3. Verify:
which gsd-sdk           # should print …/AppData/Roaming/npm/gsd-sdk
gsd-sdk --version       # should print gsd-sdk v0.1.0
```

Note the executable bit on `dist/cli.js` is preserved here because Git Bash's `npm install -g .` happens to mark the installed file 0755 — so the separate #2453 chmod workaround isn't needed on this path. The previously-fixed fatal path (`installSdkIfNeeded`'s chmod-to-755 block at `bin/install.js:6807-6823`) would handle it anyway when the spawn bug is fixed.

## Impact

- **100% reproducible** on fresh installs for Windows users on any current LTS or Current Node release.
- The installer still installs skills, agents, hooks, and settings (all non-SDK state) successfully — it only fails the final SDK step — so users are left in a partially-installed state where `/gsd-help` lists commands that silently don't work because they shell out to `gsd-sdk query`.
- The user-facing error message (`Failed to \`npm install\` in sdk/`) actively misleads — it points at an npm install problem when the real failure is in the `spawnSync` call itself.

## Related

- Closed issues #2547, #2453, #2525, #2423 addressed the adjacent chmod / executable-bit issues on the SDK install, but they all assumed the `spawnSync` calls succeed — which on current Windows + Node, they never do.

trek-e:
## TL;DR

Every fresh Windows install has been silently failing at the SDK build step since Node shipped the CVE-2024-27980 fix. `spawnSync('npm.cmd', ...)` returns EINVAL before npm runs; the installer's `status !== 0` check trips and emits a bare failure with zero diagnostic because `stdio: 'inherit'` never got a child. Fix: route npm calls through a shell-aware wrapper.

## What

- `installSdkIfNeeded` picks `npmCmd = 'npm.cmd'` on win32 and spawns five times: `npm install`, `npm run build`, `npm install -g .`, two `npm config get prefix`.
- All five now go through a local `spawnNpm(args, opts)` helper that injects `shell: process.platform === 'win32'` when the caller doesn't override it.
- Behavior on non-Windows is unchanged (no shell option forced).

## Why

Reporter (@Grips001) traced this exactly — `spawnSync('npm.cmd', ['--version'])` on Node 25.9 returns `{ status: null, error: EINVAL }`, and `null !== 0` trips the installer's failure path. The user sees:

```
Building GSD SDK from source (…\node_modules\get-shit-done-cc\sdk)…
  Reason: Failed to `npm install` in sdk/.
```

…with no npm output at all, because npm never ran.

This affects every Windows install on Node >= 18.20.2 / >= 20.12.2 / >= 21.7.3 (the CVE fix has been in every Node release for nearly two years).

### Regression test (`tests/bug-2598-windows-npm-spawn-shell.test.cjs`)

Static source-level invariants on `installSdkIfNeeded`:

1. No bare `spawnSync(npmCmd, ...)` call (excluding the helper implementation itself which explicitly forwards the shell option).
2. A shell-aware wrapper must exist (`spawnNpm` identifier or an explicit `shell:` option).
3. At least 5 npm invocations route through the wrapper.

All 4 tests pass. Spot-run alongside 15 existing install tests — all green.

### Test results

```
node --test tests/bug-2453-sdk-cli-chmod.test.cjs \
            tests/bug-2598-windows-npm-spawn-shell.test.cjs \
            tests/agent-install-validation.test.cjs
tests 19  pass 19  fail 0
```

## Docs

No user-facing behavior change on non-Windows; Windows users go from "silent install failure" to "working install." No doc update required — the install command in README is unchanged.

Closes #2598

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed npm command execution issues on Windows that could cause silent installation and build failures.

* **Tests**
  * Added regression tests to ensure Windows npm execution reliability remains stable.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -6786,20 +6786,30 @@ function installSdkIfNeeded() {
   console.log(`\n  ${cyan}Building GSD SDK from source (${sdkDir})…${reset}`);
   const npmCmd = process.platform === 'win32' ? 'npm.cmd' : 'npm';
 
+  // Windows: Node.js refuses to spawn .cmd/.bat files without `shell: true`
+  // after CVE-2024-27980 (fixed in Node ≥ 18.20.2 / ≥ 20.12.2 / ≥ 21.7.3).
+  // Without shell, spawnSync returns { status: null, error: EINVAL } and
+  // every `status !== 0` check trips — producing a silent build failure
+  // with no underlying diagnostic because stdio: 'inherit' never gets a
+  // child to stream (#2598).
+  const needsShell = process.platform === 'win32';
+  const spawnNpm = (args, opts = {}) =>
+    spawnSync(npmCmd, args, { ...opts, shell: opts.shell ?? needsShell });
+
   // 1. Install sdk build-time dependencies (tsc, etc.)
-  const installResult = spawnSync(npmCmd, ['install'], { cwd: sdkDir, stdio: 'inherit' });
+  const installResult = spawnNpm(['install'], { cwd: sdkDir, stdio: 'inherit' });
   if (installResult.status !== 0) {
     emitSdkFatal('Failed to `npm install` in sdk/.', { globalBin: null, exitCode: 1 });
   }
 
   // 2. Compile TypeScript → sdk/dist/
-  const buildResult = spawnSync(npmCmd, ['run', 'build'], { cwd: sdkDir, stdio: 'inherit' });
+  const buildResult = spawnNpm(['run', 'build'], { cwd: sdkDir, stdio: 'inherit' });
   if (buildResult.status !== 0) {
     emitSdkFatal('Failed to `npm run build` in sdk/.', { globalBin: null, exitCode: 1 });
   }
 
   // 3. Install the built package globally so `gsd-sdk` lands on PATH.
-  const globalResult = spawnSync(npmCmd, ['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
+  const globalResult = spawnNpm(['install', '-g', '.'], { cwd: sdkDir, stdio: 'inherit' });
   if (globalResult.status !== 0) {
     emitSdkFatal('Failed to `npm install -g .` from sdk/.', { globalBin: null, exitCode: 1 });
   }
@@ -6811,7 +6821,7 @@ function installSdkIfNeeded() {
   // a non-executable file and `command -v gsd-sdk` fails on every first install
   // (root cause of #2453). Mirrors the pattern used for hook files in this installer.
   try {
-    const prefixRes = spawnSync(npmCmd, ['config', 'get', 'prefix'], { encoding: 'utf-8' });
+    const prefixRes = spawnNpm(['config', 'get', 'prefix'], { encoding: 'utf-8' });
     if (prefixRes.status === 0) {
       const npmPrefix = (prefixRes.stdout || '').trim();
       const sdkPkg = JSON.parse(fs.readFileSync(path.join(sdkDir, 'package.json'), 'utf-8'));
@@ -6835,7 +6845,7 @@ function installSdkIfNeeded() {
   }
 
   // Off-PATH: resolve npm global bin dir for actionable remediation.
-  const prefixResult = spawnSync(npmCmd, ['config', 'get', 'prefix'], { encoding: 'utf-8' });
+  const prefixResult = spawnNpm(['config', 'get', 'prefix'], { encoding: 'utf-8' });
   const prefix = prefixResult.status === 0 ? (prefixResult.stdout || '').trim() : null;
   const globalBin = prefix
     ? (process.platform === 'win32' ? prefix : path.join(prefix, 'bin'))
--- a/bin/install.js
+++ b/bin/install.js
@@ -6796,22 +6796,44 @@ function installSdkIfNeeded() {
   const spawnNpm = (args, opts = {}) =>
     spawnSync(npmCmd, args, { ...opts, shell: opts.shell ?? needsShell });
 
+  // Format the underlying spawnSync failure so EINVAL / ENOENT / signal exits
+  // surface in the fatal banner instead of being swallowed. The #2598 silent
+  // failure happened precisely because `{ status: null, error: EINVAL }` was
+  // reduced to a generic "Failed to npm install" with no diagnostic — the real
+  // cause (CVE-2024-27980 on Windows) was invisible in the output.
+  const formatSpawnFailure = (result) => {
+    if (!result) return '';
+    if (result.error) return ` (${result.error.code || result.error.name || 'spawn error'}: ${result.error.message})`;
+    if (result.signal) return ` (signal: ${result.signal})`;
+    if (typeof result.status === 'number') return ` (exit status: ${result.status})`;
+    return '';
+  };
+
   // 1. Install sdk build-time dependencies (tsc, etc.)
   const installResult = spawnNpm(['install'], { cwd: sdkDir, stdio: 'inherit' });
   if (installResult.status !== 0) {
-    emitSdkFatal('Failed to `npm install` in sdk/.', { globalBin: null, exitCode: 1 });
+    emitSdkFatal(
+      `Failed to \`npm install\` in sdk/.${formatSpawnFailure(installResult)}`,
+      { globalBin: null, exitCode: 1 },
+    );
   }
 
   // 2. Compile TypeScript → sdk/dist/
   const buildResult = spawnNpm(['run', 'build'], { cwd: sdkDir, stdio: 'inherit' });
   if (buildResult.status !== 0) {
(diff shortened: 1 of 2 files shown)


trek-e:
## CodeRabbit review — resolution

| Finding | Severity | Action | Commit / Reason |
|---|---|---|---|
| [r3126987409](https://github.com/gsd-build/get-shit-done/pull/2600#discussion_r3126987409) — `bin/install.js` L6800-6814: fatal `emitSdkFatal` branches swallow `result.error` / `result.status`, so EINVAL/ENOENT/signal exits print a generic message with zero diagnostic (the exact failure mode that kept #2598 silent in the first place). | Potential issue / Minor | Fixed | ab67e245 — added `formatSpawnFailure(result)` helper that formats `result.error.code/message`, `result.signal`, or numeric `result.status`; threaded into all three fatal branches (`npm install`, `npm run build`, `npm install -g .`). |
| [r3126987419](https://github.com/gsd-build/get-shit-done/pull/2600#discussion_r3126987419) — `tests/bug-2598-windows-npm-spawn-shell.test.cjs` L81-94: `explicitShellNpm` regex overcounts by matching the `spawnNpm` helper body itself, so the `>= 5` invariant could pass with only 4 real call sites. | Potential issue / Minor | Fixed | ab67e245 — split helper detection into arrow-form (`const spawnNpm =`) vs function-form (`function spawnNpm`); exclude wrapper body from `explicitShellNpm` when a helper exists. Added a new test asserting all three fatal branches reference `formatSpawnFailure` / `result.error` / `signal` / `status`. |

Full suite (4992 tests) green locally; coverage threshold met.

trek-e:
## CodeRabbit follow-up review — status

Re-scanned all CodeRabbit findings on this PR after `ab67e245`. No new findings have been posted (CodeRabbit is rate-limited on the new commit per its auto-comment, but no inline review has landed since).

| Finding | Location | Status | Action |
|---|---|---|---|
| `r3126987409` — fatal paths swallow underlying spawnSync diagnostics | `bin/install.js` ~L6814 (pre-fix anchor) | Fixed in `ab67e245` | None (anchor outdated — `line: null` against HEAD; `formatSpawnFailure` now threads `result.error` / `signal` / `status` into all 3 fatal branches) |
| `r3126987419` — `explicitShellNpm` overcounts via helper definition match | `tests/bug-2598-windows-npm-spawn-shell.test.cjs` ~L94 | Fixed in `ab67e245` | None (helper detection split into arrow vs function form; wrapper body excluded from `explicitShellNpm` when a helper exists) |

### Adversarial re-check on `ab67e245` (no new findings)

- **Shell argument escaping** — All five `spawnNpm` call sites pass static literal arg arrays (`['install']`, `['run','build']`, `['install','-g','.']`, `['config','get','prefix']`). Zero user-controlled input reaches the shell. No injection surface.
- **`.cmd` resolution** — Node's `spawn` with `shell: true` on Windows uses `cmd.exe` (not PowerShell), which resolves `npm.cmd` on `PATH` correctly.
- **Spawn-site coverage** — `npmCmd` is local to `installSdkIfNeeded`; the only raw `spawnSync(npmCmd, ...)` in the file is inside the `spawnNpm` wrapper itself, which forwards `shell` from `opts`. All five documented invocations route through it.
- **`formatSpawnFailure` edge cases** — Handles `!result`, falls back through `error.code → error.name → 'spawn error'`, gates `status` on `typeof === 'number'` (so `null` doesn't render as `"exit status: null"`), and returns `''` when no diagnostic field is present. Safe to append unconditionally.
- **`opts.shell ?? needsShell`** — Uses `??` (not `||`), so a caller passing explicit `shell: false` is preserved. Correct.

### Test status
`node tests/bug-2598-windows-npm-spawn-shell.test.cjs` — **5 pass / 0 fail** (all four structural assertions plus the new diagnostics assertion).

No outstanding CodeRabbit items. No new commits required.

PR #2614, opened 2026-04-23
fix(#2601): include 'inherit' in VALID_PROFILES and unify allowlists

### LINKED ISSUE #2601: config-set-model-profile rejects 'inherit' despite runtime support
### What happens

Running `gsd-sdk query config-set-model-profile inherit` fails with:

```
Error: Invalid profile 'inherit'. Valid profiles: quality, balanced, budget, adaptive
```

The `inherit` profile is a first-class feature — it tells GSD to skip the `MODEL_PROFILES` map and let subagents inherit the session model instead of using per-role assignments. The runtime getter in `resolveModel` (`config-query.ts`) handles it correctly, and the health validator in `validate.ts` lists it as valid. But the setter in `configSetModelProfile` (`config-mutation.ts`) rejects it because `VALID_PROFILES` is derived from `Object.keys(MODEL_PROFILES['gsd-planner'])`, which only contains `quality`, `balanced`, `budget`, `adaptive` — `inherit` is not and cannot be a key in `MODEL_PROFILES` since its whole purpose is to bypass that map.

The workaround is to manually edit `.planning/config.json` and set `"model_profile": "inherit"` by hand, which the runtime reads fine. But the CLI path that users actually invoke is broken.

### How to trigger

1. Run any GSD project with the SDK installed
2. `gsd-sdk query config-set-model-profile inherit --raw`
3. Or use the `/gsd-set-profile inherit` slash command (which calls the same SDK handler)

Expected: profile is written to `config.json`, agents inherit session model.
Actual: `GSDError` with `ErrorClassification.Validation`, exit code 10.

Setting any other profile (`quality`, `balanced`, `budget`, `adaptive`) works fine via the same path.

### Root cause — three allowlists that disagree

The SDK has three independent definitions of "valid profiles" that have drifted out of sync:

| Source | File | Values | `inherit`? | `adaptive`? |
|---|---|---|---|---|
| `VALID_PROFILES` | `sdk/src/query/config-query.ts` (line 44) | quality, balanced, budget, adaptive | **No** | Yes |
| Setter validation | `sdk/src/query/config-mutation.ts` (imports `VALID_PROFILES`) | quality, balanced, budget, adaptive | **No** | Yes |
| Health validator | `sdk/src/query/validate.ts` (line 373) | quality, balanced, budget, inherit | **Yes** | **No** |
| Runtime getter | `sdk/src/query/config-query.ts` `resolveModel` (line 171) | special-cases `inherit` | **Yes** | N/A |

The getter knows about `inherit`. The health validator knows about `inherit` (but lost `adaptive`). The setter — the only path users actually call to change profiles — rejects it.

### Suggested fix direction

The core issue is that `VALID_PROFILES` is derived purely from `MODEL_PROFILES` keys, but `inherit` is a meta-profile that intentionally lives outside that map. Two possible approaches:

1. **Append `inherit` to `VALID_PROFILES`** after deriving from `MODEL_PROFILES` keys — e.g. `[...Object.keys(MODEL_PROFILES['gsd-planner']), 'inherit']`. Simple, keeps the single source of truth.
2. **Special-case `inherit` in `configSetModelProfile`** before the `VALID_PROFILES.includes()` check, similar to how `resolveModel` already special-cases it.

Either way, the hardcoded list in `validate.ts` (line 373) should be replaced with the shared `VALID_PROFILES` constant so the three definitions collapse to one and cannot drift again.

### Environment

- GSD SDK version: 0.1.0 (current on npm)
- Confirmed on upstream `main` (current HEAD) — same code as v1.38.1+
- Runtime: Node.js 22
- OS: macOS (also reproducible on Linux)
- Affects all runtimes that use the SDK CLI path: Claude Code, OpenCode, Gemini CLI

ssmulders:
## Fix PR

---

## Linked Issue

Fixes #2601

---

## What was broken

`gsd-sdk query config-set-model-profile inherit` failed with "Invalid profile 'inherit'. Valid profiles: quality, balanced, budget, adaptive" — despite `inherit` being a supported meta-profile that the runtime getter and health validator both accept.

## What this fix does

Adds `'inherit'` to the shared `VALID_PROFILES` constant and replaces the hardcoded allowlist in `validate.ts` with that same constant, so all three consumers (setter, getter, health validator) agree on valid profiles.

## Root cause

Three independent allowlists had drifted out of sync:

| Source | File | Has `inherit`? | Has `adaptive`? |
|---|---|---|---|
| `VALID_PROFILES` | `config-query.ts:54` | **No** | Yes |
| Health validator | `validate.ts:468` | **Yes** | **No** |
| Runtime getter | `config-query.ts:171` | **Yes** (special-cased) | N/A |

`VALID_PROFILES` was derived from `Object.keys(MODEL_PROFILES['gsd-planner'])`, and `inherit` cannot be a key there since its purpose is to bypass the map entirely. The health validator had a separate hardcoded list that included `inherit` but omitted `adaptive`.

### How I verified the fix

1. Ran `npx vitest run` on config-query, config-mutation, and validate test files — 82 tests pass
2. New regression test `'accepts inherit profile'` in `config-mutation.test.ts` calls `configSetModelProfile(['inherit'], ...)` and verifies it writes to `config.json` successfully

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None

## Summary by CodeRabbit

* **Bug Fixes**
  * The `inherit` model profile is now accepted during configuration updates and validated consistently across health checks.

* **Tests**
  * New tests verify `inherit` is recognized as valid, returned by queries, and persisted correctly.

* **Documentation**
  * Changelog updated to reflect the profile validation and behavior changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ If you use GSD **as a workflow**—milestones, phases, `.planning/` artifacts, b
 
 ### Fixed
 
+- **`config-set-model-profile` rejected `inherit` despite runtime support** — `VALID_PROFILES` was derived solely from `MODEL_PROFILES` keys, which cannot include `inherit` (a meta-profile that bypasses the map). The health validator had its own hardcoded list that included `inherit` but omitted `adaptive`. Unified all three allowlists into a single `VALID_PROFILES` constant that appends `inherit` after the `MODEL_PROFILES`-derived entries (#2601)
 - **Shell hooks falsely flagged as stale on every session** — `gsd-phase-boundary.sh`, `gsd-session-state.sh`, and `gsd-validate-commit.sh` now ship with a `# gsd-hook-version: {{GSD_VERSION}}` header; the installer substitutes `{{GSD_VERSION}}` in `.sh` hooks the same way it does for `.js` hooks; and the stale-hook detector in `gsd-check-update.js` now matches bash `#` comment syntax in addition to JS `//` syntax. All three changes are required together — neither the regex fix alone nor the install fix alone is sufficient to resolve the false positive (#2136, #2206, #2209, #2210, #2212)
 
 ### Added
--- a/sdk/src/query/config-mutation.test.ts
+++ b/sdk/src/query/config-mutation.test.ts
@@ -295,6 +295,20 @@ describe('configSetModelProfile', () => {
     await expect(configSetModelProfile(['invalid_profile'], tmpDir)).rejects.toThrow(GSDError);
   });
 
+  it('accepts inherit profile', async () => {
+    const { configSetModelProfile } = await import('./config-mutation.js');
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({ model_profile: 'balanced' }),
+    );
+    const result = await configSetModelProfile(['inherit'], tmpDir);
+    expect((result.data as { updated: boolean }).updated).toBe(true);
+    expect((result.data as { profile: string }).profile).toBe('inherit');
+
+    const raw = JSON.parse(await readFile(join(tmpDir, '.planning', 'config.json'), 'utf-8'));
+    expect(raw.model_profile).toBe('inherit');
+  });
+
   it('normalizes profile name to lowercase', async () => {
     const { configSetModelProfile } = await import('./config-mutation.js');
     await writeFile(
--- a/sdk/src/query/config-query.test.ts
+++ b/sdk/src/query/config-query.test.ts
@@ -194,3 +194,25 @@ describe('VALID_PROFILES', () => {
     expect(VALID_PROFILES).toEqual(['quality', 'balanced', 'budget', 'adaptive', 'inherit']);
   });
 });
+
+// ─── getAgentToModelMapForProfile ───────────────────────────────────────────
+
+describe('getAgentToModelMapForProfile', () => {
+  it('returns inherit for all agents when profile is inherit', async () => {
+    const { getAgentToModelMapForProfile } = await import('./config-query.js');
+    const map = getAgentToModelMapForProfile('inherit');
+    expect(Object.keys(map).length).toBeGreaterThan(0);
+    for (const model of Object.values(map)) {
+      expect(model).toBe('inherit');
+    }
+  });
+
+  it('returns concrete models for non-inherit profiles', async () => {
+    const { getAgentToModelMapForProfile } = await import('./config-query.js');
+    const map = getAgentToModelMapForProfile('balanced');
+    for (const model of Object.values(map)) {
+      expect(model).not.toBe('inherit');
+      expect(['opus', 'sonnet', 'haiku']).toContain(model);
+    }
+  });
+});
--- a/sdk/src/query/config-query.test.ts
+++ b/sdk/src/query/config-query.test.ts
@@ -189,8 +189,8 @@ describe('MODEL_PROFILES', () => {
 // ─── VALID_PROFILES ─────────────────────────────────────────────────────────
 
 describe('VALID_PROFILES', () => {
-  it('contains the four profile names', async () => {
+  it('contains the five profile names including inherit', async () => {
     const { VALID_PROFILES } = await import('./config-query.js');
-    expect(VALID_PROFILES).toEqual(['quality', 'balanced', 'budget', 'adaptive']);
+    expect(VALID_PROFILES).toEqual(['quality', 'balanced', 'budget', 'adaptive', 'inherit']);
   });
 });
--- a/sdk/src/query/config-query.ts
+++ b/sdk/src/query/config-query.ts
@@ -57,6 +57,13 @@ export const VALID_PROFILES: string[] = [...Object.keys(MODEL_PROFILES['gsd-plan
  * Flat map of agent name → model alias for one profile tier (matches `model-profiles.cjs`).
  */
 export function getAgentToModelMapForProfile(normalizedProfile: string): Record<string, string> {
+  if (normalizedProfile === 'inherit') {
+    const agentToModelMap: Record<string, string> = {};
+    for (const agent of Object.keys(MODEL_PROFILES)) {
+      agentToModelMap[agent] = 'inherit';
+    }
+    return agentToModelMap;
+  }
   const profile = VALID_PROFILES.includes(normalizedProfile) ? normalizedProfile : 'balanced';
   const agentToModelMap: Record<string, string> = {};
   for (const [agent, profileToModelMap] of Object.entries(MODEL_PROFILES)) {
--- a/sdk/src/query/config-query.ts
+++ b/sdk/src/query/config-query.ts
@@ -50,8 +50,8 @@ export const MODEL_PROFILES: Record<string, Record<string, string>> = {
   'gsd-doc-verifier': { quality: 'sonnet', balanced: 'sonnet', budget: 'haiku', adaptive: 'haiku' },
 };
 
-/** Valid model profile names. */
-export const VALID_PROFILES: string[] = Object.keys(MODEL_PROFILES['gsd-planner']);
+/** Valid model profile names (includes 'inherit' which bypasses MODEL_PROFILES). */
+export const VALID_PROFILES: string[] = [...Object.keys(MODEL_PROFILES['gsd-planner']), 'inherit'];
 
 /**
  * Flat map of agent name → model alias for one profile tier (matches `model-profiles.cjs`).
--- a/sdk/src/query/validate.ts
+++ b/sdk/src/query/validate.ts
@@ -465,8 +465,11 @@ export const validateHealth: QueryHandler = async (args, projectDir, _workstream
     try {
       const raw = await readFile(configPath, 'utf-8');
       const parsed = JSON.parse(raw) as Record<string, unknown>;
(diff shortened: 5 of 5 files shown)


trek-e (changes requested):
Review — one semantic defect, otherwise the fix is sound.

**Finding — `getAgentToModelMapForProfile` does not handle `inherit`:**

`sdk/src/query/config-query.ts:59-67`

```ts
export function getAgentToModelMapForProfile(normalizedProfile: string): Record<string, string> {
  const profile = VALID_PROFILES.includes(normalizedProfile) ? normalizedProfile : 'balanced';
  const agentToModelMap: Record<string, string> = {};
  for (const [agent, profileToModelMap] of Object.entries(MODEL_PROFILES)) {
    const mapped = profileToModelMap[profile] ?? profileToModelMap.balanced;
    agentToModelMap[agent] = mapped ?? 'sonnet';
  }
  return agentToModelMap;
}
```

After this PR, `VALID_PROFILES.includes('inherit')` is `true`, so `profile = 'inherit'`, then `profileToModelMap['inherit']` is always `undefined` (MODEL_PROFILES never had an `inherit` column), so every lookup falls through to `profileToModelMap.balanced`. Callers of `configSetModelProfile` with `inherit` get back `agentToModelMap` populated as if the user chose `balanced`.

This contradicts `inherit` semantics as implemented by `resolveModel` (sdk/src/query/config-query.ts:186-188), which correctly special-cases `inherit` and returns `{model: 'inherit'}`.

Callsite that exposes the defect: `sdk/src/query/config-mutation.ts:329` — `configSetModelProfile` returns `agentToModelMap` in its response. A user running `gsd-sdk query config-set-model-profile inherit --raw` sees `profile: 'inherit'` but `agentToModelMap: {gsd-planner: 'opus', ...}`, which is misleading and would produce wrong behavior if any caller trusts that map instead of re-checking the profile.

**Required fix:** either

(a) early-return an empty map (or `{agent: 'inherit' for each agent}`) from `getAgentToModelMapForProfile` when `normalizedProfile === 'inherit'`, mirroring the `resolveModel` special-case; or

(b) have `configSetModelProfile` omit `agentToModelMap` from the response payload when the profile is `inherit`.

(a) is safer — keeps the function shape consistent and lets every caller handle inherit uniformly. Add a regression test asserting the returned map is empty / sentinel when `inherit` is passed.

Rest of the diff (the `VALID_PROFILES` consolidation in `config-query.ts`, the `validate.ts` replacement of the hardcoded list, the CHANGELOG entry, the existing new test) all look correct.

trek-e (changes requested):
**Verdict: Request changes.** PR is `CONFLICTING` — must rebase. Beyond that, the "unification" claim overstates what the diff delivers.

### RCA on the underlying issue (#2601)
The cause is not "`inherit` missing from an allowlist." The cause is **`VALID_PROFILES` was *derived* (`Object.keys(MODEL_PROFILES['gsd-planner'])`) rather than *declared***. Derived-from-data-shape constants silently lie when the domain has members that do not fit the shape. `inherit` is a meta-profile — categorically cannot be a key in `MODEL_PROFILES`. This is a type-modeling defect, not an allowlist typo.

### Blame
- `002bcf2a` (Tom Boucher, 2026-04-12) introduced `VALID_PROFILES = Object.keys(...)` at [sdk/src/query/config-query.ts:54](sdk/src/query/config-query.ts:54) *and* the separately-hardcoded list in `validate.ts`. Both landed in the same commit and disagreed from day one.
- `c5b14455` (Rezolv, 2026-04-20) added `getAgentToModelMapForProfile`, inheriting the drift.

### Findings

**A — Goodhart (test gaming).** The test `VALID_PROFILES contains the five profile names including inherit` was rewritten from the four-name assertion to match the fix. It was not authored to catch the bug. The cause-level invariant test — "no source file under `sdk/src/query/` hardcodes a profile-name literal outside `config-query.ts`" — does not exist. Add it. Without that invariant, the three lists can re-drift.

**B — Gall's Law.** The simple form (`Object.keys(...)`) worked until it didn't. The PR reintroduces the same fragility: `[...Object.keys(MODEL_PROFILES['gsd-planner']), 'inherit']`. When the next meta-profile (`auto`, `dynamic`, `off`) appears, the same bug recurs. Model the two categories explicitly:
\`\`\`ts
const CONCRETE_PROFILES = Object.keys(MODEL_PROFILES['gsd-planner']);
const META_PROFILES = ['inherit'] as const;
export const VALID_PROFILES = [...CONCRETE_PROFILES, ...META_PROFILES];
\`\`\`

**C — New drift introduced.** [sdk/src/query/validate.ts:468](sdk/src/query/validate.ts:468) now applies `String(parsed.model_profile).trim().toLowerCase()` before the check. The setter path does not. Two normalizers, one constant — the *same drift pattern* this PR claims to eliminate, reintroduced on a different axis. Extract a `normalizeProfile()` helper and use it in both call sites.

**D — Peter Principle.** `getAgentToModelMapForProfile('inherit')` now returns `{agent: 'inherit', ...}`. The annotated return type is `Record<string, string>` — "agent name → model alias." `'inherit'` is not a model alias. Any downstream consumer that passes values from this map to a model resolver will fail at a harder-to-diagnose layer. The fix has *shifted* the failure site, not removed it. Audit callers at [sdk/src/query/config-mutation.ts:329](sdk/src/query/config-mutation.ts:329) and document the sentinel with JSDoc on the function.

**E — Kernighan / Knuth.** No violations.

### Required before merge

1. Rebase on `main`.
2. Split `VALID_PROFILES` into `CONCRETE_PROFILES` + `META_PROFILES` (Finding B).
3. Extract `normalizeProfile()` and use it in both `configSetModelProfile` and `validateHealth` (Finding C).
4. Add an invariant test preventing hardcoded profile-name literals outside `config-query.ts` (Finding A).
5. Audit callers of the returned `agentToModelMap` for `'inherit'` sentinel handling; JSDoc the sentinel (Finding D).
6. Rewrite the CHANGELOG entry to match what the code actually does — current wording overclaims.

trek-e (commented):
## Code Review: fix(#2601): include 'inherit' in VALID_PROFILES and unify allowlists

### 1. `VALID_PROFILES` is derived from `MODEL_PROFILES['gsd-planner']` keys — 'inherit' is not in there

Looking at `sdk/src/query/config-query.ts` line 54: `export const VALID_PROFILES: string[] = Object.keys(MODEL_PROFILES['gsd-planner'])`. The keys of `gsd-planner` are `['quality', 'balanced', 'budget', 'adaptive']`. 'inherit' is not a key in MODEL_PROFILES. If this PR adds 'inherit' to VALID_PROFILES by modifying how it's derived, show the exact change. If it appends 'inherit' after the Object.keys() call, that's a different (and more fragile) approach than a data-driven derivation.

### 2. `getAgentToModelMapForProfile('inherit')` silently falls back to 'balanced'

Line 60 of `config-query.ts`: `const profile = VALID_PROFILES.includes(normalizedProfile) ? normalizedProfile : 'balanced'`. After this PR, 'inherit' is in VALID_PROFILES, so the ternary takes the left branch. But 'inherit' is not a key in any `MODEL_PROFILES[agent]` object. Line 63: `const mapped = profileToModelMap[profile] ?? profileToModelMap.balanced`. So `profileToModelMap['inherit']` is `undefined`, and the fallback to `profileToModelMap.balanced` fires. Every agent gets 'balanced' model aliases — not 'inherit'. The PR description says `getAgentToModelMapForProfile('inherit')` returns `{ agent: 'inherit' }` for all agents. That is not what the code does. Either the test is wrong, the description is wrong, or the implementation is different from what's on main.

### 3. What does the executor do when it receives `model: 'inherit'`?

`resolveModel` for a single agent (lines 186–188 of config-query.ts) has a special branch: `if (profile === 'inherit') return { data: { model: 'inherit', profile } }`. So `resolve-model gsd-executor` with profile 'inherit' returns `model: 'inherit'` as a string. What does the executor do with that string? If it passes 'inherit' directly to the Anthropic API as a model ID, you've created a new failure mode: API call fails with "unknown model 'inherit'". There is no test for what happens when 'inherit' reaches a model invocation. Either test the full chain to the API boundary, or document that 'inherit' is expected to be intercepted at the executor level (and show where that interception happens).

### 4. The `adaptive` profile fix is real and unacknowledged

The old health validator hardcoded `['quality', 'balanced', 'budget', 'inherit']` — missing `adaptive`. The new VALID_PROFILES derives from MODEL_PROFILES keys, which include `adaptive`. This means the health check previously warned about valid `adaptive` configurations as invalid. That is a real user-facing bug fix that is not mentioned in the changelog entry or PR description. It also means 'adaptive' was silently broken in health validation for however long it's been in MODEL_PROFILES. Acknowledge the fix, add a test for it, and explain why it was omitted from the allowlist originally.

### 5. There are no downstream tests for 'inherit' at execution time

The test "getAgentToModelMapForProfile('inherit') returns { ...: 'inherit' }" (which, as noted above, may not match actual behavior) is a unit test for one function. The integration path — user sets `model_profile: inherit` in config → `resolve-model gsd-executor` → executor uses model string — is untested. Until that path is tested end-to-end, shipping 'inherit' as a valid profile is shipping an unverified code path.

VERDICT: NEEDS DISCUSSION — clarify what `getAgentToModelMapForProfile('inherit')` actually returns vs. what the PR claims, and trace the 'inherit' string to the API boundary before merge.

trek-e:

PR #2625, opened 2026-04-23
fix(#2620): detect HOME-relative PATH entries before suggesting absolute export

### LINKED ISSUE #2620: Installer suggests absolute path but user has relative HOME/.npm-global/bin in PATH
### Describe the bug

When installing via `npx get-shit-done-cc@latest`, the installer reports `gsd-sdk` not on PATH and suggests adding the absolute path:

```
echo 'export PATH="/home/corrm/.npm-global/bin:$PATH"' >> /home/corrm/.zshrc
```

But im already have the directory in the PATH using `$HOME`:

```
export PATH="$HOME/.npm-global/bin:$PATH"
```

The installer should recognize the relative path pattern and not suggest adding a duplicate entry.

### Steps to Reproduce

1. User has `export PATH="$HOME/.npm-global/bin:$PATH"` in their `.zshrc`
2. Run `npx get-shit-done-cc@latest`
3. Installer detects `gsd-sdk` not on PATH and suggests adding absolute path
4. This creates a duplicate PATH entry

### Expected behavior

Installer should check both absolute and `$HOME`-based PATH patterns before suggesting a fix.

### Actual behavior

Installer only checks for the exact absolute path, not the `\$HOME`-based pattern.

### Your environment

- OS: Linux (ArchLinux)
- Shell: zsh
- GSD version: latest via npx

### Scope of changes

- `get-shit-done/bin/install.js` — modify PATH detection logic to check `$HOME/.npm-global/bin` pattern in addition to absolute path

### Test plan

- Regression test: install with `$HOME/.npm-global/bin` in PATH → should detect existing entry
- Test: install with absolute path in PATH → should detect existing entry
- Test: install with neither → should suggest correct PATH

---

**Labels:** bug

trek-e:
## Summary

Closes #2620.

When the installer reported `gsd-sdk` not on PATH it suggested appending
`export PATH="/home/user/.npm-global/bin:$PATH"` to the user's rc file. A user
who had the equivalent `export PATH=\"\$HOME/.npm-global/bin:\$PATH\"` already
in `~/.zshrc` would get a duplicate entry — the installer only compared the
absolute path.

## Detection algorithm (`homePathCoveredByRc` in `bin/install.js`)

1. Iterate the common rc files in `homeDir`: `.zshrc`, `.bashrc`,
   `.bash_profile`, `.profile` (overridable via the third arg).
2. Read each file; silently skip missing / unreadable / directory entries so
   the fallback is always the existing absolute-path suggestion.
3. For every non-comment line that matches `^\\s*(?:export\\s+)?PATH\\s*=(.+)$`,
   strip one layer of surrounding quotes from the RHS.
4. Split the RHS on `:` and for each segment substitute the common HOME forms
   — `${HOME}`, `$HOME`, and a leading `~/` (or bare `~`) — with the resolved
   `homeDir`. Segments still containing `$` are skipped (can't resolve without
   a real shell).
5. Normalise trailing slashes on the expanded segment and compare against
   `path.resolve(globalBin)`. Any match → return `true`.
6. If no rc file covers `globalBin`, return `false` and let the caller emit
   the existing absolute export suggestion.

Only `\$HOME/…`, `\${HOME}/…`, and `~/…` are handled — we deliberately don't
try to fully parse bash syntax. Every read/resolve is wrapped in try/catch so
a weird rc file can never crash the installer.

## Test plan

- [x] `tests/install-path-detection.test.cjs` covers `\$HOME`, `\${HOME}`, `~/`,
  absolute-match, trailing-slash, commented-out, missing rc, unreadable rc
  (directory where a file is expected), and non-matching segments.
- [x] `node scripts/run-tests.cjs` — 5420 tests pass, 0 fail.

## Release Notes

* **New Features**
  * Installer now accurately detects when the npm global bin directory is already configured in a user's shell startup files, supporting various path notation formats and preventing duplicate entries.

* **Tests**
  * Added regression tests for PATH detection behavior across different shell configurations and edge cases.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -6798,6 +6798,94 @@ function promptLocation(runtimes) {
 }
 
 /**
+ * Check whether any common shell rc file already contains a `PATH=` line
+ * whose HOME-expanded value places `globalBin` on PATH (#2620).
+ *
+ * Parses `~/.zshrc`, `~/.bashrc`, `~/.bash_profile`, `~/.profile` (or the
+ * override list in `rcFileNames`), matches `export PATH=` / bare `PATH=`
+ * lines, and substitutes the common HOME forms (`$HOME`, `${HOME}`, `~`)
+ * with `homeDir` before comparing each PATH segment against `globalBin`.
+ *
+ * Best-effort: any unreadable / malformed / non-existent rc file is ignored
+ * and the fallback is the caller's existing absolute-path suggestion. Only
+ * the `$HOME/…`, `${HOME}/…`, and `~/…` forms are handled — we do not try
+ * to fully parse bash syntax.
+ *
+ * @param {string} globalBin  Absolute path to npm's global bin directory.
+ * @param {string} homeDir    Absolute path used to substitute HOME / ~.
+ * @param {string[]} [rcFileNames]  Override the default rc file list.
+ * @returns {boolean}         true iff any rc file adds globalBin to PATH.
+ */
+function homePathCoveredByRc(globalBin, homeDir, rcFileNames) {
+  if (!globalBin || !homeDir) return false;
+  const path = require('path');
+  const fs = require('fs');
+
+  const normalise = (p) => {
+    if (!p) return '';
+    let n = p.replace(/[\\/]+$/g, '');
+    if (n === '') n = p.startsWith('/') ? '/' : p;
+    return n;
+  };
+
+  const targetAbs = normalise(path.resolve(globalBin));
+  const homeAbs = path.resolve(homeDir);
+  const files = rcFileNames || ['.zshrc', '.bashrc', '.bash_profile', '.profile'];
+
+  const expandHome = (segment) => {
+    let s = segment;
+    s = s.replace(/\$\{HOME\}/g, homeAbs);
+    s = s.replace(/\$HOME/g, homeAbs);
+    if (s.startsWith('~/') || s === '~') {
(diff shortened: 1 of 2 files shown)


trek-e:
@coderabbitai thanks for the review — both actionable findings and the nitpick are now fixed.

**Commit 46f8e22 — relative PATH segments** (Actionable 2)
Added a `path.isAbsolute(expanded)` guard after HOME expansion in `homePathCoveredByRc`. Bare relative segments (e.g. `bin`, `node_modules/.bin`) are now skipped instead of being resolved against `homeAbs` and producing false positives. Covered by two new regression tests in `tests/install-path-detection.test.cjs`.

**Commit 5e17ee0 — wire helper into installer** (Actionable 1 + nitpick)
Added `maybeSuggestPathExport(globalBin, homeDir)` and call it from `installSdkIfNeeded` after the sdk/dist check. It:
1. Skips silently if `globalBin` is already on `process.env.PATH`.
2. Prints a "directory already on PATH via rc file — try reopening your shell" diagnostic when `homePathCoveredByRc` returns true.
3. Otherwise emits the absolute `echo 'export PATH="…:$PATH"' >> ~/.zshrc` suggestion.

`globalBin` is resolved via `npm prefix -g` (plus `/bin` on POSIX); exec failures are swallowed.

Three installer-flow regression tests added per the nitpick: rc-covers-via-$HOME → no suggestion, rc-covers-only-unrelated → suggestion emitted, already-on-PATH → no output. Full suite (5425 tests) green.

PR #2622, opened 2026-04-23
fix(#2613): preserve STATE.md frontmatter on write path (option 2)

trek-e:
## Summary

Closes #2613

Fixes the write-side analogue of #2495. `readModifyWriteStateMd` strips STATE.md frontmatter before invoking the modifier, so `syncStateFrontmatter` received body-only content and `existingFm` was always `{}` — the preservation branch at `state-mutation.ts:243` never fired. Every mutation re-derived `status` (to `'unknown'` when body had no `Status:` line) and `progress.*` (to 0/0 when the shipped milestone's phase directories were archived), silently overwriting authoritative frontmatter values.

## Fix (option 2 per escalation discussion on the issue)

Write-side analogue of #2495's READ fix: `buildStateFrontmatter` now reads the current STATE.md frontmatter from disk as a preservation backstop.

- **Status:** preserved when derived is `'unknown'` and existing is non-unknown.
- **Progress:** preserved when disk scan returns all zeros AND existing has non-zero counts.
- **Milestone / milestone_name:** already preserved via `getMilestoneInfo`'s #2495 fix — regression test added to lock it in.

Legitimate body-driven status changes (`state.update Status executing`) still win, and non-zero disk counts still win over stale frontmatter counts (mid-milestone mutations keep ground-truth progress).

## Root cause

\`\`\`ts
// state-mutation.ts:260–286 (before fix)
const body = stripFrontmatter(content);         // frontmatter gone
const modified = await modifier(body);
const synced = await syncStateFrontmatter(modified, projectDir);
//                                        ^ body-only content
// → inside syncStateFrontmatter:
const existingFm = extractFrontmatter(content); // always {} — nothing to preserve
\`\`\`

## Tests

5 new regression tests in `sdk/src/query/state-mutation.test.ts`:

1. `record-session preserves milestone + milestone_name when ROADMAP has a different current milestone` — locks in the #2495 behavior.
2. `record-session preserves status from existing frontmatter when body has no Status field` — direct repro of the reporter's `status: shipped → executing` regression.
3. `record-session preserves progress from frontmatter when disk scan returns zero counts` — direct repro of the reporter's `progress: 12/12 100% → 0/0 0%` regression.
4. `regression guard: state.update Status still updates frontmatter status when body is mutated` — ensures legitimate status transitions still propagate.
5. `regression guard: disk-scanned progress wins when scan returns non-zero counts` — ensures mid-milestone mutations keep disk as ground truth.

## Verification

[...]

PR #2611, opened 2026-04-23
feat: add gates ensuring discuss-phase decisions are translated to plans and verified (closes #2492)

trek-e:
## TL;DR

Two gates close the loop between discuss-phase decisions and downstream work, so a decision captured in CONTEXT.md `<decisions>` cannot silently disappear during planning or execution.

Closes #2492

## What

- **Plan-phase translation gate (BLOCKING).** Inserted as Step 13a in `workflows/plan-phase.md`, after the existing requirements coverage gate and before plans are committed. Each trackable decision in `<decisions>` must be cited (by id `D-NN` or by 6+-word phrase) in at least one plan's **designated sections** — front-matter `must_haves` / `truths` / `objective`, or body sections under `must_haves` / `truths` / `tasks` / `objective` headings. HTML comments and fenced code blocks are stripped before searching, so a commented-out citation or a literal example never counts as coverage. The shell snippet wraps the SDK call with `jq -e '.data.passed == true' || exit 1` so a coverage gap actually halts the workflow rather than only printing a warning.
- **Verify-phase validation gate (NON-BLOCKING).** Inserted as a `<step name="verify_decisions">` in `workflows/verify-phase.md`, before `behavioral_verification`. Each trackable decision is searched against shipped artifacts (PLAN.md, SUMMARY.md, files modified, recent commit messages — last 200). Misses are reported in VERIFICATION.md as a warning section but do **not** flip the overall verification status.
- **Shared parser** `parseDecisions()` in `sdk/src/query/decisions.ts` — exposed as `decisions.parse` query handler — so #2493 (post-planning gap checker) can consume the same logic without duplicating decision-parsing code.
- **Config key** `workflow.context_coverage_gate` (boolean, default `true`). Added to `WorkflowConfig`, `CONFIG_DEFAULTS`, `VALID_CONFIG_KEYS`, and `checkConfigGates`. When `false`, both gates skip silently. Workstream-scoped configs are honored — the handlers thread `workstream` through `loadConfig` so per-workstream overrides take effect.

## Why

The issue describes a real-world failure: discuss-phase captured ~40 decisions, plans were created without implementing several of them, and verification passed because it only checked plan objectives. Comments and workarounds don't fix this; instrumentation does.

## Why the asymmetry (plan blocks, verify warns)

- Plan time is cheap. Failing at plan time costs minutes — the user re-cites the decision id in a plan, or moves the decision to Discretion.
[...]

PR #2631, opened 2026-04-23
ci: explicit rebase check + fail-fast SDK typecheck in install-smoke

### LINKED ISSUE #2632: CI: install-smoke misattributes stale-base SDK build failures as installSdkIfNeeded regression
## Bug

When a PR's base is behind main and main has since landed a type fix (common when a regression lands and is patched within hours), the smoke CI fails with:

> ##[error]gsd-sdk is not on PATH after install — installSdkIfNeeded() regression

This message is misleading — the actual cause is \`npm run build\` failing in sdk/ with a TypeScript error that was already fixed on current main.

## Root cause

GitHub's \`refs/pull/N/merge\` is cached against the PR's recorded merge-base, not current main. When main advances after a PR is opened, the cache stays stale — CI runs against the pre-advance tree. Concrete example hit today: #2611 introduced a \`WorkflowConfig\` → \`Record<string, unknown>\` cast incompatible with \`strict\` mode. #2622 patched it with a double-cast. Every PR opened between those two commits (and not rebased since) replayed the broken intermediate state and failed smoke with the "installSdkIfNeeded regression" message.

Observed on PR #2616 — the CI fix itself was trivial (rebase), but the diagnostic pointed in the wrong direction, costing time.

## Expected

- CI should run against PR HEAD merged with **current** main, not a cached merge-base.
- When a build fails due to a stale base, the error message should say so with an actionable remediation (\`git fetch origin main && git rebase origin/main && git push --force-with-lease\`).
- Installer failures should surface install.js's own error output, not be masked by a downstream PATH assertion.

## Fix

- \`.github/workflows/install-smoke.yml\`: explicit \`git merge origin/main\` step before build; dedicated \`npm run build:sdk\` typecheck step with clear remediation message; drop \`|| true\` on \`get-shit-done-cc --claude --local\` so installer failures surface cleanly.
- \`.github/workflows/test.yml\`: same rebase-check pattern since test.yml runs on every PR.

Fix in [PR #2631](https://github.com/gsd-build/get-shit-done/pull/2631).

trek-e:
Closes #2632

## Summary

Harden CI against the stale-merge-ref failure mode that caused PR #2616's smoke CI to report "installSdkIfNeeded() regression" when the real cause was a TypeScript error that was already fixed on main.

## Root cause

GitHub's `refs/pull/N/merge` is cached against the PR's recorded merge-base, not current main. When main advances after a PR is opened, the merge ref stays stale — CI runs against the pre-advance tree. This hides fixes that have already landed on trunk and replays intermediate broken states in stale PRs.

Observed in the wild today: #2611 introduced a `WorkflowConfig` → `Record<string, unknown>` cast that trips `strict` mode. #2622 patched it with a double-cast shortly after. Every PR opened between those two commits (and not rebased since) runs its smoke CI against the broken intermediate state and fails with a misleading "gsd-sdk not on PATH — installSdkIfNeeded regression" message. PR #2616 hit this for hours.

## Changes

**`.github/workflows/install-smoke.yml`:**
- Explicit `git merge origin/main` step with a clear "rebase onto main" diagnostic on conflict. Catches stale bases before any build runs.
- Dedicated SDK-typecheck step (`npm run build:sdk`) that fails fast with actionable guidance: "rebase onto main — the error may already be fixed on trunk." Replaces the silent build-then-blame-the-assertion flow.
- Dropped `|| true` on `get-shit-done-cc --claude --local` so installer failures surface at the install step (where install.js's own error is visible), not at the downstream PATH assertion where the message misleadingly reads "shim regression."
- `fetch-depth: 0` on checkout so the merge-base check has history.

**`.github/workflows/test.yml`:**
- Same rebase-check step — test suite runs on every PR (no path filter), so catching staleness here surfaces the issue even before smoke triggers.

## Why this matters long-term

Any time a type or contract regression lands on main and is patched within hours, every PR opened before the patch will fail CI with a cryptic error until it rebases. The new rebase-check gives contributors a clear message instead of sending them hunting through install.js internals.

## Test plan

- [x] YAML parses cleanly (`python3 -c "import yaml; yaml.safe_load(open(...))"`)
- [ ] CI passes on this PR itself
- [ ] Re-trigger on a known stale PR to confirm the new error message is visible

## Summary by CodeRabbit

* **Chores**
  * Enhanced internal CI/CD workflows to validate code changes against the latest main branch before testing and installation, with improved error diagnostics for merge conflicts.

### DIFF
--- a/.github/workflows/install-smoke.yml
+++ b/.github/workflows/install-smoke.yml
@@ -102,7 +102,7 @@ jobs:
           set -euo pipefail
           git config user.email "ci@gsd-build"
           git config user.name "CI Rebase Check"
-          git fetch origin main --depth=0 || git fetch origin main
+          git fetch origin main
           if ! git merge --no-edit --no-ff origin/main; then
             echo "::error::This PR cannot cleanly merge origin/main. Rebase your branch onto current main and push again."
             echo "::error::Conflicting files:"
@@ -216,6 +216,26 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
         with:
           ref: ${{ inputs.ref || github.ref }}
+          fetch-depth: 0
+
+      # See the `smoke` job above for rationale — refs/pull/N/merge is cached
+      # against the recorded merge-base, not current main. Explicitly merge
+      # origin/main so smoke-unpacked also runs against the latest trunk.
+      - name: Rebase check — merge origin/main into PR head
+        if: github.event_name == 'pull_request'
+        shell: bash
+        run: |
+          set -euo pipefail
+          git config user.email "ci@gsd-build"
+          git config user.name "CI Rebase Check"
+          git fetch origin main
+          if ! git merge --no-edit --no-ff origin/main; then
+            echo "::error::This PR cannot cleanly merge origin/main. Rebase your branch onto current main and push again."
+            echo "::error::Conflicting files:"
+            git diff --name-only --diff-filter=U
+            git merge --abort
+            exit 1
+          fi
 
       - name: Set up Node.js 22
         uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f  # v6.3.0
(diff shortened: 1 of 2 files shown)


trek-e:
Addressed both CodeRabbit findings in fcc0d1d6:
- Added the rebase-check step to the `smoke-unpacked` job so it's not running against a stale cached merge either.
- Dropped the invalid `--depth=0` flag on `git fetch origin main` (git rejects it with 'depth 0 is not a positive number'). Plain `git fetch origin main` is what we actually want.

Issue #1020, opened 2026-03-12
[enhancement] codex hooks


Khang5687:
### Problem or motivation

codex 0.114.0 got [released](https://github.com/openai/codex/releases/tag/rust-v0.114.0) with `hooks`
> Added an experimental hooks engine with SessionStart and Stop hook events. 
Though it is experimental, and only has `SessionStart` and `Stop` hook events for now

I would love to have this used by gsd, or this issue opened until the hooks feature are stable.

### Proposed solution

Invoking gsd skills in codex cli or codex app, will use hooks just like in opencode and claude code.

You can enable hooks in codex by adding in `~/.codex/config.toml`
```
[features]
codex_hooks = true
```

An example of hook usage for codex by am.will on X: https://x.com/LLMJunky/status/2031582379291070849/photo/1

trek-e:
until the feature ships complete, it'd be really be best to not open a feature request for a beta/pre-release.   would suggest waiting until the feature is available and documented completely to be able to see what is able to be supported.  I'd suggest that @glittercowboy mark this as closed until the feature is available.

trek-e:
## Investigation

The current installer generates `config.toml` and agent `.toml` files for Codex but **skips hooks entirely** — the install function returns early before reaching the settings/hooks section.

Codex v0.114.0 added experimental `SessionStart` and `Stop` hook events. GSD already has hooks for:
- **SessionStart**: Version update checker (`gsd-update-check.js`)
- **PostToolUse/AfterTool**: Context monitor (`gsd-context-monitor.js`)
- **Statusline**: Context metrics display

To add Codex hooks support, the installer would need to:
1. Write hook configuration to `~/.codex/config.toml` under `[hooks]` section
2. Map GSD's `SessionStart` hook to Codex's equivalent format
3. Skip `PostToolUse` hooks until Codex adds that event type

Since Codex hooks are still experimental and only support `SessionStart`/`Stop`, the most useful integration right now would be the version update checker on `SessionStart`.

Keeping this issue open for tracking — will become more actionable when Codex adds `PostToolUse`-equivalent hook events.

trek-e:
PR #1138 adds Codex hooks support. The installer now enables `codex_hooks` feature flag and registers a SessionStart hook for update checking. PostToolUse hooks will follow when Codex adds that event type.

trek-e:
This issue is addressed by PR #1138. The fix is ready for review and merge.

PR #2640, opened 2026-04-23
Fix Codex hook installation to use hooks.json instead of inline TOML hooks

### LINKED ISSUE #2637: Codex 0.124.0 hook regression can leave GSD-installed Codex setups unable to start
## Summary

## Why this matters for GSD

## What I observed locally

```text
in `hooks`
```

3. reinstall Codex cleanly

After setting:

```toml
[features]
codex_hooks = false
```

## Expected behavior

## Related context

## Environment

- GSD runtime target: Codex
- Codex CLI: `0.124.0`
- Node: `v24.14.1`
- npm: `11.11.0`

## Proposed next step

hanzckernel:
## Fix PR

> Using the wrong template?
> — Enhancement: use enhancement template
> — Feature: use feature template

---

## Linked Issue

Fixes #2637

> The linked issue is still awaiting maintainer confirmation (`confirmed-bug` label not present yet).

---

## What was broken

GSD installed a Codex `SessionStart` update hook as inline `[[hooks]]` TOML in `config.toml`. Newer Codex versions expect hooks in `hooks.json`, so model-setting updates could fail with `invalid type: map, expected a sequence`. The install/uninstall path also had edge cases where managed hook references could survive in malformed or legacy configs.

## What this fix does

Moves the managed Codex update hook to `hooks.json`, preserves user-owned hook content, avoids rewriting malformed existing `hooks.json` files, tracks Codex hook files in the manifest, and removes only exact managed GSD hook references during uninstall across both `hooks.json` and legacy inline TOML hook blocks.

## Root cause

The installer treated Codex hook configuration like inline TOML hook arrays instead of Codex’s `hooks.json` format, and the cleanup logic matched only narrower legacy shapes than the real world configs users could carry forward.

### How I verified the fix

- Added/updated Codex regression coverage for:
  - `hooks.json` migration and merge behavior
  - malformed `hooks.json` preservation without rewrite
  - uninstall cleanup of exact managed hook references from malformed/non-`SessionStart`/inline TOML cases
  - compact TOML formatting and header-comment variants
  - root inline-table `features = { codex_hooks = true }` configs migrate to `hooks.json` without invalid dotted-key rewrites
  - Codex hook-file manifest tracking
- Ran:
  - `npm run build:hooks && node --test tests/codex-config.test.cjs`
  - `git diff --check`
  - `npm test`

### Regression test added?

- [x] Yes — added tests that would have caught the bug and the uninstall/migration edge cases
- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Codex

---

## Checklist

- [ ] Linked issue has the `confirmed-bug` label

## Breaking changes

None

## Summary by CodeRabbit

* **Changed**
  * Installer now records managed SessionStart hooks in a dedicated managed hooks file, merging into existing entries without overwriting user content, preserving malformed files when unsafe to migrate, preferring bundled distributable assets with fallback, improving legacy inline hook cleanup and matching, avoids deleting hook assets when cleanup fails, and refreshes the file manifest after install.
* **Tests**
  * Added end-to-end coverage for install/uninstall migration, user-hook/root-key preservation, malformed-file fallbacks, legacy escaping variants, and manifest tracking of installed hook assets.


trek-e (changes requested):
## Adversarial review — request changes

### Verdict

Request changes. The migration is well-tested and the edge-case handling (malformed `hooks.json`, zero-byte file, TOML literal apostrophes, managed-command cleanup across event names) is careful work. Two concrete problems before merge.

### RCA — one gap

**Missing version guard.** Issue #2637 explicitly asked for "detect affected Codex versions and skip enabling hooks" or equivalent mitigation. This PR unconditionally switches every Codex install to `hooks.json` (`bin/install.js:6606-6636`). Older Codex releases that only understood the inline `[[hooks]]` TOML are now silently unhooked — the `SessionStart` update check stops firing because the `hooks.json` format didn't exist yet. No runtime/version detection, no fallback. Gall's law: the new format works for 0.124.0, but you've replaced a working-for-old-Codex system with one that assumes everyone is on the hooks.json-aware build. Goodhart: the test "does not write legacy inline TOML hooks" (tests/codex-config.test.cjs:769-780) optimizes for a metric (no `[[hooks]]`) that is exactly what older Codex *needs*. Either probe `codex --version` or detect `hooks.json` support capability — the README-level promise of "update checks via SessionStart" is now false on <0.124.0 without telling the user.

### Rubber-duck — followups that look like smells

- `getManagedCodexSessionStartCommands` (`bin/install.js:2419-2421`) is a one-line alias of `getManagedCodexLegacyInlineCommands`. Kernighan: the alias obscures that session-start-json and legacy-inline-toml commands are the same string set — delete the alias or give it a reason to exist. Peter's principle applied to function names: they rise to a level of abstraction past their usefulness.
- `bin/install.js:5975-5977` quietly widens `writeManifest` hook tracking from `!isCodex && !isCopilot && !isCline` to `!isCopilot && !isCline`. Correct direction, but out of scope for a PR titled "hooks.json migration" — mention it in the PR body or split. Knuth: premature generalization of the gate is how test coverage for non-Codex runtimes silently drifts.

### Blame

Inline `[[hooks]]` TOML was introduced in **70f1f8b7** "feat: add Codex hooks support for SessionStart (#1020)" — Tom Boucher, 2026-03-17, +30 lines to `bin/install.js`. That commit hardcoded the now-deprecated format.

### Required changes

1. Add Codex version detection (or capability probe for `hooks.json` support) and fall back to legacy `[[hooks]]` write on older Codex. Skip hook install entirely on ambiguous detection rather than silently breaking update checks.
2. Inline or rename `getManagedCodexSessionStartCommands` so the call site at `bin/install.js:2438`/`2506` reads honestly.
3. Note the `writeManifest` gate change in the PR body (or split).

CI is green. Diff is +1106/−200 across 4 files, tests are the bulk of it, and the test matrix is good. Fix the version guard and this ships.

hanzckernel:
Addressed.

- Probe `codex --version` before configuring Codex update hooks.
- Codex `>=0.124.0` uses `hooks.json`; older Codex keeps the legacy inline `[[hooks]]` path.
- Unknown detection skips hook config instead of risking a broken config.
- Added regression coverage for version selection, legacy fallback, and skip mode.

Verified: `node --check bin/install.js`, `node --test tests/codex-config.test.cjs`, `git diff --check`, `npm test`.

trek-e (commented):
## Code Review: Fix Codex hook installation to use hooks.json instead of inline TOML hooks

### 1. Peter Principle concern — 400+ lines of Codex internals from a first-time contributor

This PR adds: version detection via `spawnSync`, custom semver comparison, a partial TOML parser, `hooks.json` read/write/merge logic with path normalization, and a 6-variant command enumeration system for uninstall. That is the implementation surface of an experienced systems programmer who has reverse-engineered Codex's config format in depth. What is the source of the contributor's knowledge of Codex's `hooks.json` schema, version detection API, and migration behavior? Has any of this been validated against Codex's actual behavior, or is it reverse-engineered from observation?

### 2. Kernighan's Law: `getManagedCodexUpdateHookCommands` generates 6 variants because install wrote inconsistent paths

The uninstall function generates 6 command variants (normalized × portable base × 2 hook names × multiple quoting styles) to find what the install function wrote. This is a symptom, not a fix. The root cause is that the install path was written in an inconsistent format. The correct fix is: install writes ONE canonical path, uninstall removes that exact path. Enumerating all possible variants is self-documenting evidence that the install side was not fixed. When does the install path format diverge? Fix that, and the uninstall complexity collapses.

### 3. `parseSimpleTomlStringAssignment` is a partial TOML parser that fails silently

The function handles double-quoted and single-quoted strings. TOML also has: multi-line basic strings (`"""`), multi-line literal strings (`'''`), and escaped characters within strings (`\"`, `\\`, `\n`, etc.). Adjacent keys in a real Codex TOML config may use any of these. When this parser encounters syntax it doesn't recognize, it returns `null` — silently. A caller that receives `null` from this function cannot distinguish "key not found" from "key found but unparseable." Silent null returns on parse failure will cause undetected migration failures.

### 4. `detectCodexHookInstallMode` conflates "Codex not installed" with "version check failed"

The function returns `mode: 'skip'` in both cases. On a machine where Codnet is installed but `codex --version` fails due to a PATH issue or permissions error, the GSD Codex hook is silently skipped — not installed, no error, no warning. The distinction between "Codex is not present" and "Codex is present but misconfigured" has different user-facing implications. One should skip silently; the other should warn.

### 5. `GSD_CODEX_HOOKS_MODE` env var is now a permanent production API surface

Adding an env var override to force production code into a test mode is a code smell. This env var will be discovered, documented (or not), and used by users who want to force a specific install behavior. You've created an undocumented knob with production side effects. This should be test infrastructure only — use dependency injection (pass `installMode` as a parameter) instead of an env var that bleeds into production.

### 6. The complexity-to-benefit ratio has not been justified

The old system had one failure mode: Codex rejected inline TOML hooks. The new system has: version detection failure, `spawnSync` unavailability, `hooks.json` read failure, `hooks.json` parse failure, `hooks.json` write failure, path normalization edge cases, partial TOML parse failure, and 6-variant uninstall mismatch. Gall's Law: a complex system that works must evolve from a simple system that worked. This PR replaced a simple system that had one known failure mode with a complex system that has eight. What evidence exists that the complex system handles all eight failure modes correctly?

VERDICT: CHANGES REQUIRED — fix the install path inconsistency at the root (eliminating the need for 6-variant uninstall), remove the production env var, and validate against actual Codex behavior.

trek-e:

PR #2642, opened 2026-04-23
fix(sdk): surface current milestone wrapped in <details>/<summary> (#2641)

### LINKED ISSUE #2641: roadmap.analyze + init.milestone-op miss the current milestone when it's wrapped in <details>/<summary> (common GitHub collapse layout)
## Environment

- **GSD SDK version:** `@gsd-build/sdk@0.1.0` (globally installed) — upstream `main` at commit `HEAD` exhibits the same behavior.
- **Runtime:** Claude Code
- **OS:** macOS 26.3.1
- **Node:** v25.6.1
- **Shell:** zsh
- **Install method:** `npm install -g @gsd-build/sdk`
- **Frequency:** Every time (100% reproducible)
- **Impact:** Blocker for autonomous workflows — `/gsd-autonomous` cannot discover the current milestone and drives stale archived phases instead.

## What happened?

Three related bugs in `sdk/src/query/roadmap.ts` and `sdk/src/query/init.ts` cause the SDK to produce wrong output when a ROADMAP.md encodes its current milestone inside a `<details>/<summary>` block — a layout widely used because GitHub collapses it by default, keeping the ROADMAP scannable as milestones accumulate.

Symptoms on a project whose STATE.md says `milestone: v10.5` and whose ROADMAP.md has the v10.5 block wrapped in `<details><summary>v10.5 … -- IN PROGRESS …</summary>`:

| Query | Expected | Actual |
|-------|----------|--------|
| `gsd-sdk query init.milestone-op` → `milestone_version` | `v10.5` | `v7.0` (the first shipped `###`-heading milestone) |
| `gsd-sdk query init.milestone-op` → `phase_count` / `completed_phases` | `4 / 1` | `83 / 83` (every phase directory on disk from every milestone ever shipped) |
| `gsd-sdk query init.milestone-op` → `all_phases_complete` | `false` | `true` — a stale `true` the moment a new milestone starts, because every archived directory has a `-SUMMARY.md` |
| `gsd-sdk query roadmap.analyze` → `phases[]` | Phases 110-113 (the v10.5 phases) | Phase 110-113 missing entirely; instead 13 phases from the archived v9.0 milestone appendix surface as `disk_status=no_directory, roadmap_complete=false` — ready to be mis-driven by `/gsd-autonomous` |

Net effect: `/gsd-autonomous` with no args will happily try to discuss/plan/execute 13 already-shipped v9.0 phases (no disk directories, no context, no plans), while the real v10.5 work is invisible to it.

## What did you expect?

1. **`stripShippedMilestones`** should only strip `<details>` blocks whose `<summary>` marks them as shipped (`SHIPPED`, `✅`, `COMPLETE`). A block with an `IN PROGRESS` summary or no marker should be preserved — the current milestone almost always lives in such a block for GitHub collapse UX.
2. **`roadmapAnalyze` checkbox check** should require an explicit `[x] Phase N` match in the **raw ROADMAP** (not the scoped-and-stripped content). Otherwise:
   - Archived-milestone checkboxes inside stripped `<details>` blocks are lost, so phases whose detail `### Phase N` headings leak through from appendix sections get reported as incomplete.
   - The current `(x| )` inclusive regex matches `- [ ] TBD (run /gsd-plan-phase N to break down)` placeholders in appendix "Plans:" subsections, overriding the authoritative `[x]` marker in the milestone block.
3. **`initMilestoneOp`** should scope `phase_count` / `completed_phases` to the current milestone. The simplest correct source is STATE.md's `progress.{total_phases, completed_phases}` block (the place the workflow already updates per-phase); disk-directory count is only safe as a fallback when STATE.md has no progress section.

## Steps to reproduce

Minimal fixture that triggers all three bugs:

```bash
# 1. Make a scratch project
mkdir -p /tmp/gsd-bug-repro/.planning/phases
cd /tmp/gsd-bug-repro

# 2. STATE.md says we are on v2.0 (a fresh milestone)
cat > .planning/STATE.md <<'EOF'
---
milestone: v2.0
milestone_name: Second Milestone
status: phase_in_progress
progress:
  total_phases: 2
  completed_phases: 0
---
EOF

# 3. ROADMAP.md with v1.0 SHIPPED (inside <details>) and v2.0 IN PROGRESS (also inside <details>)
cat > .planning/ROADMAP.md <<'EOF'
# Project Roadmap

<details>
<summary>v1.0 First Milestone (Phases 1-2) — SHIPPED 2026-01-01</summary>

- [x] Phase 1: Foundation (shipped)
- [x] Phase 2: Polish (shipped)

</details>

<details>
<summary>v2.0 Second Milestone (Phases 3-4) -- IN PROGRESS (planning 2026-04-23)</summary>

- [ ] Phase 3: New Feature (not started)
- [ ] Phase 4: Harden (not started)

### Phase 3: New Feature
**Goal:** Ship the new feature end to end.

### Phase 4: Harden
**Goal:** Close gaps found during Phase 3.

</details>

## v1.0 Phase Details (appendix — kept for cross-reference)

### Phase 1: Foundation
**Goal:** Original v1.0 foundation.

Plans:
- [ ] TBD (run /gsd-plan-phase 1 to break down)

### Phase 2: Polish
**Goal:** Original v1.0 polish.

Plans:
- [ ] TBD (run /gsd-plan-phase 2 to break down)
EOF

# 4. Pretend the v1.0 phases shipped (disk directories with -SUMMARY.md)
for n in 1 2; do
  mkdir -p ".planning/phases/0${n}-phase-${n}"
  touch ".planning/phases/0${n}-phase-${n}/0${n}-PLAN.md"
  touch ".planning/phases/0${n}-phase-${n}/0${n}-SUMMARY.md"
done

# 5. Reproduce
gsd-sdk query init.milestone-op | jq '{milestone_version, milestone_name, phase_count, completed_phases, all_phases_complete}'
gsd-sdk query roadmap.analyze     | jq '.phases[] | {number, name, disk_status, roadmap_complete}'
```

**Observed (bug):**

```json
// init.milestone-op
{
  "milestone_version": "v1.0",            // ← WRONG, STATE says v2.0
  "milestone_name": "First Milestone",    // ← WRONG
  "phase_count": 2,
  "completed_phases": 2,                  // ← counted archived v1.0 dirs
  "all_phases_complete": true             // ← completely misleading
}

// roadmap.analyze .phases
{"number": "1", "disk_status": "complete", "roadmap_complete": false}  // leaks from appendix, checkbox lost
{"number": "2", "disk_status": "complete", "roadmap_complete": false}
// Phase 3 and 4 MISSING — v2.0 block stripped as "shipped <details>" by mistake
```

**Expected (fix):**

```json
// init.milestone-op
{
  "milestone_version": "v2.0",
  "milestone_name": "Second Milestone",
  "phase_count": 2,
  "completed_phases": 0,
  "all_phases_complete": false
}

// roadmap.analyze .phases
{"number": "3", "disk_status": "no_directory", "roadmap_complete": false}
{"number": "4", "disk_status": "no_directory", "roadmap_complete": false}
```

## Root cause

All three symptoms stem from shared assumptions:

1. `stripShippedMilestones` (`sdk/src/query/roadmap.ts:53`) treats every `<details>` block as shipped:
   ```ts
   let result = content.replace(/<details>[\s\S]*?<\/details>/gi, '');
   ```
   This deletes the current milestone block whenever it's wrapped in `<details>` for collapse UX. The fix is to inspect each block's `<summary>` and only strip if it carries a `SHIPPED`/`✅`/`COMPLETE` marker.

2. `roadmapAnalyze` checkbox lookup (`sdk/src/query/roadmap.ts:528`):
   ```ts
   const checkboxPattern = new RegExp(`-\\s*\\[(x| )\\]\\s*.*Phase\\s+${escapeRegex(phaseNum)}[:\\s]`, 'i');
   const checkboxMatch = content.match(checkboxPattern);
   ```
   Problem one: `content` is the stripped-and-scoped text, so archived-milestone `[x]` checkboxes are unavailable. Problem two: `.*` gobbles `TBD (run /gsd-plan-phase N to break down)` in appendix Plans subsections, so the first match against a still-leaking appendix phase is a `[ ]` line, silently overriding the real `[x]` in the archived block. The fix is to scan raw content for a tight `[x] Phase N` pattern (require `**` or space after the closing bracket) and treat presence as authoritative.

3. `initMilestoneOp` (`sdk/src/query/init.ts:779-796`) computes `phase_count` / `completed_phases` from disk directories unconditionally. Because every archived milestone's phase directory carries a `-SUMMARY.md`, the totals drift as soon as a new milestone starts. The fix is to prefer STATE.md's `progress.{total_phases, completed_phases}` (authoritative — updated per-phase by the workflow); fall back to the disk scan only when STATE.md has no progress block.

## Workaround

Edit the installed SDK in place (won't survive reinstalls):

```bash
# 1. Change the strip-details helper to only strip SHIPPED blocks.
# 2. Change the checkbox check to require [x] in raw content.
# 3. Change initMilestoneOp to read STATE.md progress first.
```

I've done this locally and verified it fixes all three symptoms against our actual project ROADMAP. Happy to open a PR with the three targeted patches + regression tests — I'll wait for `confirmed-bug` labeling before pushing per `CONTRIBUTING.md`.

## Related discoveries (not in scope of this bug, but worth flagging)

- Upstream already had the right idea for `getMilestoneInfo` (STATE.md primacy via `parseMilestoneFromState`) — our globally installed `0.1.0` lagged upstream. Once you rev the published version, that fix ships automatically.
- Upstream's `extractCurrentMilestone` no longer returns a preamble at all (it returns `content.slice(sectionStart, sectionEnd)`), so the preamble-strip class of bug is already fixed. This is the right direction — keep doing that.

georgian-enache:
## Linked Issue

Fixes #2641

> Issue is currently labeled `bug, needs-triage`. Opening this PR as **draft** so the code is visible to any maintainer reviewing the issue; happy to mark ready-for-review once the bug is confirmed per `CONTRIBUTING.md`. No changes until then.

## What was broken

Three tightly-related bugs caused `roadmap.analyze` + `init.milestone-op` to miss the current milestone when its block is wrapped in `<details>/<summary>` (a common GitHub layout for ROADMAPs that collapse shipped milestones by default):

1. `stripShippedMilestones` deleted *every* `<details>` block, so the current milestone's `### Phase N` headings vanished from downstream parsing.
2. `roadmapAnalyze`'s checkbox check ran against the stripped-and-scoped content with an inclusive `(x| )` + `.*` pattern, so archived-milestone `[x]` markers were invisible (inside the stripped block) *and* `- [ ] TBD (run /gsd-plan-phase N to break down)` placeholders in appendix `Plans:` subsections silently overrode the real `[x]`.
3. `initMilestoneOp` counted every phase directory on disk — including archived milestones — giving a false `"N/N complete, all_phases_complete=true"` the moment a new milestone started, because every archived directory has a `-SUMMARY.md`.

## What this fix does

1. **`stripShippedMilestones`** now inspects `<summary>` and only strips blocks marked `SHIPPED` / `✅` / `COMPLETE`. Blocks with "IN PROGRESS" or no marker are preserved. Blocks with no `<summary>` are treated as shipped for back-compat with pre-summary ROADMAPs.
2. **`roadmapAnalyze` checkbox check** now scans the **raw ROADMAP** for a tight `[x] Phase N` pattern (requires space or `**` before `Phase`). This surfaces archived-milestone `[x]` markers and avoids false positives from appendix `[ ] TBD` placeholders.
3. **`initMilestoneOp`** reads STATE.md's `progress.{total_phases, completed_phases}` (authoritative — updated per-phase by the workflow). Falls back to the legacy disk scan only when STATE.md has no progress block.

## Root cause

The strip-details regex was written when the convention was "`<details>` = shipped" and predates the now-common GitHub layout of collapsing the *current* milestone to keep the ROADMAP scannable. The checkbox check relied on the same stripped output for truth, which meant anything stripped was lost; and `initMilestoneOp`'s disk-scan semantics conflated "total shipped across history" with "current milestone progress." Each bug reinforced the others: the strip hid the milestone, which left the analyzer looking in the wrong place, which left `initMilestoneOp` falling back to disk-globals.

### How I verified the fix

- Ran the end-to-end repro fixture from #2641 against the patched SDK. All observed outputs now match the "expected" table in the issue body.
- Ran the unit test suites for both touched files: **76/76 pass** (`vitest run src/query/roadmap.test.ts src/query/init.test.ts`).
- Ran the full suite: 1539 pass / 9 unrelated failures. Baseline-verified that all 9 failures exist on upstream `main` without any of my changes (3 state-mutation flag-parser tests, `agentSkills`, `configNewProject`, `QueryRegistry dispatch`, and 3 golden-parity integration tests that need an external `gsd-tools.cjs` binary). Zero regressions from this PR.

### Regression test added?

- [x] Yes — 8 new tests across both files:
  - `stripShippedMilestones`: preserves IN-PROGRESS `<details>`, strips SHIPPED/✅-marked blocks, back-compat for `<details>` without `<summary>` (3 tests)
  - `roadmapAnalyze`: surfaces current milestone wrapped in `<details>/<summary>`, honours archived-milestone `[x]` over appendix `[ ] TBD` placeholders (2 tests)
  - `initMilestoneOp`: scopes counts to current milestone via STATE.md, falls back to disk scan when STATE.md has no progress block (2 tests)
  - Plus a back-compat test ensuring existing behaviour preserved for legacy ROADMAPs.

### Platforms tested

- [x] macOS
- [ ] Windows — not platform-specific (pure string parsing)
- [ ] Linux — not platform-specific

### Runtimes tested

- [x] Claude Code
- [ ] Other: not runtime-specific (pure SDK query handlers)

---

## Checklist

- [x] Issue linked above with `Fixes #2641`
- [ ] Linked issue has the `confirmed-bug` label — **pending triage**; opening as draft per CONTRIBUTING
- [x] Regression test added
- [x] All roadmap.ts + init.ts tests pass; full-suite unrelated failures baseline-checked on upstream `main`

## Breaking changes

None. The `<details>` without `<summary>` case (legacy ROADMAPs) still strips identically. STATE.md fallback only activates when a `progress:` block is present — absent that, behavior is byte-identical to the previous disk scan.

### DIFF
--- a/sdk/src/query/init.test.ts
+++ b/sdk/src/query/init.test.ts
@@ -447,6 +447,68 @@ describe('initMilestoneOp', () => {
     expect(data.completed_phases).toBeGreaterThanOrEqual(0);
     expect(data.project_root).toBe(tmpDir);
   });
+
+  // Bug #2641: phase_count/completed_phases must scope to the CURRENT
+  // milestone, not every phase directory ever shipped. STATE.md's
+  // progress.{total_phases, completed_phases} is authoritative.
+  it('scopes phase_count/completed_phases to current milestone via STATE.md', async () => {
+    const { writeFile, mkdir } = await import('node:fs/promises');
+    const { join } = await import('node:path');
+
+    // Simulate 5 archived phase directories from prior milestones (each
+    // with a -SUMMARY.md, which the legacy code counted toward completion).
+    for (let i = 1; i <= 5; i++) {
+      const dir = join(tmpDir, '.planning', 'phases', `${String(i).padStart(2, '0')}-old-phase-${i}`);
+      await mkdir(dir, { recursive: true });
+      await writeFile(join(dir, `${String(i).padStart(2, '0')}-PLAN.md`), '');
+      await writeFile(join(dir, `${String(i).padStart(2, '0')}-SUMMARY.md`), '');
+    }
+
+    // STATE.md says we are on a fresh milestone with 4 phases and 1 done.
+    await writeFile(
+      join(tmpDir, '.planning', 'STATE.md'),
+      [
+        '---',
+        'milestone: v2.0',
+        'milestone_name: Current Milestone',
+        'progress:',
+        '  total_phases: 4',
+        '  completed_phases: 1',
+        '---',
+      ].join('\n'),
+    );
+
+    const result = await initMilestoneOp([], tmpDir);
+    const data = result.data as Record<string, unknown>;
+    expect(data.phase_count).toBe(4);
+    expect(data.completed_phases).toBe(1);
+    expect(data.all_phases_complete).toBe(false);
+  });
+
+  it('falls back to disk scan when STATE.md has no progress block', async () => {
+    const { writeFile, mkdir, rm } = await import('node:fs/promises');
+    const { join } = await import('node:path');
+
+    // Reset the phases/ directory so we own the entire fixture.
+    await rm(join(tmpDir, '.planning', 'phases'), { recursive: true, force: true });
+    for (let i = 1; i <= 3; i++) {
+      const dir = join(tmpDir, '.planning', 'phases', `${String(i).padStart(2, '0')}-legacy-${i}`);
+      await mkdir(dir, { recursive: true });
+      await writeFile(join(dir, `${String(i).padStart(2, '0')}-PLAN.md`), '');
+      if (i < 3) await writeFile(join(dir, `${String(i).padStart(2, '0')}-SUMMARY.md`), '');
+    }
+    // STATE.md without a progress block — legacy behaviour applies.
+    await writeFile(
+      join(tmpDir, '.planning', 'STATE.md'),
+      '---\nmilestone: v1.0\n---\n',
+    );
+
+    const result = await initMilestoneOp([], tmpDir);
+    const data = result.data as Record<string, unknown>;
+    expect(data.phase_count).toBe(3);
+    expect(data.completed_phases).toBe(2);
+    expect(data.all_phases_complete).toBe(false);
+  });
 });
 
 describe('initMapCodebase', () => {
--- a/sdk/src/query/init.ts
+++ b/sdk/src/query/init.ts
@@ -777,22 +777,42 @@ export const initMilestoneOp: QueryHandler = async (_args, projectDir) => {
   const milestone = await getMilestoneInfo(projectDir);
 
   const phasesDir = join(planningDir, 'phases');
+  // Scope phase counts to the CURRENT milestone. The previous implementation
+  // counted every phase directory on disk — including archived milestones —
+  // which produced misleading "N/N complete, all_phases_complete=true"
+  // readings the moment a new milestone started (issue #2641). Priority:
+  //   1. STATE.md `progress.{total_phases, completed_phases}` (authoritative —
+  //      updated per-phase by the workflow itself).
+  //   2. Fall back to the legacy disk scan when STATE.md has no progress block.
   let phaseCount = 0;
   let completedPhases = 0;
-
+  let progressSourceFromState = false;
   try {
-    const entries = readdirSync(phasesDir, { withFileTypes: true });
-    const dirs = entries.filter(e => e.isDirectory()).map(e => e.name);
-    phaseCount = dirs.length;
-
-    for (const dir of dirs) {
-      try {
-        const phaseFiles = readdirSync(join(phasesDir, dir));
-        const hasSummary = phaseFiles.some(f => f.endsWith('-SUMMARY.md') || f === 'SUMMARY.md');
-        if (hasSummary) completedPhases++;
-      } catch { /* intentionally empty */ }
+    const stateRaw = readFileSync(join(planningDir, 'STATE.md'), 'utf-8');
+    const totalMatch = stateRaw.match(/^\s*total_phases:\s*(\d+)/m);
+    const doneMatch = stateRaw.match(/^\s*completed_phases:\s*(\d+)/m);
+    if (totalMatch && doneMatch) {
+      phaseCount = parseInt(totalMatch[1], 10);
+      completedPhases = parseInt(doneMatch[1], 10);
+      progressSourceFromState = true;
     }
-  } catch { /* intentionally empty */ }
+  } catch { /* STATE.md optional */ }
+
+  if (!progressSourceFromState) {
+    try {
+      const entries = readdirSync(phasesDir, { withFileTypes: true });
+      const dirs = entries.filter(e => e.isDirectory()).map(e => e.name);
+      phaseCount = dirs.length;
+
+      for (const dir of dirs) {
+        try {
+          const phaseFiles = readdirSync(join(phasesDir, dir));
+          const hasSummary = phaseFiles.some(f => f.endsWith('-SUMMARY.md') || f === 'SUMMARY.md');
+          if (hasSummary) completedPhases++;
+        } catch { /* intentionally empty */ }
+      }
+    } catch { /* intentionally empty */ }
+  }
 
   const archiveDir = join(projectDir, '.planning', 'archive');
   let archivedMilestones: string[] = [];
--- a/sdk/src/query/roadmap.test.ts
+++ b/sdk/src/query/roadmap.test.ts
@@ -159,6 +159,57 @@ describe('stripShippedMilestones', () => {
     expect(stripped).toContain('v2.0');
     expect(stripped).toContain('Current');
   });
+
+  // Bug #2641: current-milestone <details> block with "IN PROGRESS"
+  // summary must NOT be stripped (common GitHub collapse layout for
+  // long-running ROADMAPs that accumulate many shipped milestones).
+  it('preserves <details> block when <summary> says IN PROGRESS', () => {
+    const content = [
+      '<details>',
+      '<summary>v1.0 Old (Phases 1-2) — SHIPPED 2026-01-01</summary>',
+      'shipped stuff',
+      '</details>',
+      '',
+      '<details>',
+      '<summary>v2.0 Current (Phases 3-4) -- IN PROGRESS 2026-04-23</summary>',
+      '',
+      '### Phase 3: Work',
+      '',
+      '</details>',
+    ].join('\n');
+    const stripped = stripShippedMilestones(content);
+    expect(stripped).not.toContain('v1.0 Old');
+    expect(stripped).not.toContain('shipped stuff');
+    expect(stripped).toContain('v2.0 Current');
+    expect(stripped).toContain('### Phase 3: Work');
+  });
+
+  it('strips <details> block when <summary> contains SHIPPED marker', () => {
+    const content = [
+      '<details>',
+      '<summary>v1.0 Old — SHIPPED 2026-01-01</summary>',
+      'shipped phase list',
+      '</details>',
+      'after',
+    ].join('\n');
+    expect(stripShippedMilestones(content)).toBe('\nafter');
+  });
+
+  it('strips <details> block when <summary> uses ✅ marker', () => {
+    const content = [
+      '<details>',
+      '<summary>✅ v1.0 Old — SHIPPED 2026-01-01</summary>',
+      'shipped phase list',
+      '</details>',
+    ].join('\n');
+    expect(stripShippedMilestones(content).trim()).toBe('');
+  });
+
+  it('strips <details> block when <summary> is missing (back-compat)', () => {
+    // Legacy ROADMAPs without a <summary> are still treated as shipped.
+    const content = '<details>legacy content</details>after';
+    expect(stripShippedMilestones(content)).toBe('after');
+  });
 });
 
 // ─── getMilestoneInfo ─────────────────────────────────────────────────────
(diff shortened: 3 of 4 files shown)


trek-e (commented):
## Adversarial review — PR #2642

**Verdict:** LGTM with one reservation. Fix is well-scoped, RCA is correct, tests are substantive, no regressions. Ready for ready-for-review once bug is triaged.

### Fix correctness (rubber-duck)

Walked end-to-end against the #2641 fixture:
1. `stripShippedMilestones` — `<details>` block replaced via callback that inspects `<summary>` for `SHIPPED|✅|COMPLETE`. IN-PROGRESS block preserved → `extractCurrentMilestone` now sees `### Phase 3/4`. Missing-summary → stripped (back-compat). Correct.
2. `roadmapAnalyze` — checkbox scan moved to `rawContent` with tight `\[x\]\s+(?:\*\*)?Phase N[:\s]`. Archived `[x]` surfaces; appendix `[ ] TBD` no longer shadows. Correct.
3. `initMilestoneOp` — reads STATE.md `total_phases`/`completed_phases` first, disk scan fallback. Correct.

### Adversarial angle: regex-based HTML

The core bug IS regex-as-HTML-parser brittleness. The fix doubles down on it. I probed:

- **Nested `<details>`?** `/<details>[\s\S]*?<\/details>/gi` is non-greedy and will mis-match on nesting (closes on the inner `</details>`, leaving the outer wrapper content orphaned). Unlikely in a ROADMAP, but: not defended against, not tested. Worth an inline TODO or a `<details ...>` attribute-tolerant variant.
- **`<details open>` / `<details class="...">`?** `<details>` literal does NOT match `<details open>`. GitHub's own collapse UX commonly uses `<details open>`. **This is a real gap** — the fix preserves the named bug class but not this adjacent one. Suggest: `/<details\b[^>]*>[\s\S]*?<\/details>/gi`.
- **Self-closing / malformed?** N/A in markdown.
- **Case of `<SUMMARY>`?** `i` flag handles it.

### Thinker lenses

- **Gall:** simple flat-markdown system worked; complex HTML-aware logic is being bolted on. Gall says the complex system evolved from a working simple one — which IS what's happening here incrementally. Pass, but: the underlying question (why is GSD emitting markdown that requires HTML parsing to consume?) is not addressed. Long-term the ROADMAP schema should carry milestone status in frontmatter or a dedicated machine-readable block, not inferred from `<summary>` prose.
- **Goodhart:** `all_phases_complete=true` became a target that decoupled from reality the moment phase dirs accumulated. STATE.md fallback re-couples it to the workflow's own truth. Good.
- **Kernighan:** debugging twice as hard as writing — the fix adds comments pointing at #2641, keeps the regex simple enough to re-debug. Pass.
- **Knuth:** premature-opt doesn't apply; no hot path. Pass.
- **Peter:** the strip regex was promoted beyond its competence (one-rule-fits-all `<details>`). Demoting it to summary-gated is right-sized.

### Blame

Bare `<details>[\s\S]*?<\/details>` regex introduced in **2980f0ec4** (Tom Boucher, 2026-04-21) — `fix(sdk): stripShippedMilestones handles inline SHIPPED headings; getMilestoneInfo prefers STATE.md (#2508)`. That commit taught the helper about inline `✅ SHIPPED` ## headings but did not gate the `<details>` branch on the `<summary>` marker — a missed symmetry. Original helper ported in 002bcf2a8 (2026-04-12).

### Asks (non-blocking)

1. Widen regex to `<details\b[^>]*>` to tolerate `<details open>` / attributes.
2. Add a test for `<details open>` to pin the contract.
3. File a follow-up issue to move milestone status to ROADMAP frontmatter (kill HTML-parsing-of-markdown class of bug at the root).

Otherwise approved in spirit — mark ready-for-review once #2641 is triaged.

trek-e (commented):
## Code Review: fix(sdk): surface current milestone wrapped in <details>/<summary> [DRAFT]

This is a draft PR. These comments are at draft quality — sufficient to block merge, not a full acceptance review.

### 1. `progressSourceFromState` is dead code

The variable is assigned `true` or `false` inside the `if (stateContent)` block but is never read after the if-block closes. It ships to production as a dead assignment. Either use it or delete it.

### 2. `parseInt` result is not NaN-guarded downstream

`parseInt(totalMatch[1], 10)` returns `NaN` if `STATE.md` contains `total_phases: abc` (corrupted or hand-edited). The code then assigns `data.phase_count = NaN`. Any consumer that does arithmetic on `phase_count` gets `NaN` propagation silently. At minimum, validate the result: `if (isNaN(phaseCount)) throw ...` or fall through to the disk-scan fallback.

### 3. The PR title and the diff describe two different bugs

"Surface current milestone wrapped in `<details>/<summary>`" describes a rendering/strip issue. The implementation change (reading `total_phases`/`completed_phases` from `STATE.md` instead of scanning disk) is a phase-counting data-source change. The test additions for `stripShippedMilestones` with `<details>` blocks are testing yet a third thing. These are separate bugs in one PR with a misleading title. Split them, or retitle accurately.

### 4. STATE.md is not transactionally safe — you've made a stale source authoritative

STATE.md is updated by workflows. If a workflow crashes after completing phase 2 but before writing STATE.md, STATE.md says `completed_phases: 1` while the disk has 2 phase summaries. The new code uses STATE.md as the authoritative source and only falls back to disk scan when STATE.md has **no progress block** — not when STATE.md is stale. The fallback is too narrow. You need staleness detection, not just absence detection.

### 5. `<details>` stripping uses `-- IN PROGRESS` (double dash) as a sentinel

Is this convention documented? `--- IN PROGRESS` (triple dash, common in markdown HR contexts) and `— IN PROGRESS` (em-dash) would not match. If the convention is double-dash-only, it needs to be enforced at write time too, not just matched at read time. Where does the `-- IN PROGRESS` marker get written?

### 6. Draft PRs should not be reviewed at merge-candidate depth

This review is being written against a draft. The author has not signaled this is ready for review. Do not merge until the draft status is lifted and all items above are resolved.

VERDICT: CHANGES REQUIRED (draft — do not merge)

trek-e:

Issue #2652, opened 2026-04-24
Codex install can still emit Claude model aliases from gsd-sdk init.quick


tomas-lm:
# Bug: Codex install can still emit Claude model aliases from `gsd-sdk query init.quick`

## Summary

When GSD is installed for Codex, the installer writes `resolve_model_ids: "omit"` to `~/.gsd/defaults.json`. The docs say this should make agent model fields empty so Codex can use its runtime/default model.

In existing projects with `.planning/config.json`, `gsd-sdk query init.quick` can still return Claude-style model aliases such as `"opus"` / `"sonnet"` for `planner_model`, `executor_model`, etc.

This causes Codex workflows to receive model hints intended for Claude, and can make `$gsd-quick` behave incorrectly or inconsistently on Codex.

## Environment

- Runtime: Codex
- Platform: macOS
- Shell: zsh
- Node used by local `gsd-sdk`: `v20.19.4`
- Installed package version observed locally: `get-shit-done-cc@1.38.3`
- Install command used:

```bash
npx get-shit-done-cc@latest --codex --global
```

The installed `gsd-sdk` binary resolves to:

```text
~/.nvm/versions/node/v20.19.4/bin/gsd-sdk
```

The package version check:

```bash
node -p "require('<installed get-shit-done-cc package>/package.json').version"
```

Output:

```text
1.38.3
```

Global defaults:

```bash
cat ~/.gsd/defaults.json
```

Output:

```json
{
  "resolve_model_ids": "omit"
}
```

## Reproduction

From an existing GSD repo that already has `.planning/config.json` and was created/used before the Codex install:

```bash
gsd-sdk query resolve-model gsd-planner
```

Observed before applying a local workaround:

```json
{
  "model": "opus",
  "profile": "balanced"
}
```

Then:

```bash
gsd-sdk query init.quick "codex model smoke"
```

Observed before applying a local workaround:

```json
{
  "planner_model": "opus",
  "executor_model": "sonnet",
  "checker_model": "sonnet",
  "verifier_model": "sonnet"
}
```

I also saw related Codex runtime drift in the same local install: `init.quick` reported `agents_installed: false` even though the Codex agents existed under `~/.codex/agents`. I noticed that upstream `main` appears to already mention a runtime-aware fix for that part in the changelog, so this report is focused on the model/defaults behavior.

## Expected Behavior

[...]

PR #2654, opened 2026-04-24
fix(sdk): honor global model defaults

tomas-lm:
Addresses #2652.

Summary:
- Teach SDK loadConfig() to merge hardcoded defaults <- ~/.gsd/defaults.json <- .planning/config.json.
- Preserve deep-merge behavior for git, workflow, hooks, agent_skills, and model_overrides.
- Preserve explicit empty model aliases from resolveModel() so init.quick does not fall back to sonnet when resolve_model_ids is omit.
- Add regression coverage for config loading, resolveModel(), and initQuick().

Verification:
- PASS: `npx vitest run --project unit src/config.test.ts src/query/config-query.test.ts src/query/init.test.ts`
- PASS: `npm run build`
- PASS: `git diff --check`
- NOTE: full `npx vitest run --project unit` still has unrelated failures on this upstream fix branch in state-mutation, config-mutation, decomposed-handlers, and registry tests.

### DIFF
--- a/sdk/src/config.test.ts
+++ b/sdk/src/config.test.ts
@@ -6,15 +6,28 @@ import { tmpdir } from 'node:os';
 
 describe('loadConfig', () => {
   let tmpDir: string;
+  let previousGsdHome: string | undefined;
 
   beforeEach(async () => {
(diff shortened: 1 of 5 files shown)


trek-e:
prs are against main only, not other branches.   please rebase to main and resubmit

Issue #1683, opened 2026-04-04
Apply ~/.gsd/defaults.json as fallback for pre-project commands


rodzved:
### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature or behavior does this improve?

`loadConfig()` in `core.cjs` — the central config resolution used by all GSD commands.

### Current behavior

`loadConfig()` tries `.planning/config.json`. If missing, it returns hardcoded defaults. `~/.gsd/defaults.json` is never consulted.

This means commands that run **before** project initialization — like `map-codebase` — ignore the user's global settings entirely. For example, a user who sets `model_profile: "quality"` and `model_overrides.gsd-codebase-mapper: "opus"` in `defaults.json` expects those to apply, but the mapper runs with `haiku` (hardcoded balanced profile) because no `config.json` exists yet.

```
# What loadConfig() does today:
.planning/config.json → if missing → hardcoded defaults
```

### Proposed behavior

When `.planning/config.json` is missing, `loadConfig()` should check `~/.gsd/defaults.json` as an intermediate fallback:

```
# Proposed merge order:
hardcoded defaults ← ~/.gsd/defaults.json ← .planning/config.json
```

The merge pattern already exists in `buildNewProjectConfig()` in `config.cjs` — it just needs to be ported to `loadConfig()`.

After this change, running `map-codebase` on a repo without `.planning/config.json` would pick up `model_profile: "quality"` from `~/.gsd/defaults.json` instead of falling back to the hardcoded balanced profile.

### Reason and benefit

**Why the current behavior is a problem:**
`~/.gsd/defaults.json` exists specifically so users can set global preferences once. But any command that runs before `/gsd-new-project` (which creates `.planning/config.json`) silently ignores those preferences. `map-codebase` is the primary example — it's designed for brownfield use before project init, yet it's the exact command where global defaults are most needed. Users set their preferred model profile in `defaults.json`, run `map-codebase`, and get unexpected results with no indication their settings were ignored.

[...]

PR #2658, opened 2026-04-24
fix(sdk): honor global model defaults

### LINKED ISSUE #2652: Codex install can still emit Claude model aliases from gsd-sdk init.quick

tomas-lm:
## Fix PR

---

## Linked Issue

Fixes #2652

---

## What was broken

Codex installs write `resolve_model_ids: "omit"` into `~/.gsd/defaults.json`, but SDK `loadConfig()` did not read user-level defaults. SDK init queries therefore fell back to Claude model aliases like `opus` and `sonnet` even in Codex runtime flows.

## What this fix does

SDK `loadConfig()` now merges `CONFIG_DEFAULTS <- ~/.gsd/defaults.json <- .planning/config.json`. It also preserves explicit empty model IDs from `resolveModel()` so `init.quick` does not replace an intentionally omitted model with `sonnet`.

## Root cause

The SDK config loader only layered hardcoded defaults under project config. It lacked the CJS config path's user-level defaults layer, so Codex runtime defaults were ignored by SDK query callers.

### How I verified the fix

- `npx vitest run --project unit src/config.test.ts src/query/config-query.test.ts src/query/init.test.ts`
- `npm run build`
- `git diff --check`

Full `npx vitest run --project unit` still has unrelated failures on current upstream main in `state-mutation`, `config-mutation`, `decomposed-handlers`, and `registry` tests.

### Regression test added?

- [x] Yes — added tests that verify global defaults merge behavior, `resolveModel()` with `resolve_model_ids: "omit"` from global defaults, and `initQuick()` preserving omitted model aliases.
- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Codex SDK query path

---

## Checklist

- [ ] All existing tests pass (`npm test`) — focused regression tests and SDK build pass; full SDK unit suite has unrelated existing failures on upstream main
- [ ] CHANGELOG.md updated if this is a user-facing fix

## Breaking changes

None

## Summary by CodeRabbit

* **New Features**
  * Global user defaults, feature flags, and per-agent model overrides are now supported and layered into project configs; new defaults influence project creation (e.g., docs commits and parallelization).

* **Bug Fixes**
  * Model alias handling tightened to accept only valid strings.
  * Malformed or missing global defaults are safely ignored in favor of sane defaults.

* **Tests**
  * Expanded, isolated tests for config layering, deep-merge behavior, and model-resolution scenarios.

### DIFF
--- a/sdk/src/config.test.ts
+++ b/sdk/src/config.test.ts
@@ -1,52 +1,47 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { loadConfig, CONFIG_DEFAULTS } from './config.js';
-import { mkdir, writeFile, rm } from 'node:fs/promises';
+import { mkdir, mkdtemp, writeFile, rm } from 'node:fs/promises';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
 
 describe('loadConfig', () => {
   let tmpDir: string;
-  let fakeHome: string;
-  let prevHome: string | undefined;
-  let prevGsdHome: string | undefined;
+  let homeDir: string;
 
   beforeEach(async () => {
-    tmpDir = join(tmpdir(), `gsd-config-test-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+    tmpDir = await mkdtemp(join(tmpdir(), 'gsd-config-test-'));
+    homeDir = await mkdtemp(join(tmpdir(), 'gsd-home-test-'));
     await mkdir(join(tmpDir, '.planning'), { recursive: true });
-    // Isolate ~/.gsd/defaults.json by pointing HOME at an empty tmp dir.
-    fakeHome = join(tmpdir(), `gsd-home-test-${Date.now()}-${Math.random().toString(36).slice(2)}`);
-    await mkdir(fakeHome, { recursive: true });
-    prevHome = process.env.HOME;
-    process.env.HOME = fakeHome;
-    // Also isolate GSD_HOME (loadUserDefaults prefers it over HOME).
-    prevGsdHome = process.env.GSD_HOME;
-    delete process.env.GSD_HOME;
   });
 
   afterEach(async () => {
     await rm(tmpDir, { recursive: true, force: true });
-    await rm(fakeHome, { recursive: true, force: true });
-    if (prevHome === undefined) delete process.env.HOME;
-    else process.env.HOME = prevHome;
-    if (prevGsdHome === undefined) delete process.env.GSD_HOME;
-    else process.env.GSD_HOME = prevGsdHome;
+    await rm(homeDir, { recursive: true, force: true });
   });
 
+  /**
+   * Write isolated user-level defaults under the test-controlled home dir.
+   */
   async function writeUserDefaults(defaults: unknown) {
-    await mkdir(join(fakeHome, '.gsd'), { recursive: true });
-    await writeFile(join(fakeHome, '.gsd', 'defaults.json'), JSON.stringify(defaults));
+    await mkdir(join(homeDir, '.gsd'), { recursive: true });
+    const content = typeof defaults === 'string' ? defaults : JSON.stringify(defaults);
+    await writeFile(join(homeDir, '.gsd', 'defaults.json'), content);
+  }
+
+  function loadTestConfig() {
+    return loadConfig(tmpDir, undefined, { homeDir });
   }
 
   it('returns all defaults when config file is missing', async () => {
     // No config.json created
     await rm(join(tmpDir, '.planning', 'config.json'), { force: true });
-    const config = await loadConfig(tmpDir);
+    const config = await loadTestConfig();
     expect(config).toEqual(CONFIG_DEFAULTS);
   });
 
   it('returns all defaults when config file is empty', async () => {
     await writeFile(join(tmpDir, '.planning', 'config.json'), '');
-    const config = await loadConfig(tmpDir);
+    const config = await loadTestConfig();
     expect(config).toEqual(CONFIG_DEFAULTS);
   });
 
@@ -60,7 +55,7 @@ describe('loadConfig', () => {
       JSON.stringify(userConfig),
     );
 
-    const config = await loadConfig(tmpDir);
+    const config = await loadTestConfig();
 
     expect(config.model_profile).toBe('fast');
     expect(config.workflow.research).toBe(false);
@@ -70,6 +65,94 @@ describe('loadConfig', () => {
     // Top-level defaults preserved
     expect(config.commit_docs).toBe(true);
     expect(config.parallelization).toBe(true);
+  });
+
+  it('layers user defaults from ~/.gsd/defaults.json when project config is missing', async () => {
+    await rm(join(tmpDir, '.planning', 'config.json'), { force: true });
+    await writeUserDefaults({
+      resolve_model_ids: 'omit',
+      workflow: { auto_advance: true },
+      agent_skills: { 'gsd-planner': ['codex-skill'] },
+      features: { global_learnings: true },
+      model_overrides: { 'gsd-planner': 'openai/gpt-5.4' },
+    });
+
+    const config = await loadTestConfig();
+
+    expect(config.resolve_model_ids).toBe('omit');
+    expect(config.workflow.auto_advance).toBe(true);
+    expect(config.workflow.research).toBe(true);
+    expect(config.agent_skills).toEqual({ 'gsd-planner': ['codex-skill'] });
+    expect(config.features).toEqual({ global_learnings: true });
+    expect(config.model_overrides).toEqual({ 'gsd-planner': 'openai/gpt-5.4' });
+  });
+
+  it('lets project config override user defaults while preserving nested fallbacks', async () => {
+    await writeUserDefaults({
+      resolve_model_ids: 'omit',
+      workflow: { auto_advance: true, research: false },
+      git: { branching_strategy: 'phase' },
+      agent_skills: { 'gsd-planner': 'global-skill' },
+      features: { global_learnings: true, thinking_partner: true },
+      model_overrides: {
+        'gsd-planner': 'global-planner',
+        'gsd-executor': 'global-executor',
+      },
+    });
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({
+        resolve_model_ids: false,
+        workflow: { auto_advance: false },
+        git: { branching_strategy: 'none' },
+        agent_skills: { 'gsd-planner': 'project-skill' },
+        features: { thinking_partner: false },
+        model_overrides: { 'gsd-planner': 'project-planner' },
+      }),
+    );
+
+    const config = await loadTestConfig();
+
+    expect(config.resolve_model_ids).toBe(false);
+    expect(config.workflow.auto_advance).toBe(false);
+    expect(config.workflow.research).toBe(false);
+    expect(config.git.branching_strategy).toBe('none');
+    expect(config.agent_skills).toEqual({ 'gsd-planner': 'project-skill' });
+    expect(config.features).toEqual({ global_learnings: true, thinking_partner: false });
+    expect(config.model_overrides).toEqual({
+      'gsd-planner': 'project-planner',
+      'gsd-executor': 'global-executor',
+    });
+  });
+
+  it('deep-merges features from defaults, user defaults, and project config', async () => {
+    await writeUserDefaults({
+      features: { global_learnings: true, thinking_partner: true },
+    });
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({ features: { thinking_partner: false } }),
+    );
+
+    const config = await loadTestConfig();
+
+    expect(config.features).toEqual({
+      global_learnings: true,
+      thinking_partner: false,
+    });
+  });
+
+  it('ignores malformed user defaults', async () => {
+    await writeUserDefaults('{bad json');
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({ model_profile: 'fast' }),
+    );
+
+    const config = await loadTestConfig();
+
+    expect(config.model_profile).toBe('fast');
+    expect(config.resolve_model_ids).toBe(CONFIG_DEFAULTS.resolve_model_ids);
   });
 
   it('partial config merges correctly for nested objects', async () => {
@@ -82,7 +165,7 @@ describe('loadConfig', () => {
       JSON.stringify(userConfig),
     );
 
-    const config = await loadConfig(tmpDir);
+    const config = await loadTestConfig();
 
     expect(config.git.branching_strategy).toBe('milestone');
     // Other git defaults preserved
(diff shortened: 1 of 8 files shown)


trek-e (changes requested):

**Verdict:** Core approach is correct. Four issues before merge: a missing section in `mergeConfig`, an undocumented env var contract, test isolation, and critical overlap with #2675.

---

### Rubber-duck: what is the problem being solved?

`loadConfig` (SDK query path) only merged `CONFIG_DEFAULTS` with project config. It missed the user-level defaults layer that the CJS path already handled. Result: Codex runtime defaults (`resolve_model_ids: "omit"`) were ignored by SDK query callers. Root cause identified correctly.

---

### Critical: `features` section missing from `mergeConfig`

```ts
function mergeConfig(base: GSDConfig, override: Record<string, unknown>): GSDConfig {
  return {
    ...structuredClone(base),
    ...override,
    git: { ...base.git, ...(isPlainRecord(override.git) ? override.git : {}) },
    workflow: { ...base.workflow, ...(isPlainRecord(override.workflow) ? override.workflow : {}) },
    hooks: { ...base.hooks, ...(isPlainRecord(override.hooks) ? override.hooks : {}) },
    agent_skills: { ...base.agent_skills, ...(isPlainRecord(override.agent_skills) ? override.agent_skills : {}) },
    model_overrides: { ...base.model_overrides, ...(isPlainRecord(override.model_overrides) ? override.model_overrides : {}) },
  };
}
```

`features` is absent. `CONFIG_DEFAULTS` has a `features` field (per `sdk/src/config.ts`). If global defaults set `features.*` keys, the top-level `...override` spread will clobber the entire `features` object from `base` with `override.features` — no nested merge. Meanwhile PR #2675 (open) correctly includes `features` in its nested merge in `config-mutation.ts`.

Two open PRs, same documented contract (`hardcoded <- globalDefaults <- userChoices`), two implementations, one of which is wrong for `features`. Add `features` to `mergeConfig` and add a test.

---

### Gall's Law — overlap with #2675

This PR teaches `loadConfig` (SDK query path) about global defaults. PR #2675 teaches `configNewProject` (CJS/config-mutation path) about global defaults. Both are open simultaneously and both implement the three-way merge inline, independently. If both merge:

- `config.ts:mergeConfig` handles 5 nested sections (missing `features`)
- `config-mutation.ts:configNewProject` handles 5 nested sections (different inline spread pattern)
- Neither uses the other's implementation

This is the "two independent allowlists" drift pattern that produced bug #2601. The fix for #2601 is currently re-seeding the same pattern at a different level. These two PRs need to be coordinated: export `mergeConfig` from `config.ts` and import it in `config-mutation.ts`, or explicitly document why they're intentionally separate.

---

### `structuredClone` is wasteful in `mergeConfig`

```ts
return {
  ...structuredClone(base),   // clones everything
  ...override,                // overwrites top-level keys including nested sections
  git: { ...base.git, ... },  // overwrites git from clone
  workflow: { ...base.workflow, ... }, // overwrites workflow from clone
  // ...etc
};
```

`structuredClone(base)` clones all nested sections, which are then immediately overwritten by the explicit nested spreads at the bottom. The clone is applied to intermediate state that's discarded. Use:

```ts
return {
  ...base,
  ...override,
  git: { ...base.git, ... },
  // ...
} as GSDConfig;
```

Or if mutation safety is required, clone the result not the input.

---

### `GSD_HOME` is an undocumented contract

`loadGlobalDefaults` uses `process.env.GSD_HOME || homedir()`. This env var is used in tests as a test isolation hook but is not documented in:
- `docs/ARCHITECTURE.md`
- `README.md`  
- Any config documentation

Users will discover it only by reading the source. Either document it (and own the support surface) or use a proper dependency injection pattern for the home path in tests:

```ts
export async function loadGlobalDefaults(homeDir = homedir()): Promise<Record<string, unknown>> {
```

This eliminates the env var entirely, makes the test contract explicit, and avoids a global env side-effect in `afterEach` failures.

---

### Test isolation — `GSD_HOME` leak risk

```ts
beforeEach(async () => {
  previousGsdHome = process.env.GSD_HOME;
  process.env.GSD_HOME = join(tmpDir, 'home');
});

afterEach(async () => {
  if (previousGsdHome === undefined) delete process.env.GSD_HOME;
  else process.env.GSD_HOME = previousGsdHome;
  await rm(tmpDir, { recursive: true, force: true });
});
```

If `rm` throws in `afterEach`, the `GSD_HOME` restoration already happened — that's fine. But if the test throws and `afterEach` is skipped (can happen under `--bail`), `GSD_HOME` is leaked. Use `try/finally`. This pattern is repeated across multiple test files in this PR — fix all instances.

---

### `model_overrides` added to `GSDConfig` without audit

`model_overrides: Record<string, unknown>` is added to the `GSDConfig` interface with a default of `{}`. Before adding a field to a public interface:
1. Search for every place `GSDConfig` is spread, destructured, or validated — new required fields break exhaustive checks.
2. The field type is `Record<string, unknown>` — the values are agent → model-id strings. Type it properly: `Record<string, string>`.
3. Add a JSDoc comment explaining the semantics (per-agent model override, format: `{ 'gsd-planner': 'claude-opus-4-7' }`).

---

### Required before merge

1. Add `features` to `mergeConfig` and add a corresponding test.
2. Coordinate with #2675: export `mergeConfig` and use it in `config-mutation.ts`, or document the two-path justification in both files.
3. Replace `GSD_HOME` env var with a `homeDir` parameter on `loadGlobalDefaults` for testability.
4. Wrap `afterEach` env restoration in `try/finally` across all modified test files.
5. Type `model_overrides` as `Record<string, string>` and add JSDoc.
6. Remove the `structuredClone` in `mergeConfig` (performance waste; see above).

tomas-lm:
Addressed the requested changes in the latest push:

- Exported and reused `mergeConfig`, including nested merge support for `features` and `model_overrides`.
- Updated `configNewProject` to import `loadUserDefaults`/`mergeConfig` instead of maintaining a separate inline nested-merge allowlist.
- Removed the `GSD_HOME` contract from config loading; `loadConfig`/`loadUserDefaults` now take injectable `homeDir` for tests.
- Wrapped env restoration in `try/finally` in the touched tests.
- Typed `model_overrides` as `Record<string, string>` with JSDoc.
- Removed `structuredClone` from `mergeConfig`.
- Addressed the CodeRabbit test nit by comparing `resolve_model_ids` to `CONFIG_DEFAULTS.resolve_model_ids`.

Validation run locally:

- `git diff --check upstream/main...HEAD` passes.
- `npx vitest run --project unit src/config.test.ts src/query/config-query.test.ts src/query/init.test.ts src/query/config-mutation.test.ts` passes: 104 tests.
- `npm run build` passes.
- Full `npx vitest run --project unit` currently has 5 failures in untouched files: `src/query/state-mutation.test.ts`, `src/query/decomposed-handlers.test.ts`, and `src/query/registry.test.ts`.

tomas-lm:
Addressed CodeRabbit's recursive merge finding in `d14dc9a`:

- `mergeConfig` now recursively merges plain-object children for known config trees instead of only doing a one-level spread.
- Added the deeper config sections CodeRabbit called out (`review`, `claude_md_assembly`, `model_profile_overrides`) plus other dotted config surfaces from the SDK schema (`planning`, `statusline`, `learnings`, `manager`, `intel`, `graphify`).
- Added regression coverage for `review.models.*`, `claude_md_assembly.blocks.*`, and `model_profile_overrides.<runtime>.<tier>` through both `loadConfig` and `configNewProject`.

Validation:
- `git diff --check upstream/main...HEAD` passes.
- `npx vitest run --project unit src/config.test.ts src/query/config-mutation.test.ts src/query/config-query.test.ts src/query/init.test.ts` passes: 105 tests.
- `npm run build` passes.

trek-e (commented):
## Code Review: fix(sdk): honor global model defaults

### 1. The implementation change is not in the diff — so what exactly is being reviewed?

The PR title is "honor global model defaults." The diff shown is `sdk/src/config.test.ts`. Looking at the actual `sdk/src/config.ts` on `main`, `loadUserDefaults()` already exists (lines 133–153) and `loadConfig()` already branches on `!projectConfigFound` to layer user defaults (lines 183–194). If the implementation was already merged separately, say so. If this PR contains the implementation, the diff is incomplete and this review cannot proceed to LGTM.

### 2. `GSD_HOME` isolation is in the current `main` tests — and the PR preserves it correctly

The existing `config.test.ts` (lines 22–23, 30–33) saves/restores both `process.env.HOME` and `process.env.GSD_HOME`. This isolation is already present on `main`. If the PR removes or regresses this, it's a test isolation bug. If it preserves it, state that explicitly in the PR description so reviewers don't have to diff-hunt.

### 3. "project config is authoritative over user defaults" test contradicts CJS parity comment — and may be wrong

The test at line 212–230 asserts that when `.planning/config.json` exists, `~/.gsd/defaults.json` is entirely ignored. The comment says this mirrors "CJS core.cjs:421 (#1683)." But the PR title is "honor global model defaults." If the fix is "only honor them pre-project," that needs to be stated explicitly in the PR description. Right now there is a tension: the PR claims to honor global model defaults, but the test asserts they are ignored once a project config exists. Which is the intended behavior?

### 4. The "recursively merges deeper config trees" test references sections not in the config schema

If the PR adds tests for `review.models`, `claude_md_assembly`, and `model_profile_overrides` as valid top-level sections in `~/.gsd/defaults.json`, those sections need schema acceptance. `config-schema.ts` covers them as dot-notation keys via `DYNAMIC_KEY_PATTERNS`, but `mergeDefaults()` in `config.ts` (lines 213–234) only does section-level merges for `git`, `workflow`, `hooks`, and `agent_skills`. A `review: { models: { claude: 'sonnet' } }` in user defaults would be shallow-merged at the top level, clobbering any existing `review` object in project config. Where is the deep-merge logic for these sections?

### 5. Rubber duck: what was the actual bug?

The issue is #2652: "Codex install can still emit claude mode." The fix PR is "honor global model defaults." The test comment says `resolve_model_ids: "omit"` in `~/.gsd/defaults.json` wasn't being applied in pre-project Codex installs. But the implementation (`loadUserDefaults` + the `!projectConfigFound` branch) is already present on `main`. Either (a) this PR added that implementation and the diff is incomplete, or (b) the bug was something else. The PR description must explain what was broken and what the fix is, not just add tests that pass against existing code.

VERDICT: CHANGES REQUIRED — the diff as presented is not reviewable without the implementation change.

trek-e:

PR #2675, opened 2026-04-24
fix(sdk): propagate nested ~/.gsd/defaults.json keys in config-new-project

### LINKED ISSUE #2673: bug: configNewProject drops nested keys from ~/.gsd/defaults.json
### GSD Version

`1.38.2` (local checkout of `upstream/main` @ `a72bebb` for development; the same code path is shipped in the published `1.38.2` tarball)

### Runtime

N/A — the bug is in `sdk/src/query/config-mutation.ts` and reproduces independently of which AI runtime drives the workflow.

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

v24.x (also reproduced on v25.2.1)

### Shell

N/A — bug is reproducible via direct SDK invocation.

### Installation Method

Local development checkout from `gsd-build/get-shit-done`.

### What happened?

`/gsd-new-project` (via `configNewProject` in the SDK) silently drops every nested key set under `workflow`, `git`, `hooks`, `features`, or `agent_skills` in `~/.gsd/defaults.json`. Only top-level globals (e.g. `model_profile`, `mode`, `granularity`, `commit_docs`) propagate to the new project's `.planning/config.json`.

This contradicts the documented behavior in `docs/CONFIGURATION.md`:

> **Location:** `~/.gsd/defaults.json`. When `/gsd-new-project` creates a new `config.json`, it reads global defaults and merges them as the starting configuration. Per-project settings always override globals.

…and the doc-comment in `sdk/src/query/config-mutation.ts`:

> `Deep merge: hardcoded <- globalDefaults <- userChoices (D11)`

Top-level keys honored that precedence; nested keys did not.

### What did you expect?

Nested keys set in `~/.gsd/defaults.json` should propagate into every new project, with per-project `userChoices` winning on conflict — i.e. the three-way precedence the doc-comment promises (`hardcoded <- globalDefaults <- userChoices`) should apply at every level of the config tree, not only the top.

### Steps to reproduce

1. Create `~/.gsd/defaults.json` (synthetic example, only keys taken from `docs/CONFIGURATION.md`):

   ```json
   {
     "workflow": { "code_review_depth": "deep", "auto_advance": true },
     "git": { "branching_strategy": "milestone" }
   }
   ```

2. `mkdir /tmp/newproj && cd /tmp/newproj`

3. `gsd-sdk query init.new-project '{}'` (or trigger `/gsd-new-project` via the workflow).

4. `cat .planning/config.json`

**Expected** `workflow` and `git` blocks include the global keys:

```json
{
  "workflow": { "code_review_depth": "deep", "auto_advance": true, "...": "..." },
  "git": { "branching_strategy": "milestone", "...": "..." }
}
```

**Actual** — globals dropped:

```json
{
  "workflow": { "code_review_depth": "standard", "auto_advance": false, "...": "..." },
  "git": { "branching_strategy": "none", "...": "..." }
}
```

### Frequency

Always reproducible — pure data flow, no race or environment dependency.

### Impact

Major — machine-wide preferences saved under any of the five nested sections in `~/.gsd/defaults.json` are silently ignored for every new project. Users who set e.g. `workflow.code_review_depth: "deep"` globally still get `"standard"` everywhere; users who set `git.branching_strategy: "milestone"` globally still get `"none"`. The precise keys that propagate vs. drop are non-obvious because top-level keys *do* propagate — making the bug easy to miss until somebody compares the produced config to `defaults.json` manually.

### Workaround

After every `/gsd-new-project`, manually re-apply each nested key from `~/.gsd/defaults.json` into the produced `.planning/config.json`.

### Additional context

**Root cause** in `sdk/src/query/config-mutation.ts` around line 435 of `configNewProject`:

```ts
const config: Record<string, unknown> = {
  ...defaults,
  ...globalDefaults,                  // surfaces nested objects whole
  ...userChoices,
  git: {
    ...(defaults.git as Record<string, unknown>),
    ...((userChoices.git as Record<string, unknown>) || {}),   // globalDefaults.git missing
  },
  workflow: {
    ...(defaults.workflow as Record<string, unknown>),
    ...((userChoices.workflow as Record<string, unknown>) || {}),  // globalDefaults.workflow missing
  },
  hooks: {
    ...(defaults.hooks as Record<string, unknown>),
    ...((userChoices.hooks as Record<string, unknown>) || {}),     // globalDefaults.hooks missing
  },
  agent_skills: {
    ...((defaults.agent_skills as Record<string, unknown>) || {}),
    ...((userChoices.agent_skills as Record<string, unknown>) || {}),  // globalDefaults.agent_skills missing
  },
  features: {
    ...((defaults.features as Record<string, unknown>) || {}),
    ...((userChoices.features as Record<string, unknown>) || {}),       // globalDefaults.features missing
  },
};
```

The `...globalDefaults` on the second line pulls each of `globalDefaults.git`, `.workflow`, `.hooks`, `.agent_skills`, `.features` into `config` as whole objects. The trailing nested-section spreads then overwrite each of those whole objects with `{ ...defaults.<section>, ...userChoices.<section> }`, dropping every key that was present only in `globalDefaults.<section>`.

**Proposed fix:** insert `...globalDefaults.<section>` as the middle spread in each of the five nested blocks, so the documented `hardcoded <- globalDefaults <- userChoices` precedence holds at every level. The change is local, the existing code shape is preserved, and `userChoices` continues to win on conflict.

**Related side observation (separate hardening, not part of the proposed fix):** the existing tests for `configNewProject` in `sdk/src/query/config-mutation.test.ts` (notably `creates config with standard defaults when no global defaults exist` and `creates config.json with defaults`) do not override `process.env.HOME`. They pass in CI only because CI has no `~/.gsd/defaults.json`. On any developer machine that does have one set, both fail today regardless of this fix. The new regression tests written for this bug do override `HOME` to an isolated tmp dir; tightening the existing tests to do the same would be a small follow-up but is out of scope here.

forfrossen:
## Linked Issue

Fixes #2673

## What was broken

`/gsd-new-project` silently dropped every nested key set under `workflow`, `git`, `hooks`, `features`, or `agent_skills` in `~/.gsd/defaults.json` — only top-level globals (e.g. `model_profile`) propagated to the new project's `.planning/config.json`.

## What this fix does

Adds `globalDefaults.<section>` as the middle spread in each of the five nested sections of the merge in `configNewProject`, restoring the documented `hardcoded <- globalDefaults <- userChoices` precedence for nested keys.

## Root cause

The merge in `sdk/src/query/config-mutation.ts` was:

```ts
const config = {
  ...defaults,
  ...globalDefaults,        // surfaces nested objects whole
  ...userChoices,
  workflow: {
    ...defaults.workflow,
    ...userChoices.workflow,   // globalDefaults.workflow missing — clobbered
  },
  // same shape for git / hooks / agent_skills / features
};
```

Each trailing nested-section spread overwrote the whole-object copy that `...globalDefaults` had pulled in at top level.

### How I verified the fix

Added four regression tests under `describe('configNewProject nested global defaults merge', ...)` in `sdk/src/query/config-mutation.test.ts` covering:

1. Nested `workflow.*` keys propagate (`code_review_depth`, `auto_advance`, `node_repair_budget`).
2. Nested `git.*` keys propagate (`branching_strategy`).
3. Nested `hooks.*` and `features.*` keys propagate.
4. Three-way precedence: `userChoices > globalDefaults > hardcoded`.

Each test overrides `process.env.HOME` to an isolated tmp dir so the assertions do not depend on what the developer's actual `~/.gsd/defaults.json` contains.

**Pre-fix:** all 4 new tests fail (e.g. `expected "deep" to be "standard"`).
**Post-fix:** all 4 pass.

**Baseline:** `pnpm test:unit` on `upstream/main @ a72bebb` reports 7 pre-existing failures (5× `state-mutation.test.ts > stateBeginPhase > bug-2420 …`, 2× HOME-mocking gaps in the existing `configNewProject` tests). After this commit the count is unchanged — same 7 failures, no regressions introduced.

### Regression test added?

### Platforms tested

- [x] Linux
- [ ] macOS / Windows — bug is not platform-specific (pure object spread)

### Runtimes tested

- [x] N/A (config-handler is runtime-agnostic)

---

## Checklist

- [x ] Linked issue has the `confirmed-bug` label
- [x] Regression test added
- [x] All tests that passed on `upstream/main` still pass on this branch (see baseline note above)
- [x] CHANGELOG.md updated

## Breaking changes

None — the change makes `~/.gsd/defaults.json` behave the way `docs/CONFIGURATION.md` already promises. Anyone who relied on the buggy "nested global keys are silently ignored" behavior would have done so accidentally; per-project settings still win on conflict via `userChoices`.

### Out of scope (related observation)

The two existing tests `configNewProject global defaults (D11) > creates config with standard defaults when no global defaults exist` and `configNewProject > creates config.json with defaults` do not override `process.env.HOME`. They pass in CI only because CI has no `~/.gsd/defaults.json`. On any developer machine that does have one set, both fail today regardless of this fix. Mentioned in the linked issue (#2673) as a side observation; deliberately not addressed in this PR per the "one concern per PR" rule.

## Release Notes

* **Bug Fixes**
  * Fixed nested configuration key preservation when using global defaults for new projects
  * Improved project-root resolution in multi-repository configurations
  * Eliminated false warnings about stale shell hooks by enhancing version detection


trek-e (changes requested):

**Verdict:** The fix is mechanically correct for the reported bug. Two blocking issues before merge: overlap with #2658 and a test isolation defect.

---

### Rubber-duck: is this solving the right thing?

The root cause is correct: nested-section spreads were two-way (`defaults + userChoices`) while the top-level spread was three-way (`defaults + globalDefaults + userChoices`). The fix adds `globalDefaults.<section>` to each nested spread. This is the right fix.

**However:** PR #2658 (currently open, also pending review) modifies `sdk/src/config.ts:loadConfig` to add a `mergeConfig` helper that performs the same three-way merge for the SDK query path. This PR modifies `sdk/src/query/config-mutation.ts:configNewProject` for the CJS path. If both land independently:

- `loadConfig` has one `mergeConfig` implementation
- `configNewProject` has a separate inline three-way merge
- They use different code for the same documented contract (`hardcoded <- globalDefaults <- userChoices`)

This is the drift that created the original bug. You're re-seeding it. Before this merges, coordinate with #2658: either extract a shared `mergeNestedConfig` helper used by both, or explicitly document that these are intentionally separate paths with separate implementations.

---

### Gall's Law — five independent spreads, not a policy

```ts
git: {
  ...(defaults.git as Record<string, unknown>),
  ...((globalDefaults.git as Record<string, unknown>) || {}),
  ...((userChoices.git as Record<string, unknown>) || {}),
},
// × 5 sections
```

This is five independent inline merge policies. When a sixth nested section is added to `GSDConfig`, the developer adding it will write another two-way spread, and the bug recurs. The PR description quotes the comment `"Deep merge: hardcoded <- globalDefaults <- userChoices (D11)"` — that comment documents the policy but the policy is not enforced. Extract a helper:

```ts
function mergeSection<T>(defaults: T, global: Partial<T>, user: Partial<T>): T {
  return { ...(defaults as object), ...(global as object || {}), ...(user as object || {}) } as T;
}
```

Then each section becomes one line and the policy is impossible to violate accidentally.

---

### Test isolation defect

```ts
beforeEach(async () => {
  ...
  originalHome = process.env.HOME;
  process.env.HOME = homeTmpDir;
});

afterEach(async () => {
  if (originalHome === undefined) {
    delete process.env.HOME;
  } else {
    process.env.HOME = originalHome;
  }
  await rm(homeTmpDir, { recursive: true, force: true });
});
```

If a test throws *before* `afterEach` completes — or if `rm` throws — `process.env.HOME` is leaked to subsequent tests. Use `try/finally` in `afterEach`:

```ts
afterEach(async () => {
  try {
    // restore HOME
  } finally {
    await rm(homeTmpDir, { recursive: true, force: true });
  }
});
```

More critically: `configNewProject` is imported inside each test via `await import('./config-mutation.js')`. Dynamic imports are cached by Node's module system. The module reads `process.env.HOME` at call time (correct), but if the module caches the resolved home path at import time (check the implementation), HOME override has no effect after the first import. Verify this.

---

### Goodhart's Law — tests assert structure not invariant

The four tests verify:
- `code_review_depth` in `workflow.*` propagates ✓
- `branching_strategy` in `git.*` propagates ✓  
- `hooks.*` and `features.*` propagate ✓
- Three-way precedence ✓

Missing: a test verifying that `globalDefaults.git` does NOT bleed into `workflow` (cross-section contamination). This is the type of bug a shallow object spread produces when the wrong section key is mistyped. One negative test: set `git.branching_strategy` in globalDefaults, verify `workflow.branching_strategy` is undefined.

---

### Peter Principle — `configNewProject` reads global defaults every call

`configNewProject` calls `loadGlobalDefaults()` (or equivalent) on every invocation. For a single interactive call this is fine. For batch workspace creation (e.g., `gsd-new-workspace` creating multiple projects), this is a disk read per project. Not a blocker, but add a comment or cache at the call site.

---

### Required before merge

1. Coordinate with #2658: either extract a shared merge helper or explicitly document the two-implementation policy with a comment in both files.
2. Wrap `afterEach` HOME restoration in `try/finally`.
3. Add a cross-section contamination test (negative case).
4. Extract a `mergeSection` helper to enforce the three-way policy structurally, not by convention.

trek-e (commented):
## Code Review: fix(sdk): propagate nested ~/.gsd/defaults.json keys in config-new-project

### 1. The fix is incomplete — `review`, `claude_md_assembly`, and `model_profile_overrides` sections are missing

The actual `configNewProject` in `sdk/src/query/config-mutation.ts` (lines 403–427) now spreads `globalDefaults` for `git`, `workflow`, `hooks`, `agent_skills`, and `features`. But `config-schema.ts` (lines 48, 70, 86–113) explicitly recognizes `review.*`, `claude_md_assembly.*`, and `model_profile_overrides.*` as valid nested config sections. If a user has any of these in `~/.gsd/defaults.json`, they are silently dropped — the top-level `...globalDefaults` spread at line 405 would have captured them as flat keys, but they are objects, and the section-level merges that follow never include them.

You found 5 sections. The schema has 8 dynamic pattern families. Audit the rest.

### 2. Type unsafety: spreading a non-object will produce garbage

At lines 408–426, every spread uses the pattern `...((globalDefaults.git as Record<string, unknown>) || {})`. The `|| {}` guard only fires if `globalDefaults.git` is falsy. If a user misconfigures `~/.gsd/defaults.json` with `"git": "milestone"`, then `globalDefaults.git` is truthy, the spread executes on a string, and you get `{ 0: 'm', 1: 'i', ... }` merged into `git`. The correct guard is:

```ts
...(typeof globalDefaults.git === 'object' && globalDefaults.git !== null
  ? globalDefaults.git as Record<string, unknown>
  : {})
```

This is not hypothetical — user-supplied JSON is unvalidated.

### 3. Inconsistent pattern for `agent_skills` vs. the other four sections

Lines 419–422 spread `defaults.agent_skills` without casting it (`defaults.agent_skills as Record<string, unknown>`), while lines 407–418 explicitly cast `defaults.git`, `defaults.workflow`, and `defaults.hooks`. The inconsistency is unforced; pick one form and apply it uniformly.

### 4. PR #2717 is solving the same problem with a different mechanism

PR #2717 (`feat: workstream config inherits from root .planning/config.json`) adds config inheritance logic. This PR and #2717 both touch config composition without referencing each other. When one of these merges, the other either has a merge conflict or silently double-applies a merge layer. The author needs to coordinate with #2717 or explain why these two PRs do not interact.

### 5. Tests confirm 5 sections pass — that is not the same as "all sections pass"

The test suite validates the sections this PR touches. Goodhart's Law applies: 5 green tests for 5 covered sections is not evidence that the feature works for all sections. The test for `review.models.claude` → global defaults propagation does not exist. Add it, or document explicitly that `review`, `claude_md_assembly`, and `model_profile_overrides` are intentionally excluded (with a reason).

VERDICT: CHANGES REQUIRED

trek-e:

PR #2677, opened 2026-04-24
fix(#2676): phase.complete next-phase lookup respects parallel milestones

### LINKED ISSUE #2676: phase.complete returns is_last_phase: true for projects with parallel milestones
### Summary

When a project has multiple active milestones and `STATE.md`'s `milestone:` frontmatter does not match the milestone that owns the phase being completed, `gsd-sdk query phase.complete <N>` incorrectly returns `is_last_phase: true` and `next_phase: null`. This causes the orchestrator to suggest milestone-end actions instead of the obvious next-phase routing.

### Repro

A project with two active milestones:

```yaml
# .planning/STATE.md frontmatter
milestone: vA
milestone_name: Milestone A
```

ROADMAP.md contains:

```md
### vA Milestone (Active)
### Phase 10: Foo
### Phase 11: Bar

### vB Milestone (Active, Parallel)
### Phase 41.1: Baz
### Phase 41.2: Qux
### Phase 41.3: Quux
```

Run:

```sh
gsd-sdk query phase.complete "41.2"
```

**Expected:** `next_phase: "41.3"`, `is_last_phase: false`.
**Actual:** `next_phase: null`, `is_last_phase: true`.

### Root Cause

`sdk/src/query/phase-lifecycle.ts` (Step E, around line 1085) calls `getMilestonePhaseFilter(projectDir)`, which builds the milestone-phase set from the milestone slice of ROADMAP.md selected by STATE.md's `milestone:` field. Because `41.2` belongs to `vB` but STATE points at `vA`, the filter excludes all `41.x` directories. Both the directory scan and the ROADMAP fallback (which also calls `extractCurrentMilestone`) then produce empty candidate sets, leaving `isLastPhase = true`.

### Why It Matters

Parallel milestones are a legitimate workflow — e.g., a rebrand milestone running alongside a product-feature milestone. STATE.md's single `milestone:` slot makes one of them "primary" but should not invalidate completion routing for the others.

### Workaround

Manually flip STATE.md `milestone:` to match the milestone of each phase before running `phase.complete` (or before running `gsd-execute-phase`, since orchestrator may pick the wrong next-step routing).

### Proposed Fix

Detect the mismatch by checking whether the completed phase itself passes the milestone filter. If not, drop the filter for next-phase detection and scan all phases / the full ROADMAP. The completed phase number is already a parameter to `phase.complete`, so the fix is local to `phase-lifecycle.ts`. PR incoming.

### Environment

- Package: `get-shit-done-cc@1.38.3`
- Node: via `npx` cached install
- Encountered: real project with `vA` (in-progress feature work) and `vB` (parallel rebrand) running concurrently

nickmagnuson2-jpg:
Fixes #2676.

## Summary

When STATE.md's `milestone:` differs from the milestone that owns the phase being completed, `getMilestonePhaseFilter` excludes the completed phase's siblings from the next-phase scan. The handler falsely reports `is_last_phase: true` and `next_phase: null`, causing orchestrators to suggest milestone-end actions instead of obvious next-phase routing.

This is a real workflow: parallel milestones (e.g. a rebrand running alongside feature work) where STATE.milestone makes one "primary" but should not invalidate routing for phases in the other.

## Fix

In `sdk/src/query/phase-lifecycle.ts` Step E (next-phase detection), detect the mismatch by checking whether the completed phase passes the milestone filter:

- **Directory scan:** if the completed phase fails the filter, drop the filter and scan all phase directories.
- **ROADMAP fallback:** if the milestone slice does not contain the completed phase, scan the full ROADMAP instead of the slice.

Single-milestone behavior is unchanged — the relaxed path only fires when the completed phase clearly does not belong to STATE's current milestone. Other call sites of `getMilestonePhaseFilter` (e.g. `buildStateFrontmatter`, milestone progress counts) keep filtering by STATE.milestone since that's the correct semantics for those uses.

## Test Plan

- All 44 existing `phase-lifecycle.test.ts` tests still pass.
- New regression test added: project with two milestone sections in ROADMAP (vA + vB), STATE points at vA, `phaseComplete('41.2')` (a vB phase) returns `next_phase: '41.3'` and `is_last_phase: false`.
- Total: **45/45 phase-lifecycle tests passing**.

## Notes

- The filter is also used in `buildStateFrontmatter`, `phase-lifecycle.ts:1411`, and `state-mutation.ts:549`. This patch only changes behavior in `phase.complete`'s next-phase lookup. The other call sites have different semantics (milestone progress counting) and should keep filtering by STATE.milestone.
- A future enhancement could add an explicit `milestone` parameter to `getMilestonePhaseFilter` so callers can request a specific milestone slice. Out of scope here.

## Release Notes

* **Bug Fixes**
  * Improved phase completion handling to correctly identify subsequent phases when milestone information doesn't match, ensuring accurate detection of final phase status.

* **Tests**
  * Added test coverage for phase completion with parallel milestones to validate next phase detection and final phase identification.

### DIFF
--- a/sdk/src/query/phase-lifecycle.test.ts
+++ b/sdk/src/query/phase-lifecycle.test.ts
@@ -951,6 +951,91 @@ describe('phaseComplete', () => {
     expect(state).toMatch(/Status:\s*Milestone complete/);
   });
 
+  it('finds next phase across parallel milestones (STATE points at different milestone)', async () => {
+    // Regression test: a project with two active milestones where STATE.md
+    // points at milestone vA (e.g. feature work), but the phase being
+    // completed belongs to vB (e.g. parallel rebrand). Without the
+    // parallel-milestone fix, getMilestonePhaseFilter excludes all vB
+    // phases, both the directory scan and ROADMAP fallback find no
+    // candidate, and is_last_phase falsely returns true.
+    const { phaseComplete } = await import('./phase-lifecycle.js');
+
+    const PARALLEL_ROADMAP = `# Roadmap
+
+## Current Milestone: vA Feature Work
+
+| Phase | Plans | Status | Completed |
+|-------|-------|--------|-----------|
+| 9.    | 3/3   | Complete | 2026-04-01 |
+
+- [x] Phase 9: Foundation (completed 2026-04-01)
+
+### Phase 9: Foundation
+
+**Goal:** Build foundation
+**Plans:** 3/3 plans complete
+
+## Parallel Milestone: vB Rebrand
+
+- [x] Phase 41.1: Tokens (completed 2026-04-23)
+- [x] Phase 41.2: Primitives
+- [ ] Phase 41.3: Screens
+
+### Phase 41.1: Tokens
+
+**Goal:** Design tokens
+
+### Phase 41.2: Primitives
+
+**Goal:** UI primitives
+
+### Phase 41.3: Screens
+
+**Goal:** Solver screens
+`;
+
+    const PARALLEL_STATE = `---
+gsd_state_version: 1.0
+milestone: vA
+milestone_name: Feature Work
+status: executing
+progress:
+  total_phases: 1
+  completed_phases: 1
+  total_plans: 3
+  completed_plans: 3
+  percent: 100
+---
+
+# Project State
+
+## Current Position
+
+Phase: 9 of 1 (Foundation)
+Plan: 3 of 3
+Status: Executing
+Last activity: 2026-04-23
+`;
+
+    await setupTestProject(tmpDir, {
+      roadmap: PARALLEL_ROADMAP,
+      state: PARALLEL_STATE,
+      phases: ['09-foundation', '41.1-tokens', '41.2-primitives', '41.3-screens'],
+    });
+
+    // Phase 41.2 has plan + summary so it's "complete" on disk
+    const p412Dir = join(tmpDir, '.planning', 'phases', '41.2-primitives');
+    await writeFile(join(p412Dir, '41.2-01-PLAN.md'), 'plan', 'utf-8');
+    await writeFile(join(p412Dir, '41.2-01-SUMMARY.md'), 'summary', 'utf-8');
+
+    const result = await phaseComplete(['41.2'], tmpDir);
+    const data = result.data as Record<string, unknown>;
+
+    // Despite STATE pointing at vA, completing 41.2 should find 41.3 as next.
+    expect(data.next_phase).toBe('41.3');
+    expect(data.is_last_phase).toBe(false);
+  });
+
   it('collects UAT/VERIFICATION warnings without blocking', async () => {
     const { phaseComplete } = await import('./phase-lifecycle.js');
     await setupTestProject(tmpDir, {
--- a/sdk/src/query/phase-lifecycle.ts
+++ b/sdk/src/query/phase-lifecycle.ts
@@ -1258,10 +1258,19 @@ export const phaseComplete: QueryHandler = async (args, projectDir, workstream) 
   let isLastPhase = true;
 
   try {
+    // Parallel-milestone fix: if STATE.milestone differs from the milestone
+    // that owns the completed phase, the milestone filter excludes sibling
+    // phases and the handler falsely returns is_last_phase: true. Detect the
+    // mismatch by checking whether the completed phase passes the filter; if
+    // not, drop the filter for next-phase detection.
     const isDirInMilestone = await getMilestonePhaseFilter(projectDir, workstream);
+    const completedPhaseInCurrentMilestone = isDirInMilestone(`${phaseNum}-x`)
+      || isDirInMilestone(phaseNum);
     const entries = await readdir(paths.phases, { withFileTypes: true });
-    const dirs = entries.filter(e => e.isDirectory()).map(e => e.name)
-      .filter(isDirInMilestone)
+    const baseDirs = entries.filter(e => e.isDirectory()).map(e => e.name);
+    const dirs = (completedPhaseInCurrentMilestone
+        ? baseDirs.filter(isDirInMilestone)
+        : baseDirs)
       .sort((a, b) => comparePhaseNum(a, b));
 
     for (const dir of dirs) {
@@ -1281,7 +1290,13 @@ export const phaseComplete: QueryHandler = async (args, projectDir, workstream) 
   if (isLastPhase && existsSync(paths.roadmap)) {
     try {
       const roadmapContent = await readFile(paths.roadmap, 'utf-8');
-      const roadmapForPhases = await extractCurrentMilestone(roadmapContent, projectDir);
+      // Same parallel-milestone fix: if the milestone slice does not contain
+      // the completed phase, scan the full ROADMAP instead.
+      const milestoneSlice = await extractCurrentMilestone(roadmapContent, projectDir);
+      const sliceContainsCompletedPhase = new RegExp(
+        `#{2,4}\\s*Phase\\s+${phaseNum.replace(/\./g, '\\.')}\\s*:`, 'i'
+      ).test(milestoneSlice);
+      const roadmapForPhases = sliceContainsCompletedPhase ? milestoneSlice : roadmapContent;
       const phasePattern = /#{2,4}\s*Phase\s+(\d+[A-Z]?(?:\.\d+)*)\s*:\s*([^\n]+)/gi;
       let pm: RegExpExecArray | null;
       while ((pm = phasePattern.exec(roadmapForPhases)) !== null) {


trek-e (changes requested):

**Verdict:** The symptom fix works for the described fixture. The root cause is not addressed. The heuristic has a correctness defect. First-time contributor: being precise.

---

### Rubber-duck: what is the bug, really?

The PR description says: `getMilestonePhaseFilter` excludes completed phase's siblings when STATE.milestone differs from the phase's milestone. 

**Correct symptom, wrong root cause.** The real root cause is: **STATE.milestone is never updated when a user is actively working on a non-primary milestone.** `phaseComplete` is called, STATE.milestone still says `vA`, but the user just completed a `vB` phase. The filter then correctly (by its own contract) excludes `vB` phases.

The PR's fix teaches `phaseComplete` to detect and work around the stale STATE.milestone. It does not fix the stale STATE.milestone. Every other call site — `buildStateFrontmatter`, `phase-lifecycle.ts:1411`, `state-mutation.ts:549` — still operates on a stale `STATE.milestone`. You've patched one symptom site while leaving the underlying state management debt intact and undocumented.

**Question to answer before this merges:** Why does STATE.milestone not update when a user starts working on a parallel milestone? Is this by design? If so, `getMilestonePhaseFilter` is correctly designed and every call site needs the same workaround — which means the fix belongs in `getMilestonePhaseFilter` itself, not scattered across callers.

---

### Correctness defect — the `-x` suffix heuristic

```ts
const completedPhaseInCurrentMilestone = isDirInMilestone(`${phaseNum}-x`)
  || isDirInMilestone(phaseNum);
```

`isDirInMilestone` filters directory names. Phase directories have the form `41.2-primitives`, not `41.2`. The bare `isDirInMilestone(phaseNum)` will never match any real directory (no directory is named `41.2` without a slug). The `${phaseNum}-x` variant is a synthetic directory name you invented — it's not a real directory, it just happens to produce a string that passes the filter's prefix check.

This works only because `getMilestonePhaseFilter` appears to use prefix matching. That's an implementation assumption, not a contract. If the filter ever changes to exact-match or slug-aware matching, this silently breaks with no test catching it.

**The correct approach:** look up the actual directory name for `phaseNum` from the entries already read into `baseDirs`, then check that directory name against the filter. You have `baseDirs` right there — no invented suffixes needed:

```ts
const completedDir = baseDirs.find(d => d.startsWith(phaseNum));
const completedPhaseInCurrentMilestone = completedDir 
  ? isDirInMilestone(completedDir) 
  : false;
```

---

### Goodhart's Law — test fixture inconsistency

The test creates STATE with `total_phases: 1` and `completed_phases: 1` but sets up 4 phase directories (`09-foundation`, `41.1-tokens`, `41.2-primitives`, `41.3-screens`). STATE claims the milestone is fully complete (100%) while there are 3 active phases in a parallel milestone.

A real user in this situation would have STATE.progress reflecting the actual phase count. Your test passes because `phaseComplete` doesn't validate this consistency — but the test fixture is lying about its own invariants. A future developer reading this test to understand parallel-milestone behavior will form a wrong mental model.

---

### ROADMAP fallback regex — boundary defect

```ts
const sliceContainsCompletedPhase = new RegExp(
  `#{2,4}\\s*Phase\\s+${phaseNum.replace(/\./g, '\\.')}\\s*:`, 'i'
).test(milestoneSlice);
```

`Phase 41.2` will match `Phase 41.21:` (leading match, no word boundary). Add `(?=\\s|:)` or use `\\b` after the phase number. This is the kind of regex defect that produces silent wrong routing in production when a project happens to have phases with similar numbering.

---

### Unverified assertion in PR description

> "Other call sites of `getMilestonePhaseFilter` ... keep filtering by STATE.milestone since that's the correct semantics for those uses."

This is asserted but not tested. If the PR claims `buildStateFrontmatter`, `phase-lifecycle.ts:1411`, and `state-mutation.ts:549` are _intentionally_ unpatched, add a comment to each of those call sites explaining why they should NOT get the parallel-milestone relaxation. Otherwise the next developer hits the same bug at a different call site and writes another workaround.

---

### Knuth's Principle — readdir called twice implicitly

`getMilestonePhaseFilter` internally reads from disk or computes from ROADMAP — the PR doesn't show its internals, but `entries` is already fetched via `readdir` at line 1266. Calling `getMilestonePhaseFilter` and then filtering `baseDirs` separately means the milestone data is derived twice (once inside `getMilestonePhaseFilter`, once implicitly via the filter result). If `getMilestonePhaseFilter` does IO, this is a double-read. Consider passing `entries` or refactoring `getMilestonePhaseFilter` to expose the raw milestone pattern rather than a filter function.

---

### Required before merge

1. Fix the heuristic: use `baseDirs.find(d => d.startsWith(phaseNum))` instead of the invented `${phaseNum}-x` suffix.
2. Fix the ROADMAP fallback regex: add a word/sequence boundary after the phase number.
3. Add a comment at each non-patched `getMilestonePhaseFilter` call site explaining why it intentionally keeps the milestone filter.
4. Add a test or comment documenting why `STATE.milestone` is not updated when switching to a parallel milestone — this is the root cause that will produce the next bug report.
5. Fix the test STATE fixture to reflect a consistent real-world state (either update `total_phases` or document the deliberate inconsistency).

trek-e (commented):
## Review — fix(#2676): phase.complete next-phase lookup respects parallel milestones

### The diff in the PR does not match the code that was actually committed

The PR diff shows this logic:

```ts
const completedPhaseInCurrentMilestone = isDirInMilestone(`${phaseNum}-x`)
  || isDirInMilestone(phaseNum);
```

The actual committed code in `sdk/src/query/phase-lifecycle.ts` (lines 1275-1280) is:

```ts
const completedDirInFilter = allDirs.some((d) => {
  const dm = d.match(/^(\d+[A-Z]?(?:\.\d+)*)-?/i);
  return dm && comparePhaseNum(dm[1], phaseNum) === 0 && isDirInMilestone(d);
});
completedPhaseInPrimaryMilestone = completedDirInFilter;
const effectiveFilter = completedDirInFilter ? isDirInMilestone : (_d: string) => true;
```

These are meaningfully different implementations. The PR description and diff describe an approach that no longer exists in the submitted code. Did this PR go through a rebase or amendment that replaced the described fix with a different one? If so, the PR description, test comments, and changelog all describe a different implementation than what's being reviewed. Update the PR description to match the actual code, or clarify why the implementation changed.

### The original `isDirInMilestone('${phaseNum}-x')` approach (from the diff) is broken

Reviewing the implementation described in the diff, since that's what was submitted for review: `isDirInMilestone` takes a directory name and extracts the phase number using the regex `^0*(\d+[A-Za-z]?(?:\.\d+)*)`. If `phaseNum` is `41.2`, then:

- `isDirInMilestone('41.2-x')` extracts `41.2` from `41.2-x` via the numeric match. This might actually work by coincidence — `41.2` is extracted and checked against the milestone's normalized set.
- `isDirInMilestone('41.2')` passes `41.2` as a bare string. The regex `^0*(\d+[A-Za-z]?(?:\.\d+)*)` would match `41.2` and extract `41.2`. Also might work.

But both of these bypass the actual check that matters: **does a real directory named something like `41.2-primitives` exist on disk and pass the filter?** The `-x` suffix is not a real directory. You're abusing the filter function by passing synthetic directory names rather than doing what the committed code correctly does: scan `allDirs` for a real directory whose numeric prefix matches `phaseNum`.

The committed code's approach (`allDirs.some(...)`) is correct. The diff's approach is wrong. If the PR was rewritten to use the better approach, that's good — but it means the PR description is reviewing the wrong code.

### Regex injection is partially mitigated but not eliminated

The ROADMAP fallback uses:

```ts
new RegExp(`#{2,4}\\s*Phase\\s+${phaseNum.replace(/\./g, '\\.')}\\s*:`, 'i')
```

`phaseNum` comes from user-controlled input (the phase argument parsed from `$ARGUMENTS`). The `.replace` escapes literal dots, but does not escape: `^`, `$`, `[`, `]`, `(`, `)`, `*`, `+`, `?`, `{`, `}`, `|`, `\`. A phase number like `41[2` would produce an invalid regex and throw. A phase number like `41(2` would produce a regex with an unintended capture group. Use a proper `escapeRegex` function. The fix is one line; the exposure surface is the entire phase argument parsing path.

### The test may pass for the wrong reason — verify `setupTestProject` creates physical directories

The test creates phases `['09-foundation', '41.1-tokens', '41.2-primitives', '41.3-screens']` via `setupTestProject`. Looking at `setupTestProject` (line 112-116 of the test file): it does `mkdir(join(phasesDir, phase), { recursive: true })` and writes a `.gitkeep`. So physical directories do exist for all four phases.

However: STATE.md in the test fixture sets `milestone: vA` with only Phase 9. The test completes `41.2` and expects `next_phase: '41.3'`. With the fix, the filter is bypassed because `41.2-primitives` is not in the vA milestone filter. The directory scan finds `41.3-screens` and returns it. This is correct.

But here is the gap: what if `41.3-screens` did NOT exist as a physical directory? The code would fall through to the ROADMAP text scan. The test does not cover the case where the next phase exists in ROADMAP but not on disk. That is the "not yet scaffolded" fallback path. Add a test case where `41.3-screens` is absent from disk but present in ROADMAP, to verify the ROADMAP regex path also works correctly under parallel milestones.

### The "drop filter entirely" fix is broader than necessary

When completing a phase from a parallel milestone, the code drops the `isDirInMilestone` filter and scans **all** phase directories. This means if a project has three milestones (vA: phases 1-5, vB: phases 41.x, vC: phases 99.x) and the user completes vB phase 41.2, the next-phase scan will see phases from vA and vC as candidates. `comparePhaseNum` will return the next numerically greater phase across *all* milestones, not necessarily the next phase in vB.

The correct fix would be: derive the milestone from the phase being completed (scan ROADMAP to find which milestone section contains `Phase 41.2:`) and filter by *that* milestone. This PR instead drops filtering entirely, which is overly broad. It works for the two-milestone case in the test but is incorrect for projects with three or more active milestones where phase numbering is not globally monotonic.

---

**VERDICT: CHANGES REQUIRED**

Three blockers:
1. Reconcile the PR description with the actual committed implementation — one of them is wrong.
2. Fix the regex injection: replace the ad-hoc `.replace(/\./g, '\\.')` with a proper escapeRegex call covering all metacharacters.
3. Add a test case for the ROADMAP-only path (next phase in ROADMAP but not on disk) under parallel milestones, and address the three-milestone case or explicitly document the known limitation.

trek-e:

PR #2711, opened 2026-04-25
fix(install): harden leaked .claude path scanning for markdown and comments

### LINKED ISSUE #2470: Unreplaced .claude path reference(s) in files when install GSD for Gemini runtime.
### GSD Version

1.38.1

### Runtime

Gemini CLI

### Operating System

Windows

### Node.js Version

24.13.0

### Shell

PowerShell PSVerion 5.1.19041.7181

### Installation Method

### What happened?

After installation GSD in Gemini runtime, by using 'npx get-shit-done-cc@latest' I have information about unreplaced .claude path.

### What did you expect?

Replace correctly all path references appropriately to installation runtime(s) environment. 

### Steps to reproduce

1. Install GSD with 'npx get-shit-done-cc@latest'
2. Select runtime: Gemini
3. Select place to install: 'Local'
4. Error appear at this step (in installation report)

### Error output / logs

```shell
⚠  Found 1 unreplaced .claude path reference(s) in 1 file(s):
     D:\Projects\MsProjectExporter\.gemini\get-shit-done\workflows\update.md (1)
  These paths may not resolve correctly for Gemini.
```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

### Workaround (if any)

_No response_

### Additional context

Image

### Privacy Checklist

HouJia:
## Summary
- Restrict leaked `.claude` detection to executable markdown regions by scanning fenced and indented code blocks only.
- Filter `#`, `//`, and `/* ... */` comments before matching so documentation/examples do not trigger false positives while preserving real path hits.
- Add focused regression coverage for scan scope + comment filtering and align `update.md` guard with `detectClaudePathRefs()`.

Fixes #2470

## Test plan
- [x] `node --test tests/install-leaked-claude-paths-filter.test.cjs tests/bug-2470-update-md-claude-path.test.cjs`
- [x] `npm test`

Made with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate detection of unreplaced ".claude" paths: scans only code sections for Markdown, scans full content otherwise, and filters common comment styles to reduce false positives; leak records only when actual matches are found.

* **Tests**
  * Added regression tests validating Markdown vs non‑Markdown scanning, comment filtering, and edge-case path detection.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -6216,8 +6216,9 @@ function install(isGlobal, runtime = 'claude') {
             }
             throw err;
           }
-          const matches = content.match(/(?:~|\$HOME)\/\.claude\b/g);
-          if (matches) {
+          const isMarkdown = entry.name.endsWith('.md');
+          const matches = detectClaudePathRefs(content, isMarkdown);
+          if (matches.length > 0) {
             leakedPaths.push({ file: fullPath.replace(targetDir + '/', ''), count: matches.length });
           }
         }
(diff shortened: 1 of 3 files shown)


HouJia:
Addressed this feedback in 908a5966.

- `_stripSingleLineComments()` now strips inline `#` / `//` comment tails (not only full-line comments).
- Added regression tests for inline single-line comments in both non-markdown and markdown code-block scans.
- Re-ran targeted regressions: `node --test tests/install-leaked-claude-paths-filter.test.cjs tests/bug-2470-update-md-claude-path.test.cjs` (pass).

trek-e (commented):
## Review — fix(install): harden leaked .claude path scanning for markdown and comments

### Kernighan's Law Is the Primary Concern

The original code was one line:
```js
content.match(/(?:~|\$HOME)\/\.claude\b/g)
```

The replacement is ~200 lines of hand-rolled markdown parser with 10+ helper functions: `_findFenceOpen`, `_isMarkdownFenceClose`, `_extractCodeBlocks`, `_findSingleLineCommentStart`, `_stripInlineComment`, `_findBlockCommentStart`, `_findBlockCommentEnd`, `_stripBlockComments`, `_stripComments`, `detectClaudePathRefs`.

Kernighan's law: debugging is twice as hard as writing the code. This PR's parser is already hard to write correctly (evidence: the issues below). It will be approximately four times harder to debug when it produces a false negative in production and someone's `.claude` path leaks undetected.

The fundamental question this PR does not answer: **what specific production failure occurred that justified this complexity?** The PR cites issue #2470, but doesn't quote a single real-world example of the one-liner producing a false positive that caused an actual installation problem. "Documented examples in a README caused false positives" is the implied story — but how many real installations were affected? What was the measured false-positive rate? Without this, the complexity is a solution in search of a measured problem.

### Gall's Law: Working Simple System → Complex Fragile System

The regex worked for the actual threat model: detecting literal `~/.claude` or `$HOME/.claude` strings in config files. Markdown documentation files are not config files. The correct fix for "README examples trigger the scanner" is to **not scan markdown files for `.claude` path leaks** — a one-line file extension check — not to write a CommonMark parser.

Instead this PR writes a parser that now has its own bugs (see below) and must be maintained indefinitely. Every time someone writes a pathological markdown file, this parser will need updating. The simple system had one attack surface; this system has ten.

### `_findSingleLineCommentStart` — Hash Detection Is Broken

```js
hashIdx === 0 || (hashIdx > 0 && /\s/.test(line[hashIdx - 1]))
```

This logic intends to detect shell-style `#` comments by checking the character before `#`. But it will incorrectly strip content in cases like:

- `url: "https://example.com/#fragment"` — the `#` follows `/`, not whitespace, so it's not stripped. Fine. But:
- `path: /home/user #comment` — stripped correctly.
- `path: "#not-a-comment"` — the `#` is at index 6 preceded by a space inside a quoted string. This strips everything after the space-hash, producing `path: "` — a mangled value that now doesn't contain `~/.claude` even if the original did.

The `#` detection has no concept of string context. In YAML, TOML, and shell configs (the file types this scanner actually runs on), `#` inside a quoted value is not a comment. This can produce false negatives: a real `.claude` leak inside a quoted string with a `#` before it will be stripped and missed.

### `GSD_TEST_MODE` Export Coupling

The function is exported conditionally:
```js
if (process.env.GSD_TEST_MODE) {
  module.exports._detectClaudePathRefs = detectClaudePathRefs;
}
```

This couples test infrastructure to production code. If `GSD_TEST_MODE` is accidentally set in a production environment (a CI system, a user's dotfiles, a Docker image with a stale env var), the internal function is exported and callable by any code that requires this module. More importantly, **tests that depend on `GSD_TEST_MODE` being set are testing a different code path than production** — if the export guard ever has a bug, the tests won't catch it because the export is what they're testing. Export the function unconditionally or put it in a separate testable module.

### Knuth's Optimization Principle

No benchmark is provided showing the original regex caused false positives at meaningful scale. The PR description says "Fixes #2470" — what does issue #2470 actually report? How many users hit it? Was it a single edge case in one repository's markdown, or a widespread problem? Without measurement, 200 lines of parser complexity is premature. The optimization should have been: (1) measure the false-positive rate, (2) implement the simplest fix that addresses measured cases, (3) re-measure.

### CommonMark Edge Cases Not Covered

The PR's test plan is two specific test files. There is no evidence the author tested against:

1. **Nested fences**: A `~~~` block inside a `` ``` `` block. CommonMark spec says fences don't nest, but the parser needs to handle the case where a `~~~` line appears inside a backtick fence — it should be treated as content, not a fence opener.
2. **Indented code blocks**: CommonMark allows 4-space indented blocks without fence markers. `_extractCodeBlocks` appears to handle `indented` type but the interaction with the fence-scanning logic is not obvious.
3. **ATX headings with code**: `## See \`~/.claude\`` — the backtick span is inline code, not a fenced block. The scanner would see this as plaintext (not in a code block) and filter or match it depending on comment stripping. Is this intended?
4. **HTML blocks within markdown**: `<pre>~/.claude</pre>` is a valid markdown HTML block that would render as code. The parser doesn't handle HTML blocks.

### Peter Principle Assessment

This contributor is implementing a CommonMark-adjacent parser without demonstrating familiarity with the CommonMark spec. The fence detection, while mostly correct, misses the nested-fence interaction. The comment stripping has a quoted-string blindspot. The test coverage is narrow. Writing a correct markdown parser is a solved problem (use `remark`, `unified`, or even a simple fence-only extractor from a tested library). The choice to hand-roll it suggests the author didn't look for existing solutions.

---

**VERDICT: CHANGES REQUIRED**

This PR makes the codebase harder to maintain without demonstrating the complexity is warranted. Minimum before merge: (1) cite the actual production failure from #2470 with a measured false-positive count, (2) fix the `#`-in-quoted-string false-negative in comment stripping, (3) remove the `GSD_TEST_MODE` conditional export, (4) add test cases for indented code blocks, ATX headings with inline code, and nested fence characters, (5) document why a library like `remark` was rejected in favor of a hand-rolled parser.

trek-e:

PR #2717, opened 2026-04-25
feat: workstream config inherits from root .planning/config.json

### LINKED ISSUE #2714: feature: workstream config.json should inherit from root .planning/config.json
**Current behavior**
When `GSD_WORKSTREAM` is set, `loadConfig()` reads only `.planning/workstreams/<ws>/config.json`. If that file doesn't exist (or doesn't define a key), the resolver falls back to `CONFIG_DEFAULTS` — NOT to root `.planning/config.json`.

**Pain point**
With multiple workstreams (e.g., `junny`, `henry`, `junny-p1`, `junny-p2`, `junny-p3`), settings like `model_overrides`, `model_profile`, `workflow.*`, `context_window`, `features.thinking_partner` must be duplicated in each workstream config. Adding a new workstream means a new copy. A team of 4-5 devs each running 2-3 parallel workstreams = 10-15 copies of the same config drift slowly out of sync.

**Repro**
1. Set `model_overrides` for all GSD agents in root `.planning/config.json`
2. `GSD_WORKSTREAM=henry gsd-tools resolve-model gsd-pattern-mapper` -> returns `sonnet` (uses defaults, ignores root override)
3. Without env var: `gsd-tools resolve-model gsd-pattern-mapper` -> returns `opus` (root override applies)

**Proposed**
On `loadConfig()` when `GSD_WORKSTREAM` is set:
1. Read root `.planning/config.json`
2. Read `.planning/workstreams/<ws>/config.json` (if exists)
3. Deep-merge: workstream wins on key conflict; otherwise inherit root.

Matches the typical config-inheritance pattern (npm, eslint, tsconfig).

**Backwards compat**
Workstreams that already have a full standalone config keep working — they just stop falling through to defaults for unset keys, instead falling through to root. Workstream-specific overrides still take precedence.

**Affected keys (most painful to duplicate)**
- `model_overrides` (32+ agent → model mappings)
- `workflow.*` (20+ flags: research, plan_check, ui_phase, security_*, code_review, etc.)
- `context_window`
- `features.*`
- `manager.flags.*`

**Workaround in use locally**
Patched `loadConfig()` in `bin/lib/core.cjs` to deep-merge root -> workstream. Verified across 4 workstreams (henry, junny, junny-p2, junny-p3) — all now inherit root `model_overrides` correctly. Happy to upstream the patch as a PR if direction approved.

Version: 1.38.3

junny-l:
## Summary

When \`GSD_WORKSTREAM\` is set, \`loadConfig()\` previously read only \`.planning/workstreams/<ws>/config.json\`. If that file was missing (or didn't define a key), the resolver fell back to \`CONFIG_DEFAULTS\` — never to root \`.planning/config.json\`. This forced users with multiple workstreams to duplicate every setting (\`model_overrides\`, \`workflow.*\`, \`context_window\`, \`features.*\`, \`manager.*\`) across every workstream config.

This PR makes workstream configs deep-merge over root config:

- **Root \`.planning/config.json\`** = project-wide defaults (single source)
- **Workstream \`config.json\`** = per-workstream overrides (overlay)
- **Workstream wins** on key conflict; unset keys **inherit from root**
- **Missing workstream config** falls back to root entirely

## Pain point this fixes

Real-world example: a team of 4-5 devs each running 2-3 parallel workstreams ends up with 10-15 copies of the same config drifting slowly out of sync. Adding a new workstream means a new copy. Updating \`model_overrides\` to add a new agent means editing every workstream copy by hand.

After this PR: set \`model_overrides\` once at root → every workstream picks it up automatically. Workstream-specific overrides remain (e.g., a workstream that wants a different \`code_review_depth\`).

## Diff

\`\`\`diff
+function _deepMergeConfig(base, overlay) {
+  if (overlay === null || overlay === undefined) return base;
+  if (base === null || base === undefined) return overlay;
+  if (typeof base !== 'object' || typeof overlay !== 'object') return overlay;
+  if (Array.isArray(base) || Array.isArray(overlay)) return overlay;
+  const out = { ...base };
+  for (const k of Object.keys(overlay)) {
+    out[k] = _deepMergeConfig(base[k], overlay[k]);
+  }
+  return out;
+}

 function loadConfig(cwd) {
   const configPath = path.join(planningDir(cwd), 'config.json');
+  const rootConfigPath = path.join(planningRoot(cwd), 'config.json');
   const defaults = CONFIG_DEFAULTS;

+  // Inherit-from-root (#2714): when GSD_WORKSTREAM is active, read root
+  // and workstream configs, deep-merge with workstream-wins.
+  let inheritedRaw = null;
+  if (configPath !== rootConfigPath && process.env.GSD_WORKSTREAM) {
+    let rootObj = null;
+    let wsObj = null;
+    try { rootObj = JSON.parse(fs.readFileSync(rootConfigPath, 'utf-8')); } catch {}
+    try { wsObj = JSON.parse(fs.readFileSync(configPath, 'utf-8')); } catch {}
+    if (rootObj && wsObj) {
+      inheritedRaw = JSON.stringify(_deepMergeConfig(rootObj, wsObj));
+    } else if (rootObj && !wsObj) {
+      inheritedRaw = JSON.stringify(rootObj);
+    }
+  }

   try {
-    const raw = fs.readFileSync(configPath, 'utf-8');
+    const raw = inheritedRaw || fs.readFileSync(configPath, 'utf-8');
     const parsed = JSON.parse(raw);
\`\`\`

## Verification

Tested locally with 4 workstreams (\`junny\`, \`henry\`, \`junny-p2\`, \`junny-p3\`):

\`\`\`
ws=junny    profile: quality, model_overrides count: 32
ws=henry    profile: quality, model_overrides count: 32
ws=junny-p2 profile: quality, model_overrides count: 32
ws=junny-p3 profile: quality, model_overrides count: 32
\`\`\`

Before this PR: each workstream returned \`balanced\` profile + 0 overrides because their workstream-local config either didn't exist or was a stripped-down copy.

## Backwards compatibility

- Workstreams with full standalone configs keep working — they just stop falling through to \`CONFIG_DEFAULTS\` for unset keys, and instead fall through to root. Concrete keys still set in workstream config still win.
- Workstreams with no config now correctly inherit root settings instead of defaulting to \`balanced\` model_profile, no \`model_overrides\`, etc.
- No new files. No new dependencies. Pure read-merge — no on-disk mutation.

## Closes

Closes #2714

## Summary by CodeRabbit

* **New Features**
  * Workstream-specific configuration overrides now supported. When workstreams are enabled, per-workstream settings automatically take precedence over global settings with intelligent fallback handling.

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -269,12 +269,50 @@ const CONFIG_DEFAULTS = {
   post_planning_gaps: true, // workflow.post_planning_gaps — unified post-planning gap report (#2493): scan REQUIREMENTS.md + CONTEXT.md decisions vs all PLAN.md files
 };
 
+/**
+ * Deep-merge two config objects. `overlay` wins on key conflict; nested
+ * objects are merged recursively. Arrays and primitives in `overlay`
+ * replace the corresponding value in `base` entirely (no array merge).
+ *
+ * Used by loadConfig to inherit root .planning/config.json into a
+ * workstream config when GSD_WORKSTREAM is set (#2714).
+ */
+function _deepMergeConfig(base, overlay) {
+  if (overlay === null || overlay === undefined) return base;
+  if (base === null || base === undefined) return overlay;
+  if (typeof base !== 'object' || typeof overlay !== 'object') return overlay;
+  if (Array.isArray(base) || Array.isArray(overlay)) return overlay;
+  const out = { ...base };
+  for (const k of Object.keys(overlay)) {
+    out[k] = _deepMergeConfig(base[k], overlay[k]);
+  }
+  return out;
+}
+
 function loadConfig(cwd) {
   const configPath = path.join(planningDir(cwd), 'config.json');
+  const rootConfigPath = path.join(planningRoot(cwd), 'config.json');
   const defaults = CONFIG_DEFAULTS;
 
+  // Inherit-from-root (#2714): when GSD_WORKSTREAM is active, read the root
+  // config and the workstream config, deep-merge with workstream-wins. If the
+  // workstream config is missing, fall back to root entirely. Workstream
+  // settings override root for shared keys; unset keys inherit from root.
+  let inheritedRaw = null;
+  if (configPath !== rootConfigPath && process.env.GSD_WORKSTREAM) {
+    let rootObj = null;
+    let wsObj = null;
+    try { rootObj = JSON.parse(fs.readFileSync(rootConfigPath, 'utf-8')); } catch { /* no root config */ }
+    try { wsObj = JSON.parse(fs.readFileSync(configPath, 'utf-8')); } catch { /* no ws config */ }
+    if (rootObj && wsObj) {
+      inheritedRaw = JSON.stringify(_deepMergeConfig(rootObj, wsObj));
+    } else if (rootObj && !wsObj) {
+      inheritedRaw = JSON.stringify(rootObj);
+    }
+  }
+
   try {
-    const raw = fs.readFileSync(configPath, 'utf-8');
+    const raw = inheritedRaw || fs.readFileSync(configPath, 'utf-8');
     const parsed = JSON.parse(raw);
 
     // Migrate deprecated "depth" key to "granularity" with value mapping


trek-e:
Pushed fixes for both CodeRabbit findings on top of the PR branch.

## Bug 1 — `null` merge not propagating

`_deepMergeConfig` treated `null === undefined`, so a workstream config setting a key to `null` was silently discarded and the root's value bled through instead.

**Fix:** split the guard — only `undefined` short-circuits to `base`; `null` returns `null` so `loadConfig`'s `??` falls back to the system default rather than inheriting root's explicit value.

```diff
-  if (overlay === null || overlay === undefined) return base;
+  if (overlay === undefined) return base;
+  if (overlay === null) return null;
```

## Bug 2 — `rootConfigPath` ignores `GSD_PROJECT`

`planningRoot(cwd)` always resolves to `.planning/` with no project awareness. When both `GSD_PROJECT` and `GSD_WORKSTREAM` are set, the merge reads config from the wrong project's root.

**Fix:** use `planningDir(cwd, null)` — strips the workstream segment but preserves project scope.

```diff
-  const rootConfigPath = path.join(planningRoot(cwd), 'config.json');
+  const rootConfigPath = path.join(planningDir(cwd, null), 'config.json');
```

## Tests

4 regression tests added to `tests/core.test.cjs` under `describe('loadConfig workstream config inheritance (#2714)')`. All 174 core tests pass.

trek-e (commented):
## Review — feat: workstream config inherits from root .planning/config.json

### Rubber Duck: Do you understand what you're fixing?

The PR description says workstream configs previously \"fell back to `CONFIG_DEFAULTS` — never to root `.planning/config.json`\". That's a valid pain point. But the fix activates *only* when `GSD_WORKSTREAM` is set. If a user runs `loadConfig()` directly in a workstream directory *without* that env var, they still get no inheritance. The guard condition means the behaviour depends on runtime environment, not on directory structure. That's a hidden coupling that will bite someone in a subshell, a cronjob, or any process that doesn't inherit the parent's env. The PR never justifies why env-var-gating is the right trigger rather than, say, detecting that `cwd` is inside a workstream directory.

### Null-Overlay Bug

```js
if (overlay === null || overlay === undefined) return base;
```

`null` is a valid JSON value. If a workstream config sets `"features": null` to explicitly disable features inherited from root, this merge treats that as \"no overlay\" and returns the base value — silently ignoring the intent. The author hasn't demonstrated they understand that JSON `null` is semantically distinct from \"key absent\". This is not a theoretical concern: any user who tries to zero out a root key will get no error and no effect, just the inherited value they were trying to override. The fix is one line: remove the `null` branch from the guard and let `null` propagate as a valid overlay.

### `rootConfigPath` Derivation Bug (flagged by CodeRabbit, not addressed)

```js
const rootConfigPath = path.join(planningRoot(cwd), 'config.json');
```

`planningRoot(cwd)` is documented to walk up from `cwd` to find the nearest `.planning` directory. In a project-scoped run where `cwd` is already the project root, this is fine. In a monorepo where a workstream's `cwd` is deep inside a sub-package, `planningRoot` may resolve to the sub-package's `.planning`, not the repo root's. Result: `rootConfigPath === configPath` and the inheritance branch is silently skipped. The author's verification ran four workstreams in a single flat repo — this does not cover the monorepo case. What is `planningRoot`'s documented contract for nested projects?

### Gall's Law: You Replaced a Simple Working System

Before this PR: workstream config stands alone. Predictable. Debuggable. Auditable with a single `cat`.

After this PR: what a workstream "sees" depends on (a) the workstream config file, (b) the root config file, (c) whether `GSD_WORKSTREAM` is in the environment, and (d) whether `planningRoot` resolves the paths differently than expected. That's four invisible inputs. When a workstream behaves unexpectedly, the debugging path is now: check workstream config → check root config → check env → check path resolution. There is no `--dry-run` or `--show-merged-config` flag added here. There is no rollback path other than reverting this commit. Where is the escape hatch?

### Goodhart's Law: The Path Comparison Is Fragile

```js
if (configPath !== rootConfigPath && process.env.GSD_WORKSTREAM) {
```

This is a string equality check on filesystem paths. On macOS (the primary dev platform for this repo, which defaults to case-insensitive HFS+/APFS), `/Users/foo/.planning/config.json` and `/Users/FOO/.planning/config.json` are the same file but would pass this check. Symlinks are worse: if `.planning` is a symlink, `path.join` won't resolve it, and you can get two different string paths pointing to the same physical file. The inheritance branch would then merge a config with itself — not catastrophic, but not correct, and the author hasn't thought about it. Use `fs.realpathSync` or `path.resolve` before comparing.

### Root Cause Analysis: Is This The Right Fix?

Issue #2714 is about config duplication across workstreams. The root cause is that workstream setup tooling (`gsd-workstreams`, `gsd-new-workspace`, etc.) creates new workstream config files with no reference to root config. The structural fix would be: make those setup tools write minimal workstream configs by default, not copy full root config. This PR adds runtime merge complexity to work around a setup-time authoring problem. That is a symptom fix. If the setup tooling is fixed, the inheritance merge becomes unnecessary overhead.

### Array Semantics Are Undocumented

```js
if (Array.isArray(base) || Array.isArray(overlay)) return overlay;
```

Arrays are replaced wholesale, not merged. A root config with `"model_overrides": [32 entries]` and a workstream config with `"model_overrides": ["gpt-4o"]` will yield `["gpt-4o"]` — not 33 entries. This is a defensible design choice. It is not documented anywhere in the PR, the code comment, or the `CLAUDE.md`. Users will expect array merge. This needs a doc comment at minimum.

### Peter Principle Flag

The PR adds a recursive merge function without handling: (1) null clobber semantics (above), (2) array merge vs replace tradeoffs (above), (3) path comparison correctness (above), (4) the env-var activation gap, (5) a way to introspect the merged result. None of these are exotic edge cases — they are the standard checklist for any \"config inheritance\" feature. The verification section shows four workstreams with identical counts, which proves the happy path works. It does not prove any of the above cases are handled.

---

**VERDICT: CHANGES REQUIRED**

Minimum before merge: (1) fix null-overlay to treat `null` as a valid override value, (2) fix path comparison to use `fs.realpathSync`, (3) document array-replace semantics in a code comment, (4) address the `planningRoot` monorepo case or explicitly document the limitation, (5) explain why env-var gating is the right trigger.

trek-e:

Issue #2645, opened 2026-04-24
GSD 1.38.3 writes invalid Codex config.toml agents table: invalid type map, expected a sequence


bigwhite11:
### GSD Version

1.38.3

### Runtime

Codex

### Operating System

macOS

### Shell

zsh

### Installation Method

```bash
npx -y get-shit-done-cc@latest --codex --global
```

Also reproduced with a local Codex install under a project directory using `--codex --local`.

### What happened?

After installing/updating GSD for Codex, Codex failed to restore/load a conversation because the generated Codex config was invalid.

The global install produced this error:

```text
failed to load configuration: /Users/<user>/.codex/config.toml:199:1: invalid type: map, expected a sequence
```

A local install produced the same class of error from a project-local config:

```text
failed to load configuration: /Users/<user>/.openclaw/.codex/config.toml:93:1: invalid type: map, expected a sequence
```

### Relevant generated config

GSD generated agent config as TOML tables:

```toml
[agents.gsd-planner]
description = "Creates executable phase plans with task breakdown, dependency analysis, and goal-backward verification. Spawned by /gsd-plan-phase orchestrator."
config_file = "/Users/<user>/.codex/agents/gsd-planner.toml"

[agents.gsd-executor]
description = "Executes GSD plans with atomic commits, deviation handling, checkpoint protocols, and state management. Spawned by execute-phase orchestrator or execute-plan command."
config_file = "/Users/<user>/.codex/agents/gsd-executor.toml"
```

Codex appears to expect `agents` to be a sequence/list rather than a TOML map/table, so `[agents.gsd-*]` causes config parsing to fail.

### Steps to reproduce

1. Install or update GSD for Codex:

```bash
npx -y get-shit-done-cc@latest --codex --global
```

2. Start or restore a Codex session.

3. Codex fails while loading `config.toml`.

A local install can also reproduce the same issue:

```bash
cd /some/project
npx -y get-shit-done-cc@latest --codex --local
```

Then Codex fails while reading:

```text
/some/project/.codex/config.toml
```

### Expected behavior

GSD should generate Codex-compatible config, or avoid writing agent definitions in a format Codex cannot parse.

Codex should continue to start normally after GSD installation.

### Actual behavior

Codex fails to load configuration with:

```text
invalid type: map, expected a sequence
```

### Workaround

Uninstalling GSD from Codex fixes the config-loading failure.

Global uninstall:

```bash
npx get-shit-done-cc --codex --global --uninstall
```

[...]

PR #2730, opened 2026-04-26
fix(codex): use current agent config schema and hooks.json

Minghou-Lei:
# Title

fix(codex): use current agent config schema and move update hook to hooks.json

# Summary

- switch Codex agent registry generation from broken `[[agents]]` array-of-tables to current `[agents.gsd-*]` subtables
- stop writing the invalid Codex `[[hooks]]` SessionStart block into `config.toml`
- register the Codex update hook in `hooks.json` instead, while preserving user hook entries
- keep cleanup/self-heal logic for previously shipped broken `[[agents]]` / `[[hooks]]` installs

# Root Cause

Current GSD Codex installs generate:

- `[[agents]]` entries in `config.toml`
- `[[hooks]]` SessionStart entries in `config.toml`

On current Codex builds, this breaks parsing:

- `[[agents]]` triggers `invalid type: sequence, expected struct AgentsToml`
- `[[hooks]]` triggers `invalid type: map, expected a sequence`

I reproduced this against a live local Codex install by pointing `CODEX_HOME` at isolated test configs and running `codex features list`.

# What Changed

## Agents

Generated Codex config now uses:

```toml
[agents.gsd-executor]
description = "..."
config_file = "/absolute/path/to/agents/gsd-executor.toml"
```

The installer still strips previously shipped broken `[[agents]]` entries on reinstall/uninstall.

## Hooks

The installer now:

- keeps enabling `codex_hooks = true`
- removes previously shipped legacy `[[hooks]]` GSD update blocks from `config.toml`
- writes the GSD SessionStart update hook to `hooks.json`

This preserves user hooks while avoiding invalid Codex TOML hook syntax.

## Tests

Updated Codex tests to cover:

- `[agents.gsd-*]` generation
- self-healing of shipped broken `[[agents]]`
- `hooks.json` install behavior
- `hooks.json` uninstall cleanup that removes only GSD hooks and preserves user hooks

# Verification

- `node --test tests/codex-config.test.cjs tests/install-hooks-copy.test.cjs`
- live parser smoke with isolated `CODEX_HOME` and `codex features list`

# Notes

- A smoke install from the git clone still exits non-zero if `sdk/dist/cli.js` has not been built locally. That pre-existing SDK build issue is separate from this Codex config schema fix.

## Summary by CodeRabbit

* **Refactor**
  * Restructured hook management system for improved configuration handling.
  * Improved agent configuration storage format.

* **Bug Fixes**
  * Enhanced uninstall cleanup to selectively remove only relevant managed hooks while preserving user-defined configurations.
  * Fixed reinstall behavior to reliably restore configurations.

* **Tests**
  * Updated test suite to validate new configuration structures and hook management behavior.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -2066,10 +2066,10 @@ function generateCodexConfigBlock(agents, targetDir) {
   ];
 
   for (const { name, description } of agents) {
-    // #2645 — Codex schema requires [[agents]] array-of-tables, not [agents.<name>] maps.
-    // Emitting [agents.<name>] produces `invalid type: map, expected a sequence` on load.
-    lines.push(`[[agents]]`);
-    lines.push(`name = ${JSON.stringify(name)}`);
+    // Live Codex parser validation on 26.422.3464.0 accepts [agents.<name>]
+    // subtables and rejects [[agents]] with:
+    // `invalid type: sequence, expected struct AgentsToml`.
+    lines.push(`[agents.${name}]`);
     lines.push(`description = ${JSON.stringify(description)}`);
     lines.push(`config_file = "${agentsPrefix}/${name}.toml"`);
     lines.push('');
@@ -2082,8 +2082,8 @@ function generateCodexConfigBlock(agents, targetDir) {
  * Strip any managed GSD agent sections from a TOML string.
  *
  * Handles BOTH shapes so reinstall self-heals broken legacy configs:
- *   - Legacy: `[agents.gsd-*]` single-keyed map tables (pre-#2645).
- *   - Current: `[[agents]]` array-of-tables whose `name = "gsd-*"`.
+ *   - Current: `[agents.gsd-*]` subtables.
+ *   - Broken v1.38.5: `[[agents]]` array-of-tables whose `name = "gsd-*"`.
  *
  * A section runs from its header to the next `[` header or EOF.
  */
@@ -2091,12 +2091,12 @@ function stripCodexGsdAgentSections(content) {
   // Use the TOML-aware section parser so we never absorb adjacent user-authored
   // tables — even if their headers are indented or otherwise oddly placed.
   const sections = getTomlTableSections(content).filter((section) => {
-    // Legacy `[agents.gsd-<name>]` map tables (pre-#2645).
+    // Current `[agents.gsd-<name>]` subtables.
     if (!section.array && /^agents\.gsd-/.test(section.path)) {
       return true;
     }
 
-    // Current `[[agents]]` array-of-tables — only strip blocks whose
+    // Broken v1.38.5 `[[agents]]` array-of-tables — only strip blocks whose
     // `name = "gsd-..."`, preserving user-authored [[agents]] entries.
     if (section.array && section.path === 'agents') {
       const body = content.slice(section.headerEnd, section.end);
@@ -2118,6 +2118,7 @@ function stripCodexGsdAgentSections(content) {
  * Returns cleaned content, or null if file would be empty.
  */
 function stripGsdFromCodexConfig(content) {
+  content = stripLegacyCodexHookBlockFromConfig(content);
   const eol = detectLineEnding(content);
   const markerIndex = content.indexOf(GSD_CODEX_MARKER);
   const codexHooksOwnership = getManagedCodexHooksOwnership(content);
@@ -2809,7 +2810,7 @@ function isLegacyGsdAgentsSection(body) {
 function stripLeakedGsdCodexSections(content) {
   const leakedSections = getTomlTableSections(content)
     .filter((section) => {
-      // Legacy [agents.gsd-<name>] map tables (pre-#2645).
+      // Current [agents.gsd-<name>] subtables.
       if (!section.array && section.path.startsWith('agents.gsd-')) return true;
 
       // Legacy bare [agents] table with only the old max_threads/max_depth keys.
@@ -2819,7 +2820,7 @@ function stripLeakedGsdCodexSections(content) {
         isLegacyGsdAgentsSection(content.slice(section.headerEnd, section.end))
       ) return true;
 
-      // Current [[agents]] array-of-tables whose name is gsd-*. Preserve
+      // Broken v1.38.5 [[agents]] array-of-tables whose name is gsd-*. Preserve
       // user-authored [[agents]] entries (other names) untouched.
       if (section.array && section.path === 'agents') {
         const body = content.slice(section.headerEnd, section.end);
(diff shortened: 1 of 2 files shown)


trek-e (commented):
## Adversarial Review

First-time contributor. One PR in the repo. Be sharp.

---

### Issue 1 — Third schema flip in this file with no canonical spec citation

Timeline so far:
- Original: `[agents.gsd-*]` subtables
- PR #2645: switched to `[[agents]]`, citing `invalid type: map, expected a sequence`
- This PR: switches back to `[agents.gsd-*]`, citing `invalid type: sequence, expected struct AgentsToml`

Line 2069–2072 contains the sole justification: `// Live Codex parser validation on 26.422.3464.0`. That is a version string. It is not a spec. It is not a schema file. It is not a link to Codex source or documentation.

PR #2645 had the same problem — its only citation was the error message it claimed to fix. Now that error message is gone but a different one appeared and the fix is to revert.

There is no authoritative Codex TOML schema linked anywhere in this codebase. Without one, this is guess-and-check that will repeat every time Codex ships a parser change. **What is the canonical Codex schema source? Where is it documented that `[agents.<name>]` is the correct shape for the agents registry?** If this cannot be linked, this PR is just shifting which wrong answer is in production.

---

### Issue 2 — Direct unresolved conflict with PR #2640

PR #2640 (hanzckernel, still open) solves the identical problem: move Codex hooks from inline TOML to `hooks.json`. It adds the same concepts — legacy TOML cleanup, `hooks.json` write logic, uninstall cleanup, test coverage. These two PRs are independent implementations of the same migration and they will merge-conflict.

Neither author appears aware of the other. **Did you check open PRs before writing this? The `fix: Codex hook installation to use hooks.json` PR has been open and is already in review. Your implementation and theirs diverge in significant ways (version detection, manifest tracking, exact regex patterns). One of these needs to be closed or they need to be coordinated before either merges.**

---

### Issue 3 — `!fs.existsSync(hooksPath)` at line 7116 is redundant, not a guard

```js
if (JSON.stringify(settings.hooks.SessionStart) !== before || !fs.existsSync(hooksPath)) {
    writeSettings(hooksPath, settings);
}
```

`readSettings()` (line 589–606) returns `{}` for non-existent files, not `null`. The `null` guard at line 7086–7088 only fires for *malformed* files. On a fresh install with no `hooks.json`: `readSettings` returns `{}`, we skip the early return, build `SessionStart` from scratch, and `before` is `"[]"` while the array now has one entry — so `before !== after` is already true. The `!fs.existsSync` branch is never the deciding condition on a fresh install; the content comparison covers it.

The comment intent seems to be "always write on fresh install" but the logic achieves that accidentally via the content comparison, not through the existence check. This is a logic smell. If `before` were ever equal to the new value on a fresh install (it can't be, but the author clearly didn't think through this), the file still wouldn't be written because the existence check comes second. **Rewrite this as two explicit conditions with comments that explain each one.**

---

### Issue 4 — `cleanupOrphanedHooks` and `validateHookFields` are not Codex-aware

Lines 7091–7092:
```js
settings = cleanupOrphanedHooks(settings);
settings = validateHookFields(settings);
```

`cleanupOrphanedHooks` (line 4612) scans for patterns like `gsd-notify.sh`, `hooks/statusline.js` — these are Claude Code hook paths. They are not Codex hook paths. Running this against a `hooks.json` from a Codex install will silently scan and potentially mutate a file that uses a completely different hook schema. `validateHookFields` (line 4683) similarly validates against Claude Code's Zod schema rules (`type: "agent"` requires `prompt`, etc.) — this schema is Claude Code specific.

**Neither function has been validated against Codex's `hooks.json` schema. If Codex's hook entries use a different shape (e.g., no `hooks` wrapper array, flat `command` at top level, or different event key names), `validateHookFields` will strip them as "invalid" and destroy user-configured Codex hooks.** There are no tests for this function called on a Codex-schema `hooks.json`.

---

### Issue 5 — TOML key injection has no name escaping (line 2072)

```js
lines.push(`[agents.${name}]`);
```

The previous `[[agents]]` shape used:
```js
lines.push(`name = ${JSON.stringify(name)}`);
```

TOML bare keys allow only `A-Za-z0-9_-`. The name is interpolated directly into a TOML table header with no validation and no quoting. A name like `gsd-executor.v2` produces `[agents.gsd-executor.v2]` — a nested subtable `agents.gsd-executor` with sub-key `v2`, not an agent named `gsd-executor.v2`. A name with `]` produces invalid TOML entirely.

GSD agent names today happen to be safe identifiers. That is not the same as the code being correct. The previous code was safer because `JSON.stringify` always produces valid TOML string values. **Either validate that agent names are bare-key safe before interpolation, or quote the key using TOML's `["quoted.key"]` syntax.**

---

### Issue 6 — `stripLeakedGsdCodexSections` now strips what `generateCodexConfigBlock` writes (line 2813)

```js
// Current [agents.gsd-<name>] subtables.
if (!section.array && section.path.startsWith('agents.gsd-')) return true;
```

This is the "leaked sections" cleanup function — it removes `[agents.gsd-*]` entries that appear *outside* the GSD marker block. With this PR, `[agents.gsd-*]` is now BOTH the managed format (written inside the marker) and the leaked format being stripped here. 

The question is whether this function is properly scoped to only sections appearing before the GSD marker. Looking at how `stripLeakedGsdCodexSections` is called: it runs on content outside the marker. But the comment was just changed from "Legacy" to "Current" — implying the author knows this is the current format. **Add a test that verifies `stripLeakedGsdCodexSections` does NOT strip a valid `[agents.gsd-executor]` entry that appears inside the GSD-managed block. Without this test, a future regression in the scoping logic would cause reinstall to write agent entries and then immediately strip them.**

---

### Issue 7 — No end-to-end clean-install/uninstall integrity test

There is no test that: (1) runs install on a clean directory, (2) runs uninstall, (3) asserts zero GSD artifacts remain in both `config.toml` and `hooks.json`. The new `runCodexUninstall` helper exists but the tests only verify individual behaviors.

This PR adds the fourth distinct cleanup path for Codex state: `[[hooks]]` TOML strips, `[[agents]]` strips, `[agents.gsd-*]` strips, and now `hooks.json` cleanup. Each path was added to fix what a previous path missed. A clean-install/uninstall roundtrip test would catch regressions in all four paths simultaneously. **Without it, you have no guarantee that the accumulation of cleanup logic actually leaves a clean state.**

---

### Issue 8 — `filterGsdHooks` with custom predicate silently inverts the default on Codex uninstall path

`filterGsdHooks` signature (line 7065):
```js
function filterGsdHooks(entries, predicate = isGsdHookCommand)
```

The predicate is `true` means "this is a GSD hook, filter it out." At line 7102–7105, `configureCodexHooksJson` passes a *narrower* predicate that only matches `gsd-check-update`/`gsd-update-check`. This is intentional — install only manages the update hook.

But the uninstall path (line 4863–4905 area) calls `filterGsdHooks(entries)` with no predicate, using the full `isGsdHookCommand` default. `isGsdHookCommand` includes `gsd-statusline`, `gsd-session-state`, `gsd-validate-commit`, and others — hooks that are written to Claude Code's `settings.json`, not Codex's `hooks.json`. If a user somehow had a Codex `hooks.json` with a `gsd-statusline` entry (unlikely but possible via manual config or a future installer change), uninstall would strip it. **The Codex uninstall path should use the same narrow predicate as install — only strip `gsd-check-update`/`gsd-update-check`. Using the broad `isGsdHookCommand` default here is applying Claude Code's hook vocabulary to a Codex-specific file.**

---

**VERDICT: CHANGES REQUIRED**

trek-e:
closing.  too many issues.

PR #2733, opened 2026-04-26
fix(config): bump MODEL_ALIAS_MAP to claude-opus-4-7

### LINKED ISSUE #2712: bug: MODEL_ALIAS_MAP pinned to claude-opus-4-6, latest is claude-opus-4-7
**Where**
`bin/lib/core.cjs:1432-1436`

**Current**
```js
const MODEL_ALIAS_MAP = {
  'opus': 'claude-opus-4-6',
  'sonnet': 'claude-sonnet-4-6',
  'haiku': 'claude-haiku-4-5',
};
```

**Expected**
`'opus': 'claude-opus-4-7'` (Opus 4.7 shipped Q1 2026, current default in Claude Code).

**Impact**
When `resolve_model_ids: true` is set in `.planning/config.json`, agents receive the stale model ID instead of the latest. No runtime error (Claude still resolves alias internally), but full IDs in commits/logs/exports are misleading and may diverge from the runtime's actual model.

**Repro**
1. `.planning/config.json` -> `"resolve_model_ids": true`
2. `gsd-tools resolve-model gsd-planner --raw` -> returns `claude-opus-4-6` (should be `claude-opus-4-7`)

**Patch (1-line)**
Bump `'opus': 'claude-opus-4-7'` in `MODEL_ALIAS_MAP`.

**Workaround**
Use `model_overrides` with bare alias `"opus"` instead of `resolve_model_ids: true`.

Version: 1.38.3

trek-e:
## Summary

- Bumps `MODEL_ALIAS_MAP['opus']` from `claude-opus-4-6` to `claude-opus-4-7` in `get-shit-done/bin/lib/core.cjs`
- Bumps `RUNTIME_PROFILE_MAP.claude.opus.model` to match
- Adds `resolve_model_ids: true` test suite — this code path had zero coverage, which allowed the stale ID to survive undetected

## Root cause

`MODEL_ALIAS_MAP` is a maintenance constant that must be updated whenever Anthropic ships a new major model. Opus 4.7 shipped Q1 2026 but the bump was never applied. The alias resolved correctly in practice (Claude's API accepts aliases internally) but the full model IDs emitted to logs, exports, and agent-tool calls were stale.

## Adversarial review

- **Goodhart's Law**: The existing test suite validated that aliases resolved to *some* string, not to the *correct* string. Tests measuring "returns a non-empty alias" can't catch this class of stale-value bug.
- **Knuth**: No optimization applies here; the fix is a one-value constant update. The added tests are O(1) and add no complexity.
- **Peter Principle**: The regex-based alias expansion was working at its level of competence — the only failure was a missing update process to bump the constant on model releases.

## Test plan
- [x] `node --test --test-name-pattern "resolve_model_ids: true" tests/core.test.cjs` — all 3 pass
- [x] `node --test tests/core.test.cjs` — full suite green (0 failures)

Closes #2712

## Summary by CodeRabbit

* **Updates**
  * Updated Claude Opus model identifier to the latest release version. The Opus tier now resolves to the current Claude Opus release, ensuring consistent access to the newest model across all model resolution methods.

### DIFF
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -1506,11 +1506,9 @@ const MODEL_ALIAS_MAP = {
  * provider-specific IDs the runtime cannot accept.
  */
 const RUNTIME_PROFILE_MAP = {
-  claude: {
-    opus:   { model: MODEL_ALIAS_MAP['opus'] },
-    sonnet: { model: MODEL_ALIAS_MAP['sonnet'] },
-    haiku:  { model: MODEL_ALIAS_MAP['haiku'] },
-  },
+  claude: Object.fromEntries(
+    Object.entries(MODEL_ALIAS_MAP).map(([tier, model]) => [tier, { model }])
+  ),
   codex: {
     opus:   { model: 'gpt-5.4',        reasoning_effort: 'xhigh' },
     sonnet: { model: 'gpt-5.3-codex',  reasoning_effort: 'medium' },
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -1488,7 +1488,7 @@ function checkAgentsInstalled() {
  * Users can override with model_overrides in config.json for custom/latest models.
  */
 const MODEL_ALIAS_MAP = {
-  'opus': 'claude-opus-4-6',
+  'opus': 'claude-opus-4-7',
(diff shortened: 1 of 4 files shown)


trek-e (commented):
Fixed in latest commit: RUNTIME_PROFILE_MAP.claude now derives from MODEL_ALIAS_MAP.

PR #2738, opened 2026-04-26
test(query): Vitest regression guard for milestone.complete version arg (fix #2644)

### LINKED ISSUE #2644: milestone.complete query handler drops version arg → always returns completed:false
### GSD Version

`1.38.3` (installed via `npx get-shit-done-cc@latest`)

### Runtime

Claude Code

### Operating System

macOS 26.3.1

### Node.js Version

v24.6.0

### Shell

/bin/zsh

### Installation Method

`npx get-shit-done-cc@latest`

### What happened?

`gsd-sdk query milestone.complete <version>` always returns `{"completed": false, "reason": "GSDError: version required for phases archive"}`, even when the version is provided and the milestone is legitimately complete. The `milestoneComplete` handler extracts its `version` arg correctly but then passes `[]` (empty array) instead of `[version]` when invoking `phasesArchive`, which then throws on the missing arg. The surrounding try/catch swallows the error into a misleading `{completed: false, reason: ...}` response — masking the defect as a "normal" negative answer.

### What did you expect?

`gsd-sdk query milestone.complete "1.1"` should return `{"completed": true, "version": "1.1", "archive": {...}}` when the milestone exists and its phases are complete.

### Steps to reproduce

1. Install GSD: `npx get-shit-done-cc@latest`
2. From any GSD project where at least one milestone is complete:
   ```
   gsd-sdk query milestone.complete "1.1"
   ```
3. Observe the failure response.

Compare with `state.load`, which correctly sees the milestone as complete:

```
$ gsd-sdk query state.load
{
  "milestone": "v1.1",
  "status": "completed",
  "progress": { "total_phases": 7, "completed_phases": 7, "percent": 100 },
  ...
}
```

And the underlying `phases.archive` handler, which works correctly when given the same arg directly:

```
$ gsd-sdk query phases.archive "1.1"
{"archived": 7, "version": "1.1", "archive_directory": ".planning/milestones/1.1-phases"}
```

So the arg parsing, milestone filter, and archive logic all work. The defect is purely in how `milestoneComplete` forwards its argument.

### Error output / logs

```shell
$ gsd-sdk query milestone.complete "1.1"
{"completed": false, "reason": "GSDError: version required for phases archive"}
```

### Impact

### Workaround (if any)

Call `gsd-sdk query phases.archive "<version>"` directly. It performs the same archive operation that `milestone.complete` was supposed to run.

## Root Cause

`sdk/src/query/phase-lifecycle.ts:1439–1453`:

```ts
export const milestoneComplete: QueryHandler = async (args, projectDir) => {
  const version = args[0] || 'current';
  try {
    const archiveResult = await phasesArchive([], projectDir);   // <-- bug is here
    return {
      data: {
        completed: true,
        version,
        archive: archiveResult.data,
      },
    };
  } catch (err) {
    return { data: { completed: false, reason: String(err) } };
  }
};
```

`phasesArchive` at line 1402 reads `args[0]` for the version and throws `GSDError('version required for phases archive', ...)` when it's undefined:

```ts
export const phasesArchive: QueryHandler = async (args, projectDir) => {
  const version = args[0];
  if (!version) {
    throw new GSDError('version required for phases archive', ErrorClassification.Validation);
  }
  ...
};
```

`milestoneComplete` extracts `version` on line 1440, then passes `[]` (empty array) instead of `[version]` on line 1442. The try/catch then swallows the resulting error into a non-descriptive `{completed: false, ...}` response.

## Suggested Fix

**Minimal one-line fix:**

```diff
-    const archiveResult = await phasesArchive([], projectDir);
+    const archiveResult = await phasesArchive([version], projectDir);
```

There's also a subtle issue with `version = args[0] || 'current'`: if the caller omits the version, this passes the literal string `"current"` through to `phasesArchive`, which will create `.planning/milestones/current-phases/` rather than resolving to the actual current milestone. Consider either:

- Requiring `version` (throw a clear error if `args[0]` is missing), or
- Resolving `"current"` to the real milestone version by reading `STATE.md` before calling `phasesArchive`.

## Test Coverage Gap

`sdk/src/query/phase-lifecycle.test.ts` has **no direct test for `milestoneComplete`**. A single unit test asserting that calling `milestoneComplete(['1.1'], projectDir)` on a fixture with v1.1 phases returns `{completed: true, ...}` would have caught this.

Suggested test:

```ts
it('milestoneComplete forwards version to phasesArchive', async () => {
  // set up .planning/phases/ with a dir belonging to milestone 1.1
  // set up STATE.md / ROADMAP.md so getMilestonePhaseFilter returns true for it
  const result = await milestoneComplete(['1.1'], tmpDir);
  expect(result.data.completed).toBe(true);
  expect(result.data.version).toBe('1.1');
  expect(result.data.archive.archived).toBeGreaterThan(0);
});
```

## Design Note (Optional)

`milestone.complete` being a **query** that performs destructive side-effects (moves directories) is surprising. Most SDK consumers would reasonably assume `query` handlers are read-only. Consider either:

- Renaming the handler to `milestone.finalize` or `milestone.archive` to signal the side-effect, and keeping `milestone.complete` as a pure read-only completion check, **or**
- Documenting loudly in the query registry that this specific query mutates the filesystem.

This is orthogonal to the bug fix — calling it out so the naming choice is deliberate when the fix lands.

## Impact

Any GSD workflow that gates on `milestone.complete` — e.g. a `/gsd-*` command that checks "is this milestone done before starting the next?" — receives an incorrect negative answer. Users who follow the error message chase data-integrity ghosts before realizing the bug is in the handler itself.

### Privacy Checklist

trek-e:
## Summary

- Adds five Vitest unit tests for `milestoneComplete` in `decomposed-handlers.test.ts` — the handler was imported but had zero test coverage
- Tests lock in the correct contract that was broken in the original implementation and fixed in c5b1445
- Regression guard ensures the `{ completed: false, reason: "GSDError: version required for phases archive" }` error shape cannot resurface silently

## Root cause (for record)

The original defect (introduced in `6f79b1d`, `feat(sdk): Phase 1 typed query foundation`):

```js
export const milestoneComplete = async (args, projectDir) => {
    const version = args[0] || 'current';
    try {
        const archiveResult = await phasesArchive([], projectDir);  // empty args — bug
        return { data: { completed: true, version, archive: archiveResult.data } };
    } catch (err) {
        return { data: { completed: false, reason: String(err) } };  // swallowed
    }
};
```

`phasesArchive([], ...)` read `args[0]` → `undefined` → threw `GSDError('version required for phases archive')`. The catch swallowed it into `{ completed: false, reason: "GSDError: version required for phases archive" }`, making a programming error look like a legitimate "milestone not complete" answer.

Fixed in `c5b1445` (golden parity harness): handler now validates version upfront with an explicit throw and uses inline archive logic, never delegating to `phasesArchive`.

## What the fix does

The source code was already corrected in `c5b1445`. This PR adds the missing test coverage that would have:
1. Caught the original bug before merge
2. Prevented regression if the inline logic is ever refactored to call `phasesArchive` again

## Test plan

- [ ] `cd sdk && npm run test:unit` — all milestoneComplete tests green
- [ ] `node --test tests/bug-2684-milestone-complete-version.test.cjs` — existing node:test regression suite still passes
- [ ] `gsd-sdk query milestone.complete "v1.0"` against a project with `.planning/` — returns `{ version: "v1.0", date: "...", phases: N, milestones_updated: true }`, not `{ completed: false, reason: "..." }`

Closes #2644

## Summary by CodeRabbit

* **Tests**
  * Added regression test suite for milestone completion functionality, including validation of version argument handling, successful result structure verification, and option propagation testing.

### DIFF
--- a/sdk/package-lock.json
+++ b/sdk/package-lock.json
@@ -22,7 +22,7 @@
         "vitest": "^3.1.1"
       },
       "engines": {
-        "node": ">=20"
+        "node": ">=22.0.0"
       }
     },
     "node_modules/@anthropic-ai/claude-agent-sdk": {
--- a/sdk/src/query/decomposed-handlers.test.ts
+++ b/sdk/src/query/decomposed-handlers.test.ts
@@ -170,6 +170,74 @@ describe('progressBar', () => {
   });
 });
 
+// ─── phase-lifecycle.ts — milestoneComplete ──────────────────────────────
+
+/**
+ * Regression tests for bug #2644: milestone.complete handler drops version arg.
+ *
+ * Original defect (first introduced in 6f79b1d): the handler called
+ * `phasesArchive([], projectDir)` instead of forwarding the version positional
+ * arg. phasesArchive read args[0] and threw GSDError('version required for
+ * phases archive'); the surrounding try/catch swallowed the throw into
+ * { completed: false, reason: String(err) }, masking it as a legitimate
+ * negative answer.
+ *
+ * Fixed in c5b1445: handler now validates version upfront and uses inline
+ * archive logic instead of delegating to phasesArchive.
+ */
+describe('milestoneComplete', () => {
+  it('accepts version as first positional arg and returns it in data', async () => {
+    const result = await milestoneComplete(['v1.19', '--name', 'Test Milestone'], tmpDir);
+    const data = result.data as Record<string, unknown>;
+
+    // Must NOT return the error shape from the old bug
+    expect(data.completed).not.toBe(false);
+    expect((data as Record<string, unknown>).reason).toBeUndefined();
+
+    // Must return version echoed in data
+    expect(data.version).toBe('v1.19');
+  });
+
+  it('does not call phasesArchive with empty args (regression: bug #2644)', async () => {
+    // If the old bug were present, this would return { completed: false, reason: 'GSDError: version required for phases archive' }
+    // The fix ensures version is extracted from args[0] before any archive operation
+    const result = await milestoneComplete(['v1.0'], tmpDir);
+    const data = result.data as Record<string, unknown>;
+
+    // Should return the success shape with version, never the error shape
+    expect(data.version).toBe('v1.0');
+    expect(typeof data.date).toBe('string');
+    expect(typeof data.phases).toBe('number');
+    expect(data.milestones_updated).toBe(true);
+  });
+
+  it('throws GSDError when version arg is missing (not masked as completed: false)', async () => {
+    // The old bug swallowed ALL errors into { completed: false, reason: String(err) }
+    // The fix explicitly throws so callers can distinguish validation failure from "not complete"
+    await expect(milestoneComplete([], tmpDir)).rejects.toThrow('version required for milestone complete');
+  });
+
+  it('archives with --archive-phases when flag is present', async () => {
+    const result = await milestoneComplete(['v1.0', '--archive-phases'], tmpDir);
+    const data = result.data as Record<string, unknown>;
+
+    expect(data.version).toBe('v1.0');
+    const archived = data.archived as Record<string, unknown>;
+    // --archive-phases was passed; phases dir should have been scoped but
+    // may result in 0 if the milestone filter finds no matching dirs.
+    // The important assertion: no error, version is correctly forwarded.
+    expect(typeof archived.phases).toBe('boolean');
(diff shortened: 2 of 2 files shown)


trek-e (commented):
Done: extracted assertMilestoneSuccess helper in latest commit.

Issue #2734, opened 2026-04-26
parseMustHavesBlock drops quoted truths containing ':' (silent {} → TypeError in roadmap.annotate-dependencies)


brian-anderson:
**Version:** `get-shit-done-cc@1.38.4`
**Trigger:** `gsd-sdk query roadmap.annotate-dependencies <phase>` after planning a phase whose `must_haves.truths` items contain a `:` character (very common — references like `D-15:`, file paths like `backend/src/models/mod.rs:`, or sentence-internal colons).

## Repro

A PLAN.md frontmatter snippet that exhibits the bug:

```yaml
---
must_haves:
  truths:
    - "App-side UUIDv4 ID generation: `uuid` crate dep is added with features v4+serde; no migration uses `DEFAULT gen_random_uuid()` (D-01)"
    - "`backend/src/models/mod.rs` declares `pub mod` for the 11 model files Wave 2-4 will populate"
---
```

Both are valid YAML quoted scalars. But:

```js
const { parseMustHavesBlock } = require('./bin/lib/frontmatter.cjs');
const truths = parseMustHavesBlock(content, 'truths');
// First item: {} (an empty object — should be the string)
// Second item: '`backend/src/models/mod.rs` declares...' (correct)
```

When `roadmap.annotate-dependencies` then runs (`bin/lib/roadmap.cjs:420`):

```js
for (const t of truths) {
  const key = t.trim().toLowerCase();  // TypeError: t.trim is not a function
}
```

…because `{}` has no `.trim()`.

## Root cause

`bin/lib/frontmatter.cjs` lines ~239–252:

```js
const afterDash = trimmed.slice(2);
// Check if it's a simple string item (no colon means not a key-value)
if (!afterDash.includes(':')) {
  current = afterDash.replace(/^["']|["']$/g, '');
} else {
  // Key-value on same line as dash: "- path: value"
  const kvMatch = afterDash.match(/^(\w+):\s*"?([^"]*)"?\s*$/);
  if (kvMatch) {
    current = {};
    current[kvMatch[1]] = kvMatch[2];
  }
}
```

The colon test `!afterDash.includes(':')` doesn't account for **quoted strings that legitimately contain colons**. When a quoted truth contains `:`, control flows to the key-value branch. The `\w+:` pattern then fails to match (because `\w+` doesn't include `\"` or `` ` ``), so `kvMatch` is `null` and `current` keeps its previous initialization of `{}` (set unconditionally at the top of the list-item branch).

[...]

PR #2759, opened 2026-04-27
fix(#2757): unquoted truths with colons crash annotate-dependencies

### LINKED ISSUE #2757: TypeError: t.trim is not a function from roadmap.cjs:420
### GSD Version

1.38.5

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

Node: v22.22.2

### Shell

_No response_

### Installation Method

### What happened?

  - Symptom — TypeError: t.trim is not a function from roadmap.cjs:420                         
  - Repro — minimal PLAN.md with a single colon-containing quoted truth
  - Root cause — parseMustHavesBlock in frontmatter.cjs dispatches on includes(':') instead of 
  checking for leading quote, so a plain-string item like "Run \rails db:seed`..."falls through
   and returns{}instead of the string.roadmap.cjs:420then callst.trim()` on the empty object.  
  - Three suggested fixes — A (parser: check for leading quote first), B (consumer: skip       
  non-strings defensively), C (structural: switch to js-yaml).                                 
  - Where it hurts — annotate step is silently skipped; ROADMAP loses wave-blocking headers +
  cross-cutting-truths section.                                                                
  - Test cases to add.                                                            
                                                                                               
  Idiomatic Rails truths trip this routinely — db:seed, Service.call(arg:), Class::Method,     
  /foo/:id all contain colons that will fall through the parser.

### What did you expect?

annotation step to not be silently skipped

### Steps to reproduce

`gsd-sdk query roadmap.annotate-dependencies <phase>` crashes when any plan's
`must_haves.truths:` block contains a quoted string list item that includes a
`:` character. This happens routinely — common Rails/Ruby idioms (`db:seed`,
`Service.call(arg:, key:)`, `route /path/:id`, `Class::Method`) all trip it.

A minimal PLAN.md that triggers the bug:

```yaml
---
plan_id: "01"
wave: 1
depends_on: []
files_modified:
  - "config/routes.rb"
autonomous: true
requirements:
  - REQ-01
must_haves:
  truths:
    - "Route `/foo/:id` resolves to controller#show"
  artifacts: []
  key_links: []
---

# Repro plan
```

Run:

```bash
gsd-sdk query roadmap.annotate-dependencies "1"
```

Result:

```
{
  "updated": false,
  "reason": "/home/.../node_modules/get-shit-done-cc/get-shit-done/bin/lib/roadmap.cjs:420
      const key = t.trim().toLowerCase();
                    ^

TypeError: t.trim is not a function
    at Object.cmdRoadmapAnnotateDependencies (.../roadmap.cjs:420:21)
    at runCommand (.../gsd-tools.cjs:705:17)
    ...
```

(The exception is caught and re-emitted as a JSON `reason`, so the orchestrator
sees a "soft failure" rather than a hard crash — but the work is silently
skipped.)

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

### Workaround (if any)

non-blocking (the rest of plan-phase succeeds), but the wave-headers + cross-cutting-truths annotation step is silently skipped, so ROADMAP.md is left without the post-planning enrichment.

## Root cause

The annotate handler calls `parseMustHavesBlock(content, 'truths')` (in
`bin/lib/frontmatter.cjs`, around line 193). That parser handles two list-item
shapes:

1. **Plain string item** — `- "some text"` → returns the unquoted string.
2. **Inline key-value item** — `- path: value` → returns an object
   `{ path: "value" }`.

Discrimination is purely on whether `afterDash.includes(':')`:

```js
// frontmatter.cjs ~ line 248
const afterDash = trimmed.slice(2);
if (!afterDash.includes(':')) {
  current = afterDash.replace(/^["']|["']$/g, '');         // string
} else {
  const kvMatch = afterDash.match(/^(\w+):\s*"?([^"]*)"?\s*$/);
  if (kvMatch) {
    current = {};
    current[kvMatch[1]] = kvMatch[2];
  }
  // else: current stays as the {} initialized just above ← silent fall-through
}
```

When a plain-string truth happens to contain a colon (anywhere — inside
backticks, inside parens, etc.), the parser takes the `else` branch.
`kvMatch` fails because the string isn't actually a YAML key-value, so
`current` is left as the empty object `{}` initialized two lines earlier.
That empty object ends up in the `truths` array.

`roadmap.cjs` then iterates assuming every element is a string:

```js
// bin/lib/roadmap.cjs:418-422
for (const t of truths) {
  const key = t.trim().toLowerCase();   // ← TypeError when t === {}
  if (!key || seen.has(key)) continue;
  ...
}
```

`{}.trim` is undefined → `TypeError: t.trim is not a function`.

## Why valid YAML triggers this

YAML allows a quoted string to contain colons freely:

```yaml
- "Running `rails db:seed` creates ..."   # valid YAML, plain string
- "GET `/foo/:id` resolves to ..."         # valid YAML, plain string
- "key: value"                             # valid YAML, plain string (still!)
```

A real YAML parser wouldn't be fooled — the leading `"` makes it unambiguous.
But the regex parser in `parseMustHavesBlock` only checks `includes(':')`,
which is too coarse.

## Suggested fixes

Two independent fixes — either alone resolves the symptom; together they
harden the contract:

### Fix A (frontmatter.cjs) — detect quoted-string items before the colon test

Rather than dispatching on `includes(':')`, check whether the item starts
with a quote. If yes, treat it as a plain string regardless of inner
colons:

```js
const afterDash = trimmed.slice(2);
const isQuoted = afterDash.startsWith('"') || afterDash.startsWith("'");
if (isQuoted || !afterDash.includes(':')) {
  current = afterDash.replace(/^["']|["']$/g, '');
} else {
  const kvMatch = afterDash.match(/^(\w+):\s*"?([^"]*)"?\s*$/);
  if (kvMatch) {
    current = {};
    current[kvMatch[1]] = kvMatch[2];
  } else {
    // Looked like KV but didn't match the regex — fall back to plain string.
    current = afterDash;
  }
}
```

The explicit `else` ensures `current` is never left as `{}`.

### Fix B (roadmap.cjs) — defend against non-string elements

In the cross-cutting-truths loop, coerce or skip non-strings:

```js
for (const t of truths) {
  if (typeof t !== 'string') continue;   // skip parser quirks defensively
  const key = t.trim().toLowerCase();
  ...
}
```

This is a one-line defensive guard. It costs nothing and protects against
any future parser bug returning unexpected shapes.

### Fix C (optional, structural) — replace the regex parser with `js-yaml`

`parseMustHavesBlock` reimplements YAML parsing from scratch with regexes.
This works for the common cases it's seen but breaks on any of:

- quoted strings containing `:`
- multi-line folded scalars
- YAML anchors/aliases
- block-style mappings

Switching to `js-yaml` (already a transitive dep in many Node projects)
parsing the slice of frontmatter under `must_haves` would eliminate this
class of bug. Larger change; A+B are the minimal fix.

## Where this hurts

`/gsd-plan-phase` step 13c calls `roadmap.annotate-dependencies`. The
crash is caught and reported as `{ updated: false, reason: "..." }`, but:

1. The wave-headers ("**Wave 2** *(blocked on Wave 1 completion)*") and
   cross-cutting truths section never land in ROADMAP.md.
2. The `reason` field returns a multi-line stack trace with absolute paths,
   so the user has to read carefully to figure out it's a tool bug, not a
   plan-quality issue.
3. The downstream `/gsd-execute-phase` workflow keeps moving but reads a
   ROADMAP that omits the parallel-execution hints the annotate step would
   have provided.

For us, none of the 7 plans in our phase were truly broken — schema
validation, plan-checker, and requirements coverage all passed. The
only artifact lost was the ROADMAP enrichment.

## Repro environment

- get-shit-done-cc: 1.38.5
- Node: v22.22.2
- Runtime: Linux 6.8.0 (Ubuntu)
- Invocation: `gsd-sdk query roadmap.annotate-dependencies "97"` from a
  Claude Code `/gsd-plan-phase` workflow

## Files involved

- `bin/lib/frontmatter.cjs:193-` — `parseMustHavesBlock`, the regex parser
- `bin/lib/roadmap.cjs:418-422` — the consumer that assumes string elements
- `bin/lib/roadmap.cjs:395-400` — calls `parseMustHavesBlock(content, 'truths')`

## Test cases worth adding

```js
// frontmatter.cjs
test('parseMustHavesBlock returns strings for quoted items containing colons', () => {
  const yaml = `---
must_haves:
  truths:
    - "Run \`rails db:seed\` creates a row"
    - "GET /foo/:id resolves correctly"
---`;
  const result = parseMustHavesBlock(yaml, 'truths');
  expect(result).toEqual([
    'Run `rails db:seed` creates a row',
    'GET /foo/:id resolves correctly',
  ]);
  // NOT { ... } — never an object for a quoted string
});

// roadmap.cjs
test('annotate-dependencies tolerates a non-string in a truths array', () => {
  // Even if parser returns [{}, "real truth"], the annotate step should not crash
  // and should treat the object as absent.
  expect(() => annotate({ truths: [{}, 'a real truth'] })).not.toThrow();
});
```

### Privacy Checklist

trek-e:
## Summary

- `parseMustHavesBlock` used `includes(':')` to detect key-value pairs, but unquoted YAML strings like `GET /foo/:id resolves...` and `Class::Method is idempotent` also contain colons
- When the KV regex failed to match, `current` was left as `{}` (the empty object initialized before the branch), which then caused `t.trim()` in `roadmap.cjs:420` to throw `TypeError: t.trim is not a function`
- The exception is caught and re-emitted as `{ updated: false, reason: "..." }`, so the orchestrator sees a soft failure — but wave headers and cross-cutting truths are silently dropped from ROADMAP.md

**Root cause — two layers:**

1. **`frontmatter.cjs`**: KV regex used `\s*` (optional space after colon), so `Class::Method is idempotent` matched `(\w+):\s*` → `Class:` with empty space. Tightened to `\s+` (required space), which matches YAML convention (`key: value` always has a space). Added explicit fallback: when the KV regex fails, treat the item as a plain string instead of leaving `current` as `{}`.

2. **`roadmap.cjs`**: No defensive guard before calling `.trim()` on each truth element. Added `typeof t !== 'string'` check as a cheap safety net against any future parser anomaly.

## Test plan

- [x] `node --test tests/frontmatter.test.cjs` — new `#2757` test passes (covers `GET /foo/:id`, `Service.call(arg:)`, `Class::Method`)
- [x] `node --test tests/enh-2447-roadmap-wave-deps.test.cjs` — new integration test passes (colon-containing truths don't crash `annotate-dependencies`)
- [x] `node --test tests/*.test.cjs` — full suite passes (5644 tests, 0 failures)
- [x] Existing `#2734` quoted-string tests still pass

Closes #2757

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of list items containing colons (e.g., `db:seed`, `Class::Method`, route patterns) to prevent incorrect key-value pair interpretation
  * Fixed roadmap annotation to properly filter and process only string values during constraint generation

* **Tests**
  * Added regression tests to ensure list items with colons parse correctly and prevent future parsing failures

### DIFF
--- a/get-shit-done/bin/lib/frontmatter.cjs
+++ b/get-shit-done/bin/lib/frontmatter.cjs
@@ -242,7 +242,6 @@ function parseMustHavesBlock(content, blockName) {
       // Only treat as a top-level list item if at the expected indent
       if (indent === listItemIndent) {
         if (current) items.push(current);
-        current = {};
         const afterDash = trimmed.slice(2);
         const trimmedAfterDash = afterDash.trim();
         // Check if it's a fully-quoted string (may contain ':' inside the quotes)
@@ -254,10 +253,15 @@ function parseMustHavesBlock(content, blockName) {
           current = afterDash.replace(/^["']|["']$/g, '');
         } else {
           // Key-value on same line as dash: "- path: value"
-          const kvMatch = afterDash.match(/^(\w+):\s*"?([^"]*)"?\s*$/);
+          // YAML KV always has at least one space after the colon: "key: value"
+          // Requiring \s+ rejects "Class::Method" and "db:seed" (no space after colon)
+          const kvMatch = afterDash.match(/^(\w+):\s+"?([^"]*)"?\s*$/);
           if (kvMatch) {
             current = {};
             current[kvMatch[1]] = kvMatch[2];
+          } else {
+            // Looks like KV but doesn't match — treat as plain string (#2757)
+            current = afterDash.replace(/^["']|["']$/g, '');
           }
         }
         continue;
--- a/get-shit-done/bin/lib/roadmap.cjs
+++ b/get-shit-done/bin/lib/roadmap.cjs
@@ -417,6 +417,7 @@ function cmdRoadmapAnnotateDependencies(cwd, phaseNum, raw) {
   for (const { truths } of planData) {
     const seen = new Set();
     for (const t of truths) {
+      if (typeof t !== 'string') continue;
       const key = t.trim().toLowerCase();
       if (!key || seen.has(key)) continue;
       seen.add(key);
--- a/tests/enh-2447-roadmap-wave-deps.test.cjs
+++ b/tests/enh-2447-roadmap-wave-deps.test.cjs
@@ -194,6 +194,25 @@ Plans:
     assert.strictEqual(out.updated, false);
   });
 
+  test('#2757: truths containing colons do not crash annotate-dependencies', () => {
+    // Unquoted truths with colons (Rails idioms: db:seed, /foo/:id, Class::Method)
+    // caused parseMustHavesBlock to return {} instead of a string, then t.trim() threw.
+    const colonTruths = [
+      'GET /foo/:id resolves to controller#show',
+      'Class::Method is idempotent',
+      '"Quoted truth with colon: inside"',
+    ];
+    tmpDir = makePlanProject({
+      '.planning/ROADMAP.md': `# Roadmap\n\n### Phase 1: Foundation\n**Goal:** Set up project\n**Plans:** 1 plan\n\nPlans:\n- [ ] 01-01-PLAN.md — Repro plan\n`,
+      '.planning/phases/01-foundation/01-01-PLAN.md': PLAN_TEMPLATE(1, colonTruths),
+    });
+
+    const result = runGsdTools('roadmap annotate-dependencies 1', tmpDir);
+    assert.ok(result.success, `Command threw on colon-containing truths: ${result.error}`);
+    const out = JSON.parse(result.output);
+    assert.ok(typeof out.updated === 'boolean', 'should return a valid result object');
+  });
+
   test('plan-phase.md documents annotate-dependencies step', () => {
     const planPhase = fs.readFileSync(
       path.join(__dirname, '../get-shit-done/workflows/plan-phase.md'), 'utf-8'
--- a/tests/frontmatter.test.cjs
+++ b/tests/frontmatter.test.cjs
@@ -547,6 +547,27 @@ must_haves:
     assert.strictEqual(result[0], 'Key: value pattern preserved');
   });
 
+  test('#2757: unquoted truth containing ":" is preserved as a string — not left as {}', () => {
+    // Unquoted strings with colons (e.g. Rails idioms) were falling through the KV
+    // regex and leaving current as {}, which caused t.trim() to throw in roadmap.cjs.
+    const content = `---
+phase: 01
+must_haves:
+  truths:
+    - GET /foo/:id resolves to controller#show
+    - Service.call(arg:, key:) returns a record
+    - Class::Method is idempotent
+---
+`;
+    const result = parseMustHavesBlock(content, 'truths');
+    assert.ok(Array.isArray(result), 'should return an array');
+    assert.strictEqual(result.length, 3, `expected 3, got ${result.length}: ${JSON.stringify(result)}`);
+    assert.ok(typeof result[0] === 'string', `result[0] should be string, got ${typeof result[0]}`);
+    assert.ok(typeof result[1] === 'string', `result[1] should be string, got ${typeof result[1]}`);
+    assert.ok(typeof result[2] === 'string', `result[2] should be string, got ${typeof result[2]}`);
+    assert.ok(result[0].includes(':'), 'colon should be preserved in the string');
+  });
+
   test('handles nested arrays within artifact objects', () => {
     const content = `---
 phase: 01


trek-e:
**Rebased onto main** — the merge conflict is resolved.

The branch previously included two pre-squash commits (3ca3f7b8, 63e5e752) that became PR #2744 on main via squash merge. After rebasing, the branch now contains only the #2757 fix commit (2c7d126c).

**CodeRabbit nitpick on `state.cjs`**: No longer applicable — that code was from the dropped pre-squash commits and is already in main via #2744. Our diff is now cleanly scoped to the 4 files in the original fix: `frontmatter.cjs`, `roadmap.cjs`, and the two test files.

Tests: 52 pass, 0 fail.

Issue #2153, opened 2026-04-12
Codex global install writes gsd-check-update hook config but does not install the hook file


5mghost:
## Environment
- GSD Version: 1.34.2
- Runtime: Codex
- Operating System: macOS 26.4
- Node.js Version: v22.14.0
- Shell: /bin/zsh
- Installation Method: `npx get-shit-done-cc --codex --global`
- Frequency: Every time (100% reproducible)
- Impact: Moderate — Feature is broken but I have a workaround

## What happened?
A fresh Codex global install writes a `SessionStart` hook into `~/.codex/config.toml`, but does not copy the referenced hook file into `~/.codex/hooks/`.

Observed after install:
- `~/.codex/config.toml` contains:
  - `command = "node /Users/REDACTED/.codex/hooks/gsd-check-update.js"`
- But `~/.codex/hooks/gsd-check-update.js` does not exist.

That leaves Codex with a broken hook reference immediately after a documented install.

## What did you expect?
`npx get-shit-done-cc --codex --global` should either:
- copy all referenced hook files into `~/.codex/hooks/`, or
- avoid writing hook entries for files that were not installed.

## Steps to reproduce
1. Start from a Codex setup without an existing GSD install.
2. Run:
   ```bash
   npx get-shit-done-cc --codex --global
   ```
3. Inspect the generated Codex config:
   ```bash
   rg -n 'gsd-check-update' ~/.codex/config.toml
   ```
4. Check whether the referenced hook exists:
   ```bash
   ls -la ~/.codex/hooks/gsd-check-update.js
   ```
5. The config contains the hook reference, but the file is missing.

## Error output / logs
Installer output included:

```text
Installing for Codex to ~/.codex
✓ Installed 68 skills to skills/
✓ Installed get-shit-done
✓ Installed agents
✓ Wrote VERSION (1.34.2)
✓ Wrote file manifest (gsd-file-manifest.json)
✓ Generated config.toml with 24 agent roles
✓ Generated 24 agent .toml config files
✓ Configured Codex hooks (SessionStart)
✓ Set resolve_model_ids: "omit" in ~/.gsd/defaults.json
```

But immediately after install:

```bash
$ test -f ~/.codex/hooks/gsd-check-update.js && echo present || echo missing
missing
```

## Workaround
I worked around it by extracting the published npm tarball and manually copying `package/hooks/*` into `~/.codex/hooks/`.

I also verified the npm package itself does include the missing file:
[...]

PR #1346, opened 2026-03-24
fix: preserve top-level config keys and use absolute agent paths for Codex ≥0.116

MiaoDX:
## Summary

Fixes two Codex `config.toml` compatibility issues that break `codex` startup after GSD install:

- **`ensureCodexHooksFeature`**: insert `[features]` block before the first table header instead of prepending it before all content. Prepending traps bare top-level keys (`model`, `model_reasoning_effort`) under `[features]`, where Codex rejects them with `invalid type: string "gpt-5.4", expected a boolean`.
- **`generateCodexConfigBlock`**: use absolute `config_file` paths when `targetDir` is provided. Codex ≥0.116 requires `AbsolutePathBuf` and cannot resolve relative `"agents/..."` paths, failing with `AbsolutePathBuf deserialized without a base path`.

## Reproduction

1. Have a `~/.codex/config.toml` with bare top-level keys:
   ```toml
   model = "gpt-5.4"
   model_reasoning_effort = "high"
   
   [projects."/home/user/myproject"]
   trust_level = "trusted"
   ```
2. Run `npx get-shit-done-cc --codex --global`
3. Try to start `codex` → crashes with:
   ```
   Error: config.toml:4:9: invalid type: string "gpt-5.4", expected a boolean
   Caused by: AbsolutePathBuf deserialized without a base path in `agents`
   ```

## Before → After

**Before** (broken): GSD prepends `[features]`, trapping `model=` under it:
```toml
[features]
codex_hooks = true

model = "gpt-5.4"          ← now under [features], Codex expects boolean
[projects."..."]
```

**After** (fixed): `[features]` inserted after bare keys, before first section:
```toml
model = "gpt-5.4"          ← stays top-level ✓

[features]
codex_hooks = true

[projects."..."]
```

## Test plan

- [x] All 78 existing tests pass (updated 2 assertions that assumed prepend behavior)
- [x] New test: `bare top-level keys are NOT trapped under [features] (#1202)`
- [x] New test: `config_file paths are absolute using CODEX_HOME`
- [x] New test: `uses absolute config_file paths when targetDir is provided`
- [x] Verified `codex --version` succeeds after GSD install with fix applied

Fixes #1202

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2763, opened 2026-04-27
fix(#2637): write Codex SessionStart hook to hooks.json, not config.toml

### LINKED ISSUE #2637: Codex 0.124.0 hook regression can leave GSD-installed Codex setups unable to start

MiaoDX:
## Linked issue

Fixes #2637.

## What it does

Moves the managed Codex `SessionStart` update-check hook from an inline `[[hooks]]` block in `~/.codex/config.toml` to `~/.codex/hooks.json` (Codex's canonical location, per https://developers.openai.com/codex/hooks). Codex 0.124.0+ rejects the inline shape with `invalid type: map, expected a sequence in 'hooks'`, leaving GSD-installed Codex setups unable to start.

## Why hooks.json (not inline config.toml)

`hooks.json` is the documented canonical location and the de-facto ecosystem convention:

| Project | Where it writes Codex hooks |
|---|---|
| [obra/superpowers](https://github.com/obra/superpowers/blob/main/hooks/hooks.json) | `hooks/hooks.json` |
| [Yeachan-Heo/oh-my-codex](https://github.com/Yeachan-Heo/oh-my-codex) | `.codex/hooks.json` (per README: *"native `.codex/hooks.json` coverage"*) |
| Codex docs | [https://developers.openai.com/codex/hooks](https://developers.openai.com/codex/hooks) |
| Inline `[[hooks]]` (added by [#18893](https://github.com/openai/codex/pull/18893)) | additive alternative — useful for repo-local config, but Codex still warns *"prefer a single representation for this layer"* when both exist |

Currently GSD is the only ecosystem tool writing inline `[[hooks]]`, which (a) breaks 0.124.0+ and (b) collides with `hooks.json` entries from other tools.

## Scope (intentionally narrow)

This PR replaces the inline-write path. It does **not** try to handle every legacy variant — prior attempts ([#2640](https://github.com/gsd-build/get-shit-done/pull/2640), [#2730](https://github.com/gsd-build/get-shit-done/pull/2730)) were closed for sprawl across TOML escape decoding, comment-after-header parsing, and aggressive malformed-file rewrites. Cleanup here is limited to one regex over the canonical block GSD itself emitted.

### Install
- keep `features.codex_hooks = true` in `config.toml` (unchanged)
- register the `SessionStart` hook in `hooks.json` via JSON merge (preserves user entries; idempotent on repeat installs)
- on (re)install, strip any legacy `# GSD Hooks\n[[hooks]]…` block from `config.toml` using a single `\r?\n`-aware regex that covers both `gsd-check-update.js` and the older `gsd-update-check.js` filenames in one pass (LF / CRLF / mixed EOLs)

### Uninstall
- remove only the managed GSD entry from `hooks.json`
- delete `hooks.json` when GSD was its only entry; otherwise preserve user hooks

### Out of scope (deliberately not handled)
- TOML escape decoding for legacy hook commands with embedded `\"` (rare; if a user hand-edited the block, they can hand-clean it)
- `[[hooks]] # comment` with trailing comments (canonical GSD writes never produced them)
- Malformed `hooks.json` repair — we read-only skip and leave user content untouched (avoids the rewrite-on-fallback footgun flagged in #2640)

## What was broken before
- GSD writes `[[hooks]]\nevent = "SessionStart"\ncommand = "node …gsd-check-update.js"` in `~/.codex/config.toml`
- Codex 0.124.0+ rejects this:
  ```
  in `hooks`
  ```
- This is reproducible on every fresh `--codex --global` install against current Codex.

### Tests
- `npm test` → 5650 / 5650 pass
- `tests/codex-config.test.cjs` updated (assertions that previously checked `[[hooks]]` in `config.toml` now check `hooks.json`); 6 new e2e tests added:
  - canonical schema in `hooks.json` after install
  - merges into pre-existing user `hooks.json` without overwriting
  - install does not write `[[hooks]]` in `config.toml`
  - malformed user `hooks.json` is left untouched
  - uninstall removes only GSD entry, preserves user hooks
  - uninstall deletes `hooks.json` when GSD was its only entry
- `tests/bug-2698-crlf-install.test.cjs` — assertions updated to look in `hooks.json` for the new hook; legacy-block removal still verified for LF / CRLF / mixed-EOL fixtures

### Live Codex 0.125.0 (Linux)
Two scenarios exercised against a real `codex exec` with isolated `CODEX_HOME`:

1. **Fresh install** — config.toml has `[features].codex_hooks = true`, no `[[hooks]]`; hooks.json has the GSD `SessionStart` entry; `codex exec` starts cleanly (no parse error).
2. **Migration from 1.38.5** — pre-populated config.toml with broken legacy `[[hooks]]` block + pre-existing user `hooks.json` (4 events from tmux-agent-status); after install the legacy block is gone, user hooks preserved, GSD hook appended; `codex exec` starts cleanly.

### Files
- `bin/install.js` — install path replaced (~30 net), two helpers added (`upsertManagedCodexHookInHooksJson`, `removeManagedCodexHookFromHooksJson`), uninstall extended by 5 lines
- `tests/codex-config.test.cjs` — assertion updates + 6 new tests
- `tests/bug-2698-crlf-install.test.cjs` — 2 assertion updates
- `CHANGELOG.md` — Unreleased entry

## Checklist

- [ ] Linked issue has the `confirmed-bug` label *(yes — see `confirmed-bug` on #2637, set by maintainer triage on 2026-04-24)*
- [x] Regression test added (would have caught the original bug + the legacy-block migration)
- [x] All existing tests pass (`npm test` → 5650 / 5650)
- [x] CHANGELOG.md updated
- [x] No new dependencies

## Breaking changes

None. Users who previously ran a GSD install will have their legacy `[[hooks]]` block migrated automatically on the next install. User-authored `hooks.json` entries are preserved.

## Notes for reviewers

`migrateCodexHooksMapFormat` (legacy `[hooks.shell]` → `[[hooks]]`) is no longer called by the install path. Its tests still pass against the standalone function. Removing the function and its tests is a small follow-up; left out here to keep this PR scoped.

This is opened as a draft so prior-PR scope concerns can be discussed before review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hook registration now uses a dedicated structured hooks file instead of inline entries in the main config. Installer migrates legacy inline hooks, preserves user-authored hooks, warns and skips malformed files, and is idempotent on reinstall. Uninstaller removes only the managed hook entry and deletes the hooks file only if it becomes empty.

* **Chores**
  * Tests updated to validate install/uninstall, migration, EOL variants, idempotency, and malformed-file handling.

* **Documentation**
  * Changelog entry added describing install/uninstall and migration behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   it). Re-run `gsd update` without `--minimal` to expand to the full surface. The
   install manifest now records `mode: "minimal" | "full"`. (#2762)
 
+### Fixed
+- **Codex hook configuration moved to `hooks.json` (#2637).** The installer no longer writes an inline `[[hooks]]` block to `~/.codex/config.toml`; Codex 0.124.0+ rejects that shape with `invalid type: map, expected a sequence in 'hooks'`. The managed `SessionStart` hook now lives in `hooks.json` (Codex's canonical location, per https://developers.openai.com/codex/hooks). Re-installing migrates legacy `[[hooks]]` blocks (both `gsd-check-update.js` and the older `gsd-update-check.js` filename, LF/CRLF/mixed EOLs). User-authored hooks in either file are preserved. Uninstall removes only the managed GSD entry.
+
 ## [1.38.5] - 2026-04-25
 
 ### Fixed
--- a/bin/install.js
+++ b/bin/install.js
@@ -2785,6 +2785,111 @@ function setManagedCodexHooksOwnership(content, ownership) {
     remainder;
 }
 
+// Substring marker used to identify the GSD-managed hook in hooks.json.
+// Matching by command-substring stays stable across install locations.
+const GSD_CODEX_HOOK_MARKER = 'gsd-check-update';
+const GSD_CODEX_HOOK_EVENT = 'SessionStart';
+
+function isManagedCodexHookCommand(command) {
+  return typeof command === 'string' && command.includes(GSD_CODEX_HOOK_MARKER);
+}
+
+/**
+ * Insert (or update) the managed GSD SessionStart hook in Codex's hooks.json.
+ * Schema: https://developers.openai.com/codex/hooks. User-authored entries are
+ * preserved; unsupported root/`hooks` shapes return false without rewriting.
+ */
+function upsertManagedCodexHookInHooksJson(hooksJsonPath, command) {
+  let data = { hooks: {} };
+  if (fs.existsSync(hooksJsonPath)) {
+    try {
+      const parsed = JSON.parse(fs.readFileSync(hooksJsonPath, 'utf-8'));
+      if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) return false;
+      data = parsed;
+    } catch (e) {
+      return false;
+    }
+  }
+  if (data.hooks === undefined) {
+    data.hooks = {};
+  } else if (!data.hooks || typeof data.hooks !== 'object' || Array.isArray(data.hooks)) {
+    return false;
+  }
+  if (data.hooks[GSD_CODEX_HOOK_EVENT] === undefined) {
+    data.hooks[GSD_CODEX_HOOK_EVENT] = [];
+  } else if (!Array.isArray(data.hooks[GSD_CODEX_HOOK_EVENT])) {
+    return false;
+  }
+
+  let updated = false;
+  for (const entry of data.hooks[GSD_CODEX_HOOK_EVENT]) {
+    if (!entry || !Array.isArray(entry.hooks)) continue;
+    for (const h of entry.hooks) {
+      if (h && isManagedCodexHookCommand(h.command)) {
+        h.command = command;
+        updated = true;
+      }
+    }
+  }
+  if (!updated) {
+    data.hooks[GSD_CODEX_HOOK_EVENT].push({ hooks: [{ type: 'command', command }] });
+  }
+
+  fs.writeFileSync(hooksJsonPath, JSON.stringify(data, null, 2) + '\n');
+  return true;
+}
+
+/**
+ * Remove the managed GSD hook from Codex's hooks.json. User entries stay.
+ * Deletes the file only when the whole root object is empty after cleanup.
+ */
+function removeManagedCodexHookFromHooksJson(hooksJsonPath) {
+  if (!fs.existsSync(hooksJsonPath)) return false;
+  let data;
+  try {
+    data = JSON.parse(fs.readFileSync(hooksJsonPath, 'utf-8'));
+  } catch (e) {
+    return false;
+  }
+  if (!data || typeof data !== 'object' || Array.isArray(data) ||
+      !data.hooks || typeof data.hooks !== 'object' || Array.isArray(data.hooks)) {
+    // Unsupported root or `hooks` shape — leave the user's file alone.
+    return false;
+  }
+
+  let changed = false;
+  for (const event of Object.keys(data.hooks)) {
+    if (!Array.isArray(data.hooks[event])) continue;
+    const filtered = data.hooks[event]
+      .map((entry) => {
+        if (!entry || !Array.isArray(entry.hooks)) return entry;
+        const before = entry.hooks.length;
+        entry.hooks = entry.hooks.filter((h) => !(h && isManagedCodexHookCommand(h.command)));
+        if (entry.hooks.length !== before) changed = true;
+        return entry.hooks.length > 0 ? entry : null;
+      })
+      .filter(Boolean);
+    if (filtered.length !== data.hooks[event].length) changed = true;
+    if (filtered.length === 0) delete data.hooks[event];
+    else data.hooks[event] = filtered;
+  }
+
+  if (!changed) return false;
+  if (Object.keys(data.hooks).length === 0) {
+    // Only delete the file when the entire root is empty — otherwise the user's
+    // sibling top-level keys (e.g. version markers, custom metadata) survive.
+    delete data.hooks;
+    if (Object.keys(data).length === 0) {
+      fs.unlinkSync(hooksJsonPath);
+    } else {
+      fs.writeFileSync(hooksJsonPath, JSON.stringify(data, null, 2) + '\n');
+    }
+  } else {
+    fs.writeFileSync(hooksJsonPath, JSON.stringify(data, null, 2) + '\n');
+  }
+  return true;
+}
+
 function isLegacyGsdAgentsSection(body) {
   const lineRecords = getTomlLineRecords(body);
   const legacyKeys = new Set(['max_threads', 'max_depth']);
@@ -4971,6 +5076,13 @@ function uninstall(isGlobal, runtime = 'claude') {
           console.log(`  ${green}✓${reset} Cleaned GSD sections from config.toml`);
         }
       }
+
+      // Codex: remove the managed GSD hook from hooks.json (preserves user hooks).
+      const hooksJsonPath = path.join(targetDir, 'hooks.json');
+      if (removeManagedCodexHookFromHooksJson(hooksJsonPath)) {
+        removedCount++;
+        console.log(`  ${green}✓${reset} Removed GSD hook from hooks.json`);
+      }
     }
   } else if (isCopilot) {
     // Copilot: remove skills/gsd-*/ directories (same layout as Codex skills)
@@ -6420,48 +6532,35 @@ function install(isGlobal, runtime = 'claude') {
       console.log(`  ${green}✓${reset} Installed hooks`);
     }
 
-    // Add Codex hooks (SessionStart for update checking) — requires codex_hooks feature flag
+    // SessionStart hook lives in hooks.json (https://developers.openai.com/codex/hooks);
+    // older GSD installs wrote it inline in config.toml — strip that on migration.
     const configPath = path.join(targetDir, 'config.toml');
+    const hooksJsonPath = path.join(targetDir, 'hooks.json');
     try {
       let configContent = fs.existsSync(configPath) ? fs.readFileSync(configPath, 'utf-8') : '';
-      const eol = detectLineEnding(configContent);
-
-      // Migrate legacy [hooks] map format to [[hooks]] array-of-tables (#2637).
-      // Codex 0.124.0 requires [[hooks]] array-of-tables; old GSD installs wrote
-      // [hooks.shell] map tables which now cause a startup parse error.
-      const migratedContent = migrateCodexHooksMapFormat(configContent);
-      if (migratedContent !== configContent) {
-        configContent = migratedContent;
-        console.log(`  ${green}✓${reset} Migrated legacy Codex [hooks] map format to [[hooks]] array-of-tables`);
-      }
 
       const codexHooksFeature = ensureCodexHooksFeature(configContent);
       configContent = setManagedCodexHooksOwnership(codexHooksFeature.content, codexHooksFeature.ownership);
 
-      // Add SessionStart hook for update checking
-      const updateCheckScript = path.resolve(targetDir, 'hooks', 'gsd-check-update.js').replace(/\\/g, '/');
-      const hookBlock =
-        `${eol}# GSD Hooks${eol}` +
-        `[[hooks]]${eol}` +
-        `event = "SessionStart"${eol}` +
-        `command = "node ${updateCheckScript}"${eol}`;
-
-      // Migrate legacy gsd-update-check entries from prior installs (#1755 followup)
-      // Remove stale hook blocks that used the inverted filename or wrong path.
-      // Single \r?\n-aware regex handles LF, CRLF, and block-at-file-start (#2698).
-      if (configContent.includes('gsd-update-check')) {
-        configContent = configContent.replace(
-          /(?:\r?\n|^)# GSD Hooks\r?\n\[\[hooks\]\]\r?\nevent = "SessionStart"\r?\ncommand = "node [^\r\n]*gsd-update-check\.js"\r?\n/gm,
-          (match) => (match.startsWith('\r\n') ? '\r\n' : match.startsWith('\n') ? '\n' : ''),
-        );
-      }
-
-      if (hasEnabledCodexHooksFeature(configContent) && !configContent.includes('gsd-check-update')) {
-        configContent += hookBlock;
-      }
+      // Single \r?\n-aware regex covers LF, CRLF, mixed EOLs, and both filenames
+      // (gsd-check-update.js and the older gsd-update-check.js).
+      configContent = configContent.replace(
+        /(?:\r?\n|^)# GSD Hooks(?:\r?\n)\[\[hooks\]\](?:\r?\n)event = "SessionStart"(?:\r?\n)command = "node [^\r\n]*gsd-(?:check-update|update-check)\.js"(?:\r?\n)/g,
+        (match) => (match.startsWith('\r\n') ? '\r\n' : match.startsWith('\n') ? '\n' : ''),
+      );
 
       fs.writeFileSync(configPath, configContent, 'utf-8');
-      console.log(`  ${green}✓${reset} Configured Codex hooks (SessionStart)`);
+
+      if (hasEnabledCodexHooksFeature(configContent)) {
+        const command = buildHookCommand(targetDir, 'gsd-check-update.js');
+        if (upsertManagedCodexHookInHooksJson(hooksJsonPath, command)) {
+          console.log(`  ${green}✓${reset} Configured Codex hooks (SessionStart)`);
+        } else {
+          console.warn(`  ${yellow}⚠${reset}  Skipped hooks.json update — file is malformed or has an unsupported root shape`);
+        }
+      } else {
+        console.warn(`  ${yellow}⚠${reset}  Skipped Codex hooks — could not enable codex_hooks feature flag in config.toml`);
+      }
     } catch (e) {
       console.warn(`  ${yellow}⚠${reset}  Could not configure Codex hooks: ${e.message}`);
     }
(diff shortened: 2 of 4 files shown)


MiaoDX:
Marking ready for human review. Brief summary of the iteration loop with @coderabbitai for transparency:

**Round 1** (initial review)
- 🟠 Reject array root in `hooks.json` upsert — fixed
- 🟠 Don't delete file when sibling top-level keys exist — fixed
- 🟡 Don't log success when upsert was skipped — fixed

**Round 2**
- 🟠 Don't normalize unsupported user `hooks` payloads (skip instead) — fixed
- 🟠 Quote hook script path so config dirs with spaces work — fixed (switched to existing `buildHookCommand` helper)

**Round 3**
- 🟠 Reject array-shaped `hooks` in uninstall path (symmetry with upsert guard) — fixed
- 🟠 Don't claim success when `codex_hooks` couldn't be enabled — fixed
- 🟡 Defensive `Array.isArray` in test helper — **deferred** (helper only reads files we just wrote; the unsupported shape is unreachable in practice)
- 🟡 Export `GSD_CODEX_HOOK_MARKER` for tests instead of the duplicated literal — **deferred** (small style win, ~3 lines saved, would expand the public surface)

**Final state:**
- 12 / 12 real CI checks pass ✅
- `mergeable_state: clean` ✅
- Live-verified on Codex 0.125.0 against fresh install, migration from 1.38.5, and uninstall — all behave correctly with both empty and pre-existing user `hooks.json` (e.g. tmux-agent-status setups)
- Total diff: ~390 lines across `bin/install.js` (~80 net), `tests/codex-config.test.cjs` (added 9 e2e tests), `tests/bug-2698-crlf-install.test.cjs` (assertion updates), `CHANGELOG.md`

Both deferred nitpicks are pure polish with no behavioural impact. Happy to take them in a follow-up if preferred.

trek-e (changes requested):
## Adversarial review — Codex API verification + line-traced critique

Applying Gall's Law, Goodhart's Law, Kernighan's Law, Knuth's optimization principle, and Peter Principle. Every line traced. Codex API claims independently verified against [developers.openai.com/codex/hooks](https://developers.openai.com/codex/hooks).

## Track record context

5 prior PRs to this repo: 2 merged (#1346, #1144), **3 closed without merge** (#1763, #1757, #1143). That 60% rejection rate triggers elevated scrutiny — three previous attempts to fix Codex hooks have already failed in this repo (the PR body itself acknowledges #2640 and #2730 closed for sprawl). The fourth attempt deserves more skepticism, not less, because three working hypotheses about "what's wrong with Codex hooks" have already been wrong.

---

## Codex API verification — your premise is partially overstated

You claim:

> Codex 0.124.0+ rejects the inline shape with `invalid type: map, expected a sequence in 'hooks'`, leaving GSD-installed Codex setups unable to start.

I queried the [official Codex hooks docs](https://developers.openai.com/codex/hooks) directly. The truth is narrower than your framing:

> "In practice, the four most useful locations are: `~/.codex/hooks.json`, `~/.codex/config.toml`, `<repo>/.codex/hooks.json`, `<repo>/.codex/config.toml`"

Inline `[[hooks]]` in `config.toml` is **still supported** — but the schema is `[[hooks.SessionStart]]` (nested table arrays keyed by event), not the flat `[[hooks]]\nevent = "SessionStart"` shape GSD historically wrote. Codex 0.124.0 rejects the **flat-with-event-field** shape, not "inline TOML hooks" generically. Your migration is correct in *approach* (hooks.json is canonical) but your *justification* for moving away from TOML conflates two different schemas.

This matters because a future contributor reading the PR body will believe "TOML hooks are broken" — they aren't. They are alive, and the docs explicitly enumerate them as one of four supported locations. The honest framing:

> The flat `[[hooks]]\nevent = "SessionStart"` GSD wrote is a deprecated shape Codex 0.124.0+ rejects. The fix could either (a) emit nested `[[hooks.SessionStart]]` in config.toml, or (b) move to canonical `hooks.json`. We chose (b) because hooks.json is the documented canonical location and matches ecosystem convention (obra/superpowers, oh-my-codex). When both representations exist, Codex warns at startup — single-representation is preferred.

Please update the PR body with the precise framing.

---

## Goodhart violations — tests measure substrings, not behavior

### `countCodexGsdHooks` is a substring counter, not a hook validator

`tests/codex-config.test.cjs:74-90`:

```js
function countCodexGsdHooks(codexHome) {
  const data = readCodexHooksJson(codexHome);
  if (!data || !data.hooks || typeof data.hooks !== 'object') return 0;
  let count = 0;
  for (const event of Object.keys(data.hooks)) {
    if (!Array.isArray(data.hooks[event])) continue;
    for (const entry of data.hooks[event]) {
      if (!entry || !Array.isArray(entry.hooks)) continue;
      for (const h of entry.hooks) {
        if (h && typeof h.command === 'string' && h.command.includes(GSD_CODEX_HOOK_MARKER)) count++;
      }
    }
  }
  return count;
}
```

This counts **any hook in any event slot** whose command contains the substring `gsd-check-update`. What passes this test while shipping a broken install:

1. **Wrong event slot**: A regression that puts the GSD hook into `PostToolUse` instead of `SessionStart` would still count as 1. The dedicated `'hooks.json: GSD hook is registered in canonical schema'` test pins the SessionStart slot, but every other test using `countCodexGsdHooks` is event-blind. If a future refactor breaks the event registration, only one of your fifteen test sites catches it.

2. **Broken path**: A hook with `command = "node /wrong/path/gsd-check-update.js"` — file doesn't exist, hook silently fails at runtime — passes this counter. The `#2153` test (`gsd-check-update.js must exist`) checks file existence but not that the *path in the registered command* matches. It's possible to register a stale path and have the file exist at a different location and have both tests pass.

3. **Substring collision**: A user hook whose command happens to mention `gsd-check-update` (e.g., `echo "running gsd-check-update test"`) counts as a GSD hook. You marked this "deferred" in your round-3 summary — but the helper exports the marker name, encouraging external tests to use the same loose check. At minimum, anchor on `gsd-check-update.js` (with the file extension) rather than the bare token.

**Fix**: rename the helper to `findManagedGsdHookEntries(codexHome)` and return the actual matched entries with their event slot. Then assertions become `assert.strictEqual(matches.length, 1)` AND `assert.strictEqual(matches[0].event, 'SessionStart')` — pinning both shape and slot.

### `'hooks.json: malformed user hooks.json is left untouched'` — under-asserts

`tests/codex-config.test.cjs:1786-1796`:

```js
const malformed = '{ this is not valid json';
fs.writeFileSync(userHooksPath, malformed);
runCodexInstall(codexHome);
assert.strictEqual(fs.readFileSync(userHooksPath, 'utf8'), malformed, 'malformed file preserved');
```

Asserts the malformed file is preserved. Does **not** assert that `config.toml` was *not* mutated, that `codex_hooks` was *not* enabled, or that the install logged a warning rather than a success. This is the round-4 ordering bug exactly — the test passes today but doesn't lock the ordering invariant.

---

## Round 4 unresolved — persistence ordering bug

You marked this PR "ready for human review" at 06:36:52Z. CodeRabbit's round-4 inline comment posted at 06:38:04Z (1 minute later) flagged a real bug:

> "The code currently writes the modified configContent (with codex_hooks set) via fs.writeFileSync before attempting to upsert hooks.json, which can enable codex_hooks even if upsertManagedCodexHookInHooksJson(hooksJsonPath, 'SessionStart', command) fails"

You did not respond. Tracing the diff confirms the bug:

`bin/install.js:6532-6595` (sequence):

```js
fs.writeFileSync(configPath, configContent, 'utf-8');  // ← codex_hooks=true PERSISTED HERE

if (hasEnabledCodexHooksFeature(configContent)) {
  const command = buildHookCommand(targetDir, 'gsd-check-update.js');
  if (upsertManagedCodexHookInHooksJson(hooksJsonPath, command)) {
    console.log(`  ✓ Configured Codex hooks (SessionStart)`);
  } else {
    console.warn(`  ⚠  Skipped hooks.json update — file is malformed or has an unsupported root shape`);
  }
}
```

If a user has a malformed `hooks.json`, this writes `codex_hooks = true` to `config.toml`, then logs a warning that hooks.json wasn't updated. Net effect: Codex starts up with the feature flag enabled but no GSD hook registered. The user sees the warning once during install, then likely never again. On the next `gsd update` run, the same warning fires; `codex_hooks` stays enabled forever.

**Fix**: defer `fs.writeFileSync(configPath, ...)` until after the upsert succeeds. If upsert fails, leave config.toml at its prior content. Alternatively, predicate the config.toml write on the upsert outcome.

Add a regression test for it: `'install does not enable codex_hooks in config.toml when hooks.json upsert fails'`.

---

## Gall's Law — partial credit

You explicitly evolved from the simpler #2640/#2730 attempts and narrowed scope (TOML escape decoding, comment-after-header parsing, malformed-file repair all deliberately deferred). That's correct application of Gall's Law to scope.

But within the PR you make **two** distinct changes that could have been separate:

1. Migrate `[[hooks]]` block out of config.toml.
2. Register hook in canonical hooks.json with full read/parse/merge/write flow.

You could have shipped (1) alone and let users add hook to hooks.json themselves. The 60-line `upsertManagedCodexHookInHooksJson` plus the 80-line `removeManagedCodexHookFromHooksJson` plus the matched-pair test surface (5 install + 6 uninstall scenarios) is a lot of code for a cleanup. You skipped the simpler intermediate.

This isn't disqualifying — coupled (1)+(2) is a reasonable choice if you treat hooks.json as a true contract. But the PR body should justify the choice rather than glossing it.

---

## Kernighan's Law

The migration regex:

```
/(?:\r?\n|^)# GSD Hooks(?:\r?\n)\[\[hooks\]\](?:\r?\n)event = "SessionStart"(?:\r?\n)command = "node [^\r\n]*gsd-(?:check-update|update-check)\.js"(?:\r?\n)/g
```

Edge case I traced:

`command = "node /path/to/gsd-check-update.js extra-arg"` — does NOT match, because `[^\r\n]*\.js` requires `.js` to be followed by `(?:\r?\n)`, and here it's followed by `\s`. Legacy installs that hand-edited the command to add args won't be migrated. Probably never happens in practice, but worth a one-line comment so a future reader knows the regex is filename-anchored, not command-anchored.

The two `changed = true` sites in `removeManagedCodexHookFromHooksJson` are partially redundant:

```js
const filtered = data.hooks[event]
  .map((entry) => {
    ...
    if (entry.hooks.length !== before) changed = true;  // ← inner: hook removed within an entry
    return entry.hooks.length > 0 ? entry : null;
  })
  .filter(Boolean);
if (filtered.length !== data.hooks[event].length) changed = true;  // ← outer: entry dropped
```

When a GSD hook is the only hook in its entry, both fire. Confusing on first read. Either consolidate (track a single "dirty" flag) or comment that the redundancy is intentional.

---

## Knuth optimization — JSON formatting churn

`upsertManagedCodexHookInHooksJson` and `removeManagedCodexHookFromHooksJson` both write:

```js
fs.writeFileSync(hooksJsonPath, JSON.stringify(data, null, 2) + '\n');
```

Forces 2-space indent and trailing newline. If a user authored `hooks.json` with tabs, 4-space indent, or no trailing newline, GSD silently reformats their file on every install. Not a bug, but a user-data churn defect: the user's editor diffs flag every install as a whitespace edit, even when the only logical change is your hook.

**Fix**: detect existing indentation/EOL on read, preserve on write. There's `detectLineEnding` in scope already (you removed its callsite in this PR). Adopt the same pattern for indent.

---

## Peter Principle / scope-of-competence

Your domain knowledge for Codex is strong — the body cites OpenAI's docs URL, references ecosystem convention (obra/superpowers, Yeachan-Heo/oh-my-codex), and acknowledges prior failure modes. That's the opposite of drive-by.

What I'm pushing back on is the **review-readiness threshold**. Marking ready 1 minute before CodeRabbit's round-4 comment is normal timing noise. Not responding to it for 6+ hours after it landed, while leaving the persistence-ordering bug unresolved, is the smell. The bug is real and reachable in production. A first-pass response either fixes it or contests it; silence reads as "ship it before more findings come in."

This is a Peter-Principle-adjacent concern: domain knowledge is high, but PR-management discipline is loose. The earlier closed PRs (#2640, #2730) closed for sprawl suggest the same pattern.

---

## Summary — what to fix before merge

Blocking:
- [ ] Defer `config.toml` write until after `hooks.json` upsert succeeds; rollback on failure (round-4 bug).
- [ ] Add regression test asserting `codex_hooks` is NOT enabled in config.toml when hooks.json upsert fails.
- [ ] Update PR body to clarify that the rejected shape is the flat `[[hooks]]\nevent = ...` form, not "inline TOML hooks generically".

Should fix:
- [ ] `countCodexGsdHooks` should match `gsd-check-update.js` with file extension, not the bare token.
- [ ] At least one assertion per code path that pins `event === 'SessionStart'`, not just the count.
- [ ] Preserve user indentation/EOL in `hooks.json` writes.
- [ ] Comment the migration regex's filename-anchored constraint.

Nit:
- [ ] Consolidate the two `changed = true` sites in `removeManagedCodexHookFromHooksJson`.

The technical core is solid. The hooks.json schema you implement matches the official documented shape. But three previous attempts at this same fix were closed for not-quite-right reasons, and the round-4 bug is real. Address the blocking items and this is mergeable.

trek-e:
## Follow-up — softening the "premise overstated" critique with timeline context

@MiaoDX I dug into the PR history for #2637 and the picture is more nuanced than my review reflected. Sharing what I found and asking you to close the loop with reproduction artifacts.

### Timeline of #2637 fixes

| Date | PR | Author | State | Approach |
|---|---|---|---|---|
| Apr 26 | #2640 | hanzckernel | Closed (sprawl) | Move to `hooks.json` |
| Apr 26 | #2730 | Minghou-Lei | Closed (sprawl) | Move to `hooks.json` + agent schema rewrite |
| Apr 26 | #2747 | trek-e | **Merged**, body says `Closes #2637` | Migrate `[hooks]` map → `[[hooks]]` array, **stayed in config.toml** |
| Apr 27 | #2763 | MiaoDX (this PR) | Open | Move to `hooks.json` |

So three independent contributors have proposed `hooks.json` migrations citing the same error string (`invalid type: map, expected a sequence in 'hooks'`). #2730 said "I reproduced this against a live local Codex install." That's a strong signal the rejection is real.

### Why I called your premise overstated

The official [Codex hooks docs](https://developers.openai.com/codex/hooks) explicitly list inline TOML hooks as supported, with the schema `[[hooks.SessionStart]]` — **nested table-array under the event name**.

What #2747 produces is `[[hooks]]\nevent = "SessionStart"` — **flat array with the event as a field**. That's a *different* TOML shape. The docs example doesn't match it.

So one of two things is true:
1. **Codex 0.124.0+ accepts both shapes** — then #2747's output is fine, your "rejects the inline shape" framing is wrong, and #2637 was correctly closed.
2. **Codex 0.124.0+ rejects flat-with-event but accepts nested `[[hooks.SessionStart]]`** — then #2747 was a partial fix that needs follow-up, and your PR is a legitimate next stage even though "rejects the inline shape" is still imprecise (nested would work; flat doesn't).

Given that two outside contributors (#2640, #2730) reported the same regression after #2747 merged, position 2 is more likely. Which makes your PR a legitimate follow-up rather than a re-litigation of already-decided territory.

### What I'm asking for to close #2637 definitively

To merge this with confidence (and to avoid re-opening the issue a fourth time when the next Codex release ships), we need a reproduction artifact, not opinion. Specifically:

1. **Codex version**: `codex --version` output. Confirm it's ≥ 0.124.0.
2. **Failing config**: a minimal `~/.codex/config.toml` that produces the error, ideally the exact output `bin/install.js` writes today after #2747 merged. Paste it verbatim.
3. **Full error**: not just the substring `invalid type: map, expected a sequence in 'hooks'` — the full stderr trace from `codex` startup, with line/column if Codex reports it.
4. **Working config**: the same config but with the hook moved to nested `[[hooks.SessionStart]]` form (i.e., what #2747 *should* have produced if our reading of the docs is right). Does that startup cleanly? If yes, it confirms the rejection is shape-specific, not "inline broken generically."
5. **Working hooks.json**: the same hook in your PR's hooks.json shape startup-clean? Confirms the alternative.

If items 2 + 3 + 4 + 5 all check out, you've demonstrated:
- #2747's output really is rejected on 0.124.0+
- The alternative `[[hooks.SessionStart]]` shape would also work (so we have two paths)
- Your hooks.json approach works (justifying choice of canonical location over fixing the TOML shape)

That's enough to update the PR body to the precise framing and merge with confidence.

### What to do about my other findings

The other items in my review still stand regardless of which position is right:

- **Round-4 ordering bug** (write `config.toml` before upserting `hooks.json`) — real, reachable, blocking.
- **Test helper Goodhart** (`countCodexGsdHooks` is event-blind across 14 of 15 sites) — real, should fix.
- **JSON formatting churn** on user files — real, should fix.
- **PR body precision** — update once the reproduction confirms which schema actually works.

The track-record / drive-by framing in the review was triggered by the maintainer's elevated-scrutiny request based on prior closure ratios. Now that I see the full timeline (you correctly read the #2640/#2730 closures and intentionally narrowed scope to avoid the same fate), that framing was harsher than warranted. The discipline you showed in scoping this PR is the *opposite* of drive-by behavior.

### Three-attempt pattern signal

Worth flagging to maintainers separately: when three independent contributors all hit the same Codex error message and propose the same fix within 36 hours of each other, the underlying issue probably isn't fully resolved by the merged PR. That's a process signal more than a code signal — issue closure should require either reproduction-of-fix-on-target-version or stronger contributor consensus before auto-closing via `Closes #NNNN`.

trek-e:
**Conflict notice:** PRs #2807 and #2802 both address the same issue (#2637) and modify overlapping files (`bin/install.js`, `tests/codex-config.test.cjs`, `CHANGELOG.md`). Additionally, PR #2809 has already merged a related fix to main. All three competing PRs currently show as CONFLICTING with main. @trek-e — a maintainer decision is needed on which approach to proceed with before any of the three can rebase and re-submit.

trek-e:
closing in favor of maintainer fix. please install the RC5 as mentioned in the issue and test, provide results in a new issue if not resolved.

PR #2807, opened 2026-04-28
fix(#2637): migrate Codex hook to hooks.json and repair PR template links

### LINKED ISSUE #2637: Codex 0.124.0 hook regression can leave GSD-installed Codex setups unable to start

LakshmanTurlapati:
## Fix PR

---

## Linked Issue

Fixes #2637.
Supersedes #2763.
Related: #2802.

The linked issue has the `confirmed-bug` label.

---

## What was broken

GSD installed Codex `SessionStart` hooks into `~/.codex/config.toml` using the flat `[[hooks]]` shape with `event = "SessionStart"`. Codex 0.124.0+ rejects that flat-with-event-field shape on startup; this is not a claim that TOML hooks are unsupported generally.

The default PR template also intentionally shows a “Wrong template” sentinel, but its selector links pointed at `?template=PULL_REQUEST_TEMPLATE/*.md` while the repo’s typed templates use filename-only links like `?template=fix.md`. That made the guidance inconsistent with GitHub’s PR template query behavior and with the repo’s own typed-template links.

## What this fix does

Moves the managed Codex update-check hook to `.codex/hooks.json`, preserves user hooks, removes the legacy GSD flat `[[hooks]]` block during install, and removes only the managed GSD hook during uninstall. It also guards persistence so `config.toml` does not enable `codex_hooks` unless the `hooks.json` upsert succeeds.

Repairs the default sentinel template links so Fix, Enhancement, and Feature selection use `?template=fix.md`, `?template=enhancement.md`, and `?template=feature.md`.

## Root cause

The installer was still writing GSD-managed hooks in a legacy flat TOML shape instead of Codex’s current canonical `hooks.json` structure. The PR-template sentinel had path-prefixed query parameters that did not match the filename-only convention used everywhere else in the repo.

### How I verified the fix

- `git diff --check upstream/main...HEAD`
- `npm test` -> 5778 tests passed, 0 failed
- Checked open PRs before publishing: no existing open PR fixes the PR-template selector links; #2802 overlaps Codex hook schema work but not the template-link issue.

### Regression test added?

- [x] Yes — added/updated Codex install tests for `hooks.json` registration, malformed `hooks.json` handling, uninstall cleanup, and the config-persistence guard.
- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Codex

---

## Checklist

## Breaking changes

None. Existing users migrate away from the legacy flat GSD hook block on the next install, and user-authored `hooks.json` entries are preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hook management during installation by migrating Codex SessionStart hooks to a dedicated configuration file, ensuring better preservation of user-authored hooks during migrations.
  * Enhanced handling of legacy hook configurations across different line ending formats.

* **Documentation**
  * Updated changelog to document hook configuration migration behavior and uninstall procedures.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,9 +6,9 @@ Select the template that matches your PR:
 
 | PR Type | When to use | Template link |
 |---------|-------------|---------------|
-| **Fix** | Correcting a bug, crash, or behavior that doesn't match documentation | [Use fix template](?template=PULL_REQUEST_TEMPLATE/fix.md) |
-| **Enhancement** | Improving an existing feature — better output, expanded edge cases, performance | [Use enhancement template](?template=PULL_REQUEST_TEMPLATE/enhancement.md) |
-| **Feature** | Adding something new — new command, workflow, concept, or integration | [Use feature template](?template=PULL_REQUEST_TEMPLATE/feature.md) |
+| **Fix** | Correcting a bug, crash, or behavior that doesn't match documentation | [Use fix template](?template=fix.md) |
+| **Enhancement** | Improving an existing feature — better output, expanded edge cases, performance | [Use enhancement template](?template=enhancement.md) |
+| **Feature** | Adding something new — new command, workflow, concept, or integration | [Use feature template](?template=feature.md) |
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   it). Re-run `gsd update` without `--minimal` to expand to the full surface. The
   install manifest now records `mode: "minimal" | "full"`. (#2762)
 
+### Fixed
+- **Codex hook configuration moved to `hooks.json` (#2637).** The installer no longer writes an inline `[[hooks]]` block to `~/.codex/config.toml`; Codex 0.124.0+ rejects that shape with `invalid type: map, expected a sequence in 'hooks'`. The managed `SessionStart` hook now lives in `hooks.json` (Codex's canonical location, per https://developers.openai.com/codex/hooks). Re-installing migrates legacy `[[hooks]]` blocks (both `gsd-check-update.js` and the older `gsd-update-check.js` filename, LF/CRLF/mixed EOLs). User-authored hooks in either file are preserved. Uninstall removes only the managed GSD entry.
+
 ## [1.38.5] - 2026-04-25
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   install manifest now records `mode: "minimal" | "full"`. (#2762)
 
 ### Fixed
-- **Codex hook configuration moved to `hooks.json` (#2637).** The installer no longer writes an inline `[[hooks]]` block to `~/.codex/config.toml`; Codex 0.124.0+ rejects that shape with `invalid type: map, expected a sequence in 'hooks'`. The managed `SessionStart` hook now lives in `hooks.json` (Codex's canonical location, per https://developers.openai.com/codex/hooks). Re-installing migrates legacy `[[hooks]]` blocks (both `gsd-check-update.js` and the older `gsd-update-check.js` filename, LF/CRLF/mixed EOLs). User-authored hooks in either file are preserved. Uninstall removes only the managed GSD entry.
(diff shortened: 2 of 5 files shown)


trek-e (commented):
## Adversarial review — PR 2807

Applying Gall's Law, Goodhart's Law, Kernighan's Law, Knuth's optimization principle, and Peter Principle. Traces every line.

---

### Conflict status

PR is currently **CONFLICTING** with main. The most recently merged fix on the same surface is PR #2809 (`fix(#2773)`), which touched `bin/install.js` and `tests/codex-config.test.cjs`. Before any further review, the author must rebase onto main and resolve conflicts. Posting findings now so the author can address them simultaneously.

---

### Relationship to competing PRs

Three PRs target the same issue (#2637): this one (2807), PR #2763 (MiaoDX, already has a trek-e CHANGES_REQUESTED review), and PR #2802 (JOHNNYMACONNY, reviewed separately). The three approaches are architecturally different:

- **PR 2763** — moves hook to `hooks.json`, Trek-e CHANGES_REQUESTED (26h ago, not yet 72h stale)
- **PR 2802** — stays in `config.toml` but senses style from user content only (style-sensing approach)
- **PR 2807** (this PR) — also moves to `hooks.json`, with PR-template link fix bundled in

**Concern (Peter Principle / Gall's Law):** Bundling PR-template link repair into a Codex hook fix is scope creep. Two unrelated concerns are harder to revert independently if one has a defect. The template-link fix is one line per row and should be its own two-line PR.

---

### CodeRabbit actionable findings unaddressed

Three actionable findings from the CodeRabbit review are still open:

1. **`upsertManagedCodexHookInHooksJson` uses `fs.writeFileSync` directly** — not `atomicWriteFileSync`, and not captured in the rollback snapshot. A crash after `fs.writeFileSync` and before `restoreCodexSnapshot` can run leaves a partially-written `hooks.json` with no recovery path. The rest of the installer uses `atomicWriteFileSync` precisely to prevent this class of failure.

2. **Legacy config-cleanup writes even on upsert failure** — The clean-up of legacy `# GSD Hooks` blocks from `config.toml` is guarded by `if (upsertManagedCodexHookInHooksJson(...)) { writeClean } else { warn }`. If `upsertManagedCodexHookInHooksJson` fails (malformed `hooks.json`, permission error, disk full), the legacy TOML block is left in place even though the user's Codex 0.124.0+ will reject it on next startup. CodeRabbit correctly identifies this: the cleanup write should be unconditional (or logged-and-skipped gracefully) rather than gated on upsert success.

3. **Duplicate `### Fixed` heading in CHANGELOG.md** — A second `### Fixed` section was added inside the `[Unreleased]` block rather than merging the entry into the existing one. The CHANGELOG now has two consecutive `### Fixed` headings in Unreleased, which violates Keep a Changelog format.

---

### Test policy violation

The test at `tests/bug-2698-crlf-install.test.cjs:142–166` (mixed-EOL case) verifies legacy block removal from `config.toml` but does not assert that the new `hooks.json` entry was written. The LF and CRLF tests both have this assertion via `hasManagedGsdHookInHooksJson`. The mixed-EOL test should have the same. Without it, a regression where mixed-EOL stripping succeeds but `hooks.json` write fails will pass CI silently. CodeRabbit's nitpick on this is correct — it is a substantive gap, not a style preference.

---

### Goodhart: the `hooks.json` schema shape is the two-level nesting from PR #2809

PR #2809 (already merged) defines the canonical `hooks.json` schema as:

```json
{
  "hooks": {
    "SessionStart": [
      { "hooks": [{ "type": "command", "command": "node ..." }] }
    ]
  }
}
```

This PR's `upsertManagedCodexHookInHooksJson` produces the same two-level shape. Verify the schema exactly matches what PR #2809 shipped — specifically that `data.hooks[event]` is an array of `{ hooks: [...] }` wrapper objects, not a flat array of `{ type, command }` entries. If there is a mismatch, Codex will reject the JSON on load.

---

### Required before merge

1. Rebase onto main — resolve conflicts.
2. Use `atomicWriteFileSync` in `upsertManagedCodexHookInHooksJson` and integrate with the rollback snapshot.
3. Make config-cleanup unconditional (log warning on upsert failure but still write the cleaned TOML).
4. Fix duplicate `### Fixed` heading in CHANGELOG.
5. Add `hasManagedGsdHookInHooksJson` assertion to the mixed-EOL test case.
6. Split PR-template-link fix into its own PR, or document explicitly why it belongs here.

@trek-e — adversarial review posted. Flagging for maintainer decision on which of the three competing #2637 fixes to proceed with.

trek-e:
closing in favor of maintainer fix.   please install the RC5 as mentioned in the issue and test, provide results in a new issue if not resolved.

PR #2870, opened 2026-04-29
fix(#2866): Codex installer strips legacy hooks at EOF without trailing newline

### LINKED ISSUE #2866: Codex Installer (RC.7) fails to strip legacy flat hooks if trailing newline is missing
### GSD Version

1.39.0-rc.7

### Runtime

Codex

### Operating System

macOS

### Node.js Version

v20.11.0

### Shell

_No response_

### Installation Method

## Description
I upgraded to `1.39.0-rc.7` to resolve the Codex configuration error. However, the installer failed to remove my existing legacy `[[hooks]]` block. 

## Root Cause
The cleanup regex in `bin/install.js` (lines 7207-7224) is extremely strict about trailing newlines. If a user's `config.toml` has a legacy block at the very end of the file (missing a trailing `\n`), or if there is any minor whitespace variation, the `replace()` call fails to match.

Because the stale block is not stripped, the installer sees that `gsd-check-update` is already present and skips writing the new **Nested AoT** block.

## Result
Codex (0.125.0+) fails to load with an "invalid type" error because it sees the legacy flat format mixed with (or instead of) the required namespaced format.

## Reproduction TOML
If the user has this at the end of their file:
```toml
# GSD Hooks
[[hooks]]
event = "SessionStart"
command = "node /Users/USER/.codex/hooks/gsd-check-update.js"
```
(No newline after the last line)

The RC.7 installer will leave it untouched.

## Suggested Fix
Use a more lenient regex (e.g. `\s*` instead of `\r?\n`) or, preferably, use structural TOML sensing (as suggested in the closed PR #2802) to identify and remove the block regardless of formatting.

### What did you expect?

The installer should have successfully stripped the legacy `[[hooks]]` block and replaced it with the new namespaced format.

### Steps to reproduce

1. Have a `config.toml` with a legacy flat hook block at the end of the file (missing a trailing newline).
2. 2. Run `npm install -g get-shit-done-cc@1.39.0-rc.7`.
3. 3. Observe that the legacy block is still present in `~/.codex/config.toml`.

### Error output / logs

```shell
invalid configuration: invalid type: map, expected a sequence in `hooks`
```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

### Workaround (if any)

Manually editing `config.toml` to use the new namespaced format.

### Additional context

_No response_

### Privacy Checklist

trek-e:
## Linked Issue

Fixes #2866

## What was broken

The four GSD-managed Codex hook-block strip regexes in \`bin/install.js\` required \`\\r?\\n\` at end. A stale block sitting at end-of-file with no trailing newline (the reporter's exact case) matched none of them — installer logged \"already present\", skipped writing the new Nested-AoT block, and Codex 0.125+ refused to load with \`invalid type: map, expected a sequence in \\\`hooks\\\`\`.

## What this fix does

Each shape's terminator changed from \`\\r?\\n\` to \`(?:\\r?\\n|$)\` so end-of-file counts as a valid terminator.

Strip logic was lifted into a pure helper \`stripStaleGsdHookBlocks(configContent)\` and the install pipeline now calls that helper in place of the inline replace chain. Helper is exported via the \`GSD_TEST_MODE\` module.exports for direct unit-test coverage.

## Root cause

Test gap. The original regression-test fixtures all included a trailing newline. The \`(?:\\r?\\n|^)\` anchor on the leading side correctly handles start-of-file, but the trailing \`\\r?\\n\` was strict — there was no analogous \`(?:\\r?\\n|$)\` to handle end-of-file.

### How I verified the fix

Added \`tests/bug-2866-codex-strip-no-trailing-newline.test.cjs\`. It exercises all four historical shapes:

- Shape 1 — pre-#1755 \`gsd-update-check.js\`
- Shape 2 — flat \`[[hooks]]\` + \`gsd-check-update.js\` (#2637 era, never correct for Codex 0.124+)
- Shape 3 — single \`[[hooks.SessionStart]]\` without nested \`.hooks\` (#2760 era)
- Shape 4 — correct two-block nested

Each shape is driven through \`stripStaleGsdHookBlocks\` twice — once with a trailing newline (regression guard against pre-fix behavior) and once at end-of-file without a trailing newline (reporter's exact repro). It also asserts the helper is a no-op on benign configs, and that Shape 4 stripping does not leave an orphaned \`[[hooks.SessionStart]]\` header.

The test loads \`bin/install.js\` via \`package.json\` \`bin\` field (not hardcoded path) per project convention.

Without the fix, the no-trailing-newline assertions fail. After the fix, all 11 sub-tests pass and the full suite is 5858/5858 (\`npm test\`).

### Regression test added?

- [x] Yes — \`tests/bug-2866-codex-strip-no-trailing-newline.test.cjs\`

### Platforms tested

- [x] macOS — \`npm test\` 5858/5858 locally

The fix uses native JS regex semantics (\`\$\` matches end-of-string under \`g\` flag without \`m\`); no platform-specific behavior involved.

### Runtimes tested

- [x] Codex (the runtime affected by the bug)
- [x] N/A for other runtimes (this code path is gated on Codex install only)

## Checklist

- [x] Issue linked above with \`Fixes #2866\`
- [x] Linked issue has the \`confirmed-bug\` label (added during triage)
- [x] Regression test added
- [x] All existing tests pass (\`npm test\`: 5858/5858)
- [ ] CHANGELOG.md updated — happy to add an entry under \`Fixed\` if maintainers want one before the v1.39.0 final cut

## Breaking changes

None. The new helper produces identical output to the old inline replace chain on every input that previously matched; it additionally strips configurations that previously leaked through.

## Summary by CodeRabbit

* **Bug Fixes**
  * Installer now reliably strips legacy GSD-managed hook configuration blocks (including header comments) and removes any orphaned parent hook tables, even when blocks occur at EOF without a trailing newline.

* **Tests**
  * Added a regression suite validating stripping across multiple legacy hook formats, formatting variations, EOF/no-newline cases, and preservation of unrelated user hook entries and history settings.

* **Chore**
  * Made the block-stripping helper available in test mode for direct validation.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -2896,55 +2896,126 @@ function stripLeakedGsdCodexSections(content) {
 }
 
 /**
- * Strip GSD-managed legacy Codex hook blocks from a config.toml string.
+ * Strip GSD-managed legacy Codex hook blocks from a config.toml string
+ * using the TOML AST already used elsewhere in this file
+ * (`getTomlTableSections` + `removeContentRanges`). The earlier regex-based
+ * implementation required a precise key order, exact single-space padding
+ * around `=`, and exactly one blank line between Shape 4's parent/child
+ * tables — any deviation (an extra blank line, key reorder, an added
+ * `timeout` key, `event="SessionStart"` without spaces) silently leaked the
+ * stale block, sometimes corrupting the file by leaving orphaned key=value
+ * lines outside any table.
  *
- * Targets four historical shapes that GSD installers may have written:
- *
- *   Shape 1 — pre-#1755: flat [[hooks]] entry with the legacy
- *             `gsd-update-check.js` filename.
- *   Shape 2 — #2637 era: flat [[hooks]] entry with the current
- *             `gsd-check-update.js` filename. Never the correct shape for
- *             Codex 0.124.0+, which requires nested `[[hooks.SessionStart]]`.
- *   Shape 4 — correct two-block nested shape that we still want to strip
- *             before re-adding to keep the file clean. Stripped before
- *             Shape 3 to avoid leaving an orphaned `[[hooks.SessionStart]]`
- *             header behind.
- *   Shape 3 — single-block `[[hooks.SessionStart]]` written without the
- *             nested `.hooks` table (#2760 era).
- *
- * Each shape's trailing `(?:\r?\n|$)` accepts either a newline OR
- * end-of-file as terminator, so a stale block at the very end of a config
- * without a trailing newline still strips (#2866).
+ * The structural approach: find every `hooks*` table whose body contains a
+ * `command = "...gsd-(check-update|update-check).js"` value, remove its
+ * exact byte range, and additionally remove any orphaned parent
+ * `[[hooks.SessionStart]]` whose body becomes empty as a result (Shape 4).
+ * The leading `# GSD Hooks` header line is swallowed by extending the
+ * removal range backward through any single preceding comment line.
  *
  * Pure function, exported for test coverage. Returns the input unchanged
- * if no shape matches.
+ * if no GSD-managed hook section is present.
  */
+// Legacy hook command basenames to detect during strip. Template-literal
+// form so install-hooks-copy.test.cjs's quoted-literal guard continues to
+// catch accidental regressions where someone *registers* the inverted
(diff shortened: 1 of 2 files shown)


trek-e:
@coderabbitai full review

PR #2881, opened 2026-04-30
fix(#2876): yamlQuote SKILL.md description for Copilot/Antigravity/Trae/CodeBuddy

### LINKED ISSUE #2876: skills/gsd-ultraplan-phase/SKILL.md: failed to parse YAML frontmatter
### GSD Version

1.38.5

### Runtime

Copilot

### Operating System

macOS

### Node.js Version

v25.9.0

### Shell

/bin/zsh

### Installation Method

### What happened?

When I run `gh copilot` it warns me

This is caused by not quoting the description (the next field, argument-hint, is quoted for presumably the same reason: it starts with a [ which confuses the parser)

### What did you expect?

The skill to parse correctly

### Steps to reproduce

Install GSD
run `gh copilot`

### Error output / logs

```shell
Error: The following skills failed to load:
✖ ~/.copilot/skills/gsd-ultraplan-phase/SKILL.md: failed to parse YAML frontmatter: Unexpected scalar at node end at line 2, column 21: description: [BETA] Offload plan phase to Claude Code's ultraplan cloud — draft… …
```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

### Workaround (if any)

Wrap the description in quotes, eg

```
description: "[BETA] Offload plan phase to Claude Code's ultraplan cloud — drafts remotely while terminal stays free, review in browser with inline comments, import back via /gsd-import. Claude Code only."
```

### Additional context

_No response_

### Privacy Checklist

trek-e:
## Linked Issue

Fixes #2876

## What was broken

Copilot's YAML 1.2-strict frontmatter loader rejects any GSD skill whose description starts with a flow indicator (`[`, `{`, `*`, `&`, `!`, `|`, `>`, `%`, `@`, backtick). The reporter hit this on `gsd-ultraplan-phase`, whose description begins `[BETA] Offload plan phase to Claude Code's…`:

```
✖ ~/.copilot/skills/gsd-ultraplan-phase/SKILL.md: failed to parse YAML
  frontmatter: Unexpected scalar at node end at line 2, column 21:
  description: [BETA] Offload plan phase to Claude Code's ultraplan…
```

The source command at `commands/gsd/ultraplan-phase.md` already quotes its description; the bug was in the converters that re-emit frontmatter for non-Claude runtimes — six sites passed `description: ${description}` through unquoted while the Claude variant already routed through `yamlQuote(...)`.

## What this fix does

Wraps the value in `yamlQuote(...)` (which is `JSON.stringify`, producing a YAML-safe double-quoted scalar) at every emission site:

- `convertClaudeCommandToCopilotSkill`
- `convertClaudeAgentToCopilotAgent`
- `convertClaudeCommandToAntigravitySkill`
- `convertClaudeAgentToAntigravityAgent`
- `convertClaudeCommandToTraeSkill`
- `convertClaudeCommandToCodebuddySkill`

## Root cause

YAML 1.2 §7.3.3 / §6.9: an unquoted scalar that starts with `[`, `{`, `*`, `&`, `!`, `|`, `>`, `%`, `@`, or backtick is an indicator for flow collections / aliases / tags / block scalars. gh-copilot is the strict loader; older Claude variants got away with it because they accept loose parsing. The converters were inconsistent — Claude routed through `yamlQuote`, the rest didn't.

### How I verified the fix

1. Wrote `tests/bug-2876-skill-frontmatter-quote.test.cjs` first (red); confirmed all 46 assertions failed against the pre-fix code, including the reporter's exact `[BETA]` description.
2. Applied the six-site fix.
3. Test now passes (green); also round-trips the value for the four converters that don't substitute runtime names.
4. Updated 7 pre-existing substring assertions in `tests/copilot-install.test.cjs` and `tests/antigravity-install.test.cjs` that hard-coded the unquoted form.

`npm test`: 5893/5893 pass.

### Regression test added?

- [x] Yes — `tests/bug-2876-skill-frontmatter-quote.test.cjs`

### Platforms tested

- [x] macOS — `npm test` 5893/5893 locally

### Runtimes tested

- [x] Copilot (the runtime affected by the bug)
- [x] Antigravity, Trae, CodeBuddy (same code path; verified via test)

## Checklist

- [x] Issue linked above with \`Fixes #2876\`
- [x] Linked issue has the \`bug\` label
- [x] Regression test added
- [x] All existing tests pass

## Breaking changes

None. The new output is YAML-equivalent — every parser that accepted the old unquoted form also accepts the quoted form, and the strict loaders that rejected it now succeed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve skill/agent/command descriptions that start with YAML-like characters by emitting them as quoted frontmatter scalars, preventing misinterpretation.

* **Tests**
  * Added a regression test ensuring descriptions beginning with hostile characters are quoted and round-trip correctly.
  * Updated existing tests to parse frontmatter instead of substring matching.
  * Added and exported a strict frontmatter parser helper for reliable structured assertions.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -1137,7 +1137,11 @@ function convertClaudeCommandToCopilotSkill(content, skillName, isGlobal = false
   }
 
   // Reconstruct frontmatter in Copilot format
-  let fm = `---\nname: ${skillName}\ndescription: ${description}\n`;
+  // #2876: descriptions starting with a YAML flow indicator (`[BETA] …`,
+  // `{ … }`, `*ref`, `&anchor`, etc.) parse as flow sequences/mappings and
+  // crash gh-copilot's frontmatter loader. Always quote so any leading
+  // character is parser-safe.
+  let fm = `---\nname: ${skillName}\ndescription: ${yamlQuote(description)}\n`;
   if (argumentHint) fm += `argument-hint: ${yamlQuote(argumentHint)}\n`;
   if (agent) fm += `agent: ${agent}\n`;
   if (toolsLine) fm += `allowed-tools: ${toolsLine}\n`;
@@ -1223,8 +1227,10 @@ function convertClaudeAgentToCopilotAgent(content, isGlobal = false) {
     ? "['" + uniqueTools.join("', '") + "']"
     : '[]';
 
-  // Reconstruct frontmatter in Copilot format
-  let fm = `---\nname: ${name}\ndescription: ${description}\ntools: ${toolsArray}\n`;
+  // Reconstruct frontmatter in Copilot format. Quote description (#2876)
+  // so a leading YAML flow indicator (`[BETA] …`, `{ … }`, etc.) doesn't
+  // crash the Copilot frontmatter loader.
+  let fm = `---\nname: ${name}\ndescription: ${yamlQuote(description)}\ntools: ${toolsArray}\n`;
   if (color) fm += `color: ${color}\n`;
   fm += '---';
 
@@ -1277,7 +1283,9 @@ function convertClaudeCommandToAntigravitySkill(content, skillName, isGlobal = f
   const name = skillName || extractFrontmatterField(frontmatter, 'name') || 'unknown';
   const description = extractFrontmatterField(frontmatter, 'description') || '';
 
-  const fm = `---\nname: ${name}\ndescription: ${description}\n---`;
+  // #2876: quote description so YAML flow indicators in the source
+  // (e.g. `[BETA] …`) don't break downstream frontmatter parsers.
+  const fm = `---\nname: ${name}\ndescription: ${yamlQuote(description)}\n---`;
   return `${fm}\n${body}`;
 }
 
@@ -1299,7 +1307,8 @@ function convertClaudeAgentToAntigravityAgent(content, isGlobal = false) {
   const claudeTools = toolsRaw.split(',').map(t => t.trim()).filter(Boolean);
   const mappedTools = claudeTools.map(t => convertGeminiToolName(t)).filter(Boolean);
 
-  let fm = `---\nname: ${name}\ndescription: ${description}\ntools: ${mappedTools.join(', ')}\n`;
+  // #2876: quote description for the same reason as the skill variant.
+  let fm = `---\nname: ${name}\ndescription: ${yamlQuote(description)}\ntools: ${mappedTools.join(', ')}\n`;
   if (color) fm += `color: ${color}\n`;
   fm += '---';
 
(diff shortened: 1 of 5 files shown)


trek-e:
## Diagnosis: fix is solid; addressing CR's two nitpicks

**Validation of the fix:**
- All 6 emission sites correctly wrapped in `yamlQuote()` — `JSON.stringify`, which produces a valid YAML 1.2 double-quoted scalar (JSON is a strict YAML 1.2 subset). Round-trip-safe.
- Bug correctly diagnosed (YAML 1.2 §7.3.3/§6.9 — flow indicators `[` `{` `*` `&` `!` `|` `>` `%` `@` backtick require quoting). Strict loaders (gh-copilot) reject; loose ones accept — explains why this bypassed Claude variants for so long.
- Regression test (`tests/bug-2876-skill-frontmatter-quote.test.cjs`) is structurally rigorous: 10 flow-hostile prefixes × 4 command + 2 agent converters, parses frontmatter rather than greps, round-trips through `JSON.parse`.
- CI all green, 5893/5893 pass.

**CR nitpicks — both valid per the project's "tests parse, never grep" convention.** The pre-existing assertions in `copilot-install.test.cjs` (4 sites) and `antigravity-install.test.cjs` (3 sites) only got bumped from `description: Diagnose...` to `description: "Diagnose...`. Both are still raw-string checks against rendered YAML and drift on quoting/order changes — exactly what the convention exists to prevent.

**Pushed follow-up commit** addressing both:
- Added `parseFrontmatter()` to `tests/helpers.cjs` — handles double-quoted (JSON), single-quoted (`''` escape), and bare scalar forms. Throws on missing closed `---` block so emitter regressions fail loudly.
- Refactored all 7 sites: assertions now read `fm.description === 'Diagnose planning directory health'` instead of `result.includes('description: "Diagnose planning directory health')`. Same coverage of #2876, no coupling to byte-level quote style.

`npm test`: 5893/5893 pass after the change.

trek-e:
**CR nitpick disposition (review 03:08:08Z)**

| Finding | File | Verdict | Action |
|---|---|---|---|
| Frontmatter delimiter check too strict | [tests/helpers.cjs:133-140](tests/helpers.cjs:133) — \`split('\n')\` + \`lines[i] === '---'\` | **Valid** — verified: CRLF inputs would leave a trailing \`\r\` on every line, so \`'---\r' === '---'\` fails and the helper throws on valid input. Same problem with whitespace-padded delimiter lines. The helper is used by both \`tests/copilot-install.test.cjs\` and \`tests/antigravity-install.test.cjs\`, so this is also a Windows-runner de-flake. | **Fixed** — \`split(/\r?\n/)\` + \`lines[i].trim() === '---'\` per CR's suggested patch. |

Earlier review (02:12:50Z) flagged \`assert.ok(result.includes(...))\` as raw-string brittleness on \`tests/copilot-install.test.cjs\` and \`tests/antigravity-install.test.cjs\`. That was already addressed by commit \`3b1677bb\` (parses frontmatter structurally via the new \`parseFrontmatter\` helper). The current nitpick is on that helper itself.

5893/5893 on \`npm test\`.

PR #2896, opened 2026-04-30
fix(#2893): surface non-canonical plan filenames instead of silently returning zero plans

### LINKED ISSUE #2893: Planner-produced plan filenames don't match executor's filename filter — phase-plan-index returns empty plans
### GSD Version
1.22.4

### Runtime
Claude Code

### Operating System
macOS

### Node.js Version
v25.2.1

### Shell
/bin/zsh

### Installation Method

### What happened?
After `/gsd:plan-phase` produced a multi-plan decomposition (5 plans for one phase under "Standard" granularity), `/gsd:execute-phase` failed at init with `plan_count: 0`. Investigation showed `phase-plan-index` returned an empty array even though five plan files existed on disk.

The mismatch is between two filename forms:
- The form the planner writes: `{phase-id}-PLAN-{N}-{slug}.md` (e.g., `01-PLAN-01-foundation.md`)
- The form the executor's filter accepts: `{N}-PLAN.md` or `PLAN.md`

`gsd-tools.cjs` at `lib/phase.cjs:231` uses:

```js
files.filter(f => f.endsWith('-PLAN.md') || f === 'PLAN.md')
```

The slugged form ends in `-{slug}.md`, not `-PLAN.md`, so none match. Every downstream call (init → `phase-plan-index` → wave grouping) sees zero plans.

### What did you expect?
Either: (a) the planner produces filenames the executor's filter accepts (e.g., drop the slug from the filename, keeping it in the file's frontmatter/title), or (b) the executor's filter accepts the slugged form (e.g., `^\d+-PLAN(-.*)?\.md$`). The two agents need to agree on the filename contract — currently they don't.

### Steps to reproduce
1. `/gsd:new-project` — init a new project
2. `/gsd:add-phase {anything}` — add a phase
3. `/gsd:plan-phase {N}` with `workflow.granularity = "standard"` so the planner produces ≥2 plans
4. Observe plan files under `.planning/phases/{phase-dir}/` — named like `{phase-id}-PLAN-{N}-{slug}.md`
5. `/gsd:execute-phase {N}`
6. Init reports `plan_count: 0`; running `node $HOME/.claude/get-shit-done/bin/gsd-tools.cjs phase-plan-index "{phase-id}"` returns `{"plans": [], "count": 0}`

### Error output / logs
```
ERROR: Plan detection mismatch — execute-phase blocked
The init reports plan_count: 0 and phase-plan-index returns an empty
plans array, even though N plan files exist in .planning/phases/{phase-dir}/.
```

Every time (100% reproducible).

### Impact
Major — Core feature is broken, no workaround.

### Workaround
Manually rename each plan file to drop the slug (`{phase-id}-PLAN-{N}-{slug}.md` → `{N}-PLAN.md`). Slug content survives in each file's frontmatter and title, so renaming is information-preserving but requires hand-intervention on every multi-plan phase.

`/gsd:health` does not detect this — it validates planning-directory integrity but not plan filename conventions.

trek-e:
## Linked Issue

Closes #2893

## Diagnosis

Reporter saw \`plan_count: 0\` from \`/gsd:execute-phase\` even though five plan files existed on disk. Their planner had written files like \`01-PLAN-01-foundation.md\`, while \`phase-plan-index\`'s strict filter (\`f.endsWith('-PLAN.md') || f === 'PLAN.md'\`) rejected them silently — collapsing two distinct states into the same \`plans: []\` return:

- directory truly has no plans (legit empty)
- directory has plans but the filter rejected them (user/agent error)

The canonical contract is documented in three places:
- \`agents/gsd-planner.md\` \`write_phase_prompt\` step (lines 1063-1080) — explicit \"NEVER use \`PLAN-NN.md\` / \`01-PLAN-01.md\` / \`plan-NN.md\`\"
- \`commands/gsd/plan-phase.md\`
- \`references/universal-anti-patterns.md\` rule 26

It mandates \`{padded_phase}-{NN}-PLAN.md\`. The strict filter is correct per that contract. **The bug is that the executor never tells the user when the contract was violated** — they just see \`plan_count: 0\` with no signal. LLM-driven planners can deviate, especially older versions. Silent-empty is a class-of-bug.

## What this fix does

Adds \`describeNonCanonicalPlans()\` — a diagnostic helper that scans the phase directory for files matching \`*PLAN*.md\` (the diagnostic net) that the canonical filter rejected, **excluding legit derivatives** like \`*-PLAN-OUTLINE.md\` and \`*-PLAN.pre-bounce.md\` so healthy installs don't get false warnings.

When offenders exist, the response gets a \`warning\` field naming each one and citing the canonical pattern. Wired into all three filter sites:
- \`phase-plan-index\` (the executor's main entry point — the symptom site)
- \`phases list --type plans\`
- \`find-phase\`

**The strict filter itself is unchanged.** Existing canonical plans behave identically. This is purely diagnostic — converts silent-empty into loud-with-actionable-error. **Why not loosen the filter to accept \`*PLAN*.md\`?** Because the canonical contract is enforced in three other places. Loosening here would silently accept whatever the model produces and erode the contract. Loud rejection at one end is better than silent acceptance at both.

## Reporter version

Reporter listed \`1.22.4\` in the issue, but per maintainer guidance treat as \`1.38.4\` for fix scoping. The bug is present on current main (1.39.0-rc.4) and earlier — the filter line has been unchanged across these versions.

### How I verified the fix works

Three new tests in \`tests/phase.test.cjs\` under the existing \`phase-plan-index command\` describe block:

1. **\`non-canonical plan filenames surface a warning naming each offender (#2893)\`** — drives the reporter's exact filename pattern (\`01-PLAN-01-foundation.md\`, \`01-PLAN-02-api.md\`), asserts \`plans: []\` is preserved (no silent acceptance), asserts \`warning\` field is a string that names every offender and cites \`{padded_phase}-{NN}-PLAN.md\`.
2. **\`canonical plans suppress the warning even alongside derivative files (#2893)\`** — canonical \`03-01-PLAN.md\` plus \`03-PLAN-OUTLINE.md\` plus \`03-01-PLAN.pre-bounce.md\` (the planner's legit derivatives). Asserts no false-positive warning.
3. **\`empty phase directory returns empty plans array\`** (existing test, extended) — asserts \`warning === undefined\` for truly empty directories.

\`node --test tests/phase.test.cjs\`: 94/94 pass (was 92, +2 new).

\`npm test\`: 5820/5828 pass. The 8 failures are pre-existing failures from \`tests/code-review.test.cjs\` and \`tests/bug-2838-summary-rescue-gitignored-planning.test.cjs\` — they reference user-installed plugin paths under \`~/.claude/get-shit-done/workflows/\` and reproduce on plain \`main\` with my changes stashed. Unrelated to this fix.

### Regression test added?

- [x] Yes — three new tests cover the bug, the false-positive case, and the no-op case.

### Platforms tested

- [x] macOS

### Runtimes tested

- [x] N/A — \`gsd-tools.cjs\` runs the same way under every runtime

## Checklist

- [x] Issue linked above with \`Closes #NNN\`
- [x] Linked issue has the \`bug\` label (added during triage)
- [x] Regression tests added
- [x] All existing \`phase.test.cjs\` tests pass

## Breaking changes

None. The \`warning\` field is additive on the response object. Existing consumers that key off \`plans\` / \`waves\` / \`incomplete\` / \`has_checkpoints\` see no behavioral change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Plan discovery tightened to accept only canonical plan filenames (nonconforming files are ignored).
  * Commands now emit validation warnings that list non-canonical plan files and instruct renaming to the canonical plan filename pattern.

### DIFF
--- a/get-shit-done/bin/lib/phase.cjs
+++ b/get-shit-done/bin/lib/phase.cjs
@@ -8,6 +8,47 @@ const { escapeRegex, loadConfig, normalizePhaseName, comparePhaseNum, findPhaseI
 const { extractFrontmatter } = require('./frontmatter.cjs');
 const { writeStateMd, readModifyWriteStateMd, stateExtractField, stateReplaceField, stateReplaceFieldWithFallback, updatePerformanceMetricsSection } = require('./state.cjs');
 
+// #2893 — strict canonical filter: `{padded_phase}-{NN}-PLAN.md` or `PLAN.md`.
+// Documented in agents/gsd-planner.md (write_phase_prompt step). The wider
+// "looks like a plan but isn't canonical" probe below is used to surface a
+// loud warning instead of silently returning zero plans.
+const isCanonicalPlanFile = (f) => f.endsWith('-PLAN.md') || f === 'PLAN.md';
+
+// Any .md file with PLAN anywhere in the basename — the diagnostic net for
+// catching agent deviations like `01-PLAN-01-foundation.md` (#2893).
+// Excludes derivative files (`-PLAN-OUTLINE.md`, `*.pre-bounce.md`, etc.) that
+// the planner legitimately produces alongside canonical plans.
+const PLAN_OUTLINE_RE = /-PLAN-OUTLINE\.md$/i;
+const PLAN_PRE_BOUNCE_RE = /-PLAN.*\.pre-bounce\.md$/i;
+const looksLikePlanFile = (f) =>
+  /\.md$/i.test(f)
+  && /PLAN/i.test(f)
+  && !PLAN_OUTLINE_RE.test(f)
+  && !PLAN_PRE_BOUNCE_RE.test(f);
+
+/**
+ * Detect plan-shaped files that the canonical filter would reject. Returns
+ * a warning string when offenders exist, else null. Centralised so every
+ * read site (phase-plan-index, phases list --type plans, find-phase) emits
+ * the same message.
+ *
+ * @param {string[]} dirFiles — readdirSync output for one phase directory
+ * @param {string[]} matchedFiles — what the canonical filter accepted
+ * @returns {string|null}
+ */
+function describeNonCanonicalPlans(dirFiles, matchedFiles) {
+  const matched = new Set(matchedFiles);
+  const offenders = dirFiles.filter((f) => looksLikePlanFile(f) && !matched.has(f));
+  if (offenders.length === 0) return null;
+  return (
+    `Found ${offenders.length} plan-shaped file(s) in this phase that don't match the canonical ` +
+    `naming convention "{padded_phase}-{NN}-PLAN.md" (or bare "PLAN.md") and were skipped: ` +
+    offenders.map((f) => `"${f}"`).join(', ') +
+    `. Rename to the canonical form (e.g. "01-01-PLAN.md") so the executor can detect them. ` +
+    `See agents/gsd-planner.md write_phase_prompt step for the full contract.`
+  );
+}
+
 function cmdPhasesList(cwd, options, raw) {
   const phasesDir = path.join(planningDir(cwd), 'phases');
   const { type, phase, includeArchived } = options;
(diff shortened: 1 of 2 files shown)


trek-e:
Force-pushed to drop 3 unrelated canary.yml commits that snuck into the branch from a stale local main snapshot. Diff is now clean — only `get-shit-done/bin/lib/phase.cjs` and `tests/phase.test.cjs`.

@coderabbitai review

trek-e:
Addressed CodeRabbit's parity nitpick. Added 4 tests mirroring the existing #2893 pair across the two uncovered code paths:

- `find-phase`: non-canonical case (warning, names each offender, cites canonical pattern) + canonical case (no warning).
- `phases list --type plans`: same pair, with the additional assertion that the warning is prefixed with `${dir}: ` because that path aggregates across phase directories.

Both pairs exercise the same fixtures as the original `phase-plan-index` tests (reporter's `01-PLAN-01-foundation.md` shape + the canonical+derivatives shape), so all three filter sites are now structurally locked.

`node --test tests/phase.test.cjs`: 98/98 pass (was 94, +4).

PR #2917, opened 2026-04-30
fix(query/agent-skills): emit raw <agent_skills> block instead of JSON-wrapped string

### LINKED ISSUE #2914: agent-skills query ignores agent_skills config, embeds JSON literal in spawn prompts instead of <agent_skills> block [SDK 0.1.0]
## Summary

`gsd-sdk query agent-skills <agentType>` silently ignores the `agent_skills` mapping in `.planning/config.json` and returns a JSON object instead of the raw `<agent_skills>...</agent_skills>` injection block. All 18 GSD workflow files that use `$(gsd-sdk query agent-skills ...)` embed a JSON literal mid-prompt, so project-local skills are never delivered to spawned subagents. The legacy `gsd-tools.cjs` handled this correctly; the SDK port dropped the contract.

Split from #2913 (which covered two bugs — split per one-concern-per-PR policy).

---

## Workflow contract

`workflows/plan-phase.md:36-38` (and ~30 other call sites across 18 workflow files):

```bash
AGENT_SKILLS_PLANNER=$(gsd-sdk query agent-skills gsd-planner)
…
${AGENT_SKILLS_PLANNER}   # embedded directly into the Task() spawn prompt
```

Expected output (what legacy `cmdAgentSkills` / `init.cjs:1678-1692` produced):

```
<agent_skills>
- @skills/dynamic/gsd-planner/SKILL.md
</agent_skills>
```

---

## Repro

```jsonc
// .planning/config.json
{
  "agent_skills": {
    "gsd-planner": ["skills/dynamic/gsd-planner"]
  }
}
```

```bash
# Bug — returns JSON, ignores agent_skills config
$ gsd-sdk query agent-skills gsd-planner
{
  "agent_type": "gsd-planner",
  "skills": ["find-docs", "gsd-add-backlog", …],
  "skill_count": 200+
}

# Expected — raw injection block (legacy CLI still works)
$ node "$HOME/.claude/get-shit-done/bin/gsd-tools.cjs" agent-skills gsd-planner
<agent_skills>
- @skills/dynamic/gsd-planner/SKILL.md
</agent_skills>
```

---

## Root cause

`sdk/src/query/skills.ts` was rewritten as a skill-discovery handler (scans `.claude/skills/`, `.agents/skills/`, etc.). It:

1. **Ignores `config.agent_skills[agentType]`** from `.planning/config.json` entirely.
2. Scans fixed directories that do **not** include project-local paths configured under `agent_skills`.
3. Returns `{ data: { agent_type, skills, skill_count } }`, which `dist/cli.js:277` JSON-stringifies.

The `agent_skills` config-driven injection logic from `buildAgentSkillsBlock` (init.cjs:1602-1672) was never ported.

---

## Proposed fix

Port `cmdAgentSkills` semantics to `sdk/src/query/skills.ts`. Read `agent_skills` from `.planning/config.json`, validate paths (project-relative or `global:` prefix), emit the formatted block via `process.stdout.write` + `process.exit(0)` to bypass `cli.js:277` JSON-stringify. Verified working as a local patch:

```ts
// sdk/src/query/skills.ts
import { existsSync, readFileSync } from 'node:fs';
import { join, resolve, isAbsolute } from 'node:path';
import { homedir } from 'node:os';
import type { QueryHandler } from './utils.js';

function loadAgentSkillsConfig(projectDir: string): Record<string, unknown> | null {
  const configPath = join(projectDir, '.planning', 'config.json');
  if (!existsSync(configPath)) return null;
  try {
    const parsed = JSON.parse(readFileSync(configPath, 'utf-8'));
    return parsed && typeof parsed === 'object' ? parsed : null;
  } catch { return null; }
}

function isPathInside(child: string, parent: string): boolean {
  const sep = process.platform === 'win32' ? '\' : '/';
  const c = resolve(child).toLowerCase();
  const p = resolve(parent).toLowerCase();
  return c === p || c.startsWith(p + sep);
}

function buildAgentSkillsBlock(config: any, agentType: string, projectRoot: string): string {
  if (!config?.agent_skills || !agentType) return '';
  let skillPaths = config.agent_skills[agentType];
  if (!skillPaths) return '';
  if (typeof skillPaths === 'string') skillPaths = [skillPaths];
  if (!Array.isArray(skillPaths) || skillPaths.length === 0) return '';

  const globalSkillsBase = join(homedir(), '.claude', 'skills');
  const validPaths: Array<{ ref: string }> = [];
  for (const skillPath of skillPaths) {
    if (typeof skillPath !== 'string') continue;
    if (skillPath.startsWith('global:')) {
      const skillName = skillPath.slice(7);
      if (!skillName || !/^[a-zA-Z0-9_-]+$/.test(skillName)) continue;
      const md = join(globalSkillsBase, skillName, 'SKILL.md');
      if (!existsSync(md) || !isPathInside(md, globalSkillsBase)) continue;
      validPaths.push({ ref: `${join(globalSkillsBase, skillName)}/SKILL.md` });
      continue;
    }
    if (isAbsolute(skillPath)) continue;
    const dir = resolve(projectRoot, skillPath);
    if (!isPathInside(dir, projectRoot)) continue;
    const md = join(dir, 'SKILL.md');
    if (!existsSync(md)) continue;
    validPaths.push({ ref: `${skillPath}/SKILL.md` });
  }

  if (validPaths.length === 0) return '';
  const lines = validPaths.map(p => `- @${p.ref}`).join('\n');
  return `<agent_skills>\nRead these user-configured skills:\n${lines}\n</agent_skills>`;
}

  const agentType = args[0] || '';
  const config = loadAgentSkillsConfig(projectDir);
  const block = buildAgentSkillsBlock(config, agentType, projectDir);
  // Bypass cli.js:277 JSON.stringify wrapper — workflows need raw text
  if (block) process.stdout.write(block);
  process.exit(0);
};
```

The `process.exit(0)` mirrors legacy `cmdAgentSkills:1688-1691`. If `cli.js` is taught to special-case `agent-skills` output, the exit can be removed.

---

## Environment

- Windows 11 Pro 22631 (bug is platform-independent)
- `get-shit-done-cc@1.38.5` (latest, `npx get-shit-done-cc@latest`)
- `@gsd-build/sdk@0.1.0` (only published version)
- Runtime: Claude Code

## Workaround

Patch `node_modules/@gsd-build/sdk/dist/query/skills.js` with the JS equivalent of the proposed fix above. Verified passing on four expected behaviors: (1) block emitted for mapped agent; (2) byte-match with legacy `cmdAgentSkills`; (3) unmapped agent → empty stdout + exit 0; (4) absolute and out-of-root paths rejected with stderr warning.

Happy to submit a PR once this is labeled `confirmed-bug`.

javeroff:
Fixes #2914.

## What was broken

`gsd-sdk query agent-skills <agent>` returned a JSON-quoted string literal (e.g. `"<agent_skills>\nRead these…"`) instead of the raw `<agent_skills>` block. Workflows that embed the output via `$(gsd-sdk query agent-skills gsd-planner)` injected escaped JSON into spawned subagent prompts, silently breaking all `<agent_skills>` injection on every run.

## What this fix does

Adds an opt-in `format?: 'json' | 'text'` field to `QueryResult`. When a handler returns `format: 'text'`, the CLI writes the string raw via `process.stdout.write` instead of `JSON.stringify`. `agentSkills` sets `format: 'text'` for non-empty blocks. `--pick` always receives JSON regardless, so existing pick consumers are unaffected.

## Root cause

The SDK port of the CLI dispatcher dropped the original `cmdAgentSkills` text-output semantics and unconditionally `JSON.stringify`'d every handler result at `cli.ts:446`. There was no per-result format hint, so the `<agent_skills>` block — which is meant to be embedded raw into a prompt — was serialized as a JSON string like every other query result.

### How I verified the fix

```bash
cd sdk && npm run build && npx vitest run src/query/skills.test.ts
# 12/12 pass including 2 new CLI stdout integration tests

node dist/cli.js query --project-dir /path/to/project agent-skills gsd-planner
# Before: "<agent_skills>\nRead these…"
# After:
# <agent_skills>
# Read these user-configured skills:
# - @skills/dynamic/gsd-planner/SKILL.md
# </agent_skills>
```

### Regression test added?

- [x] Yes — added 2 CLI integration tests in `sdk/src/query/skills.test.ts` that shell out via `execSync` and assert raw stdout (not JSON-wrapped). These would have caught the bug.
- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Windows (including backslash path handling)
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None. `format?` is optional on `QueryResult`; existing handlers default to JSON output unchanged. `--pick` always returns JSON regardless of `format`, so pick-based consumers are unaffected.

---

## Files changed

- `sdk/src/query/utils.ts` — `format?` field on `QueryResult`
- `sdk/src/query/skills.ts` — return `{ data: block, format: 'text' }` for non-empty block
- `sdk/src/cli.ts` — check `result.format === 'text'` in dispatch path
- `sdk/src/query/skills.test.ts` — 2 new CLI integration tests via `execSync` (regression guard)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed gsd-sdk query output so certain handler results (notably agent-skills) are emitted as raw text (e.g., an <agent_skills> XML block) instead of a JSON-quoted string, while preserving existing JSON output behavior (including when using --pick).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- **`gsd-sdk query agent-skills` emits raw `<agent_skills>` block instead of JSON-wrapped string** — workflows that embed via `$(gsd-sdk query agent-skills <agent>)` were receiving a JSON-quoted string literal mid-prompt (e.g. `"<agent_skills>\n…"`), silently breaking all `<agent_skills>` injection into spawned subagents. The CLI dispatcher now honors an opt-in `format: 'text'` field on `QueryResult` and writes such results raw via `process.stdout.write`; `--pick` always returns JSON regardless. (#2917)
 - **`sketch --wrap-up` now dispatches correctly** — `/gsd-sketch --wrap-up` was silently no-oping because the flag dispatch wiring was omitted when the micro-skill entry point was absorbed in #2790. (#2949)
 
 ### Added — 1.40.0-rc.1
--- a/sdk/src/cli.ts
+++ b/sdk/src/cli.ts
@@ -443,7 +443,13 @@ export async function main(argv: string[] = process.argv.slice(2)): Promise<void
           output = extractField(output, pickField);
         }
 
-        console.log(JSON.stringify(output, null, 2));
+        // Handlers can signal format:'text' to emit a raw string (e.g. agent-skills
+        // emits an <agent_skills> XML block workflows embed via $(...) substitution).
+        if (!pickField && result.format === 'text' && typeof output === 'string') {
+          process.stdout.write(output);
+        } else {
+          console.log(JSON.stringify(output, null, 2));
+        }
       }
     } catch (err) {
       if (err instanceof GSDError) {
--- a/sdk/src/query/skills.test.ts
+++ b/sdk/src/query/skills.test.ts
@@ -8,10 +8,14 @@
 
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { mkdtemp, mkdir, rm, writeFile } from 'node:fs/promises';
-import { join } from 'node:path';
-import { tmpdir } from 'node:os';
+import { execSync } from 'node:child_process';
+import { join, resolve } from 'node:path';
+import { tmpdir, homedir } from 'node:os';
+import { fileURLToPath } from 'node:url';
 
 import { agentSkills } from './skills.js';
+
+const CLI = resolve(fileURLToPath(import.meta.url), '../../../dist/cli.js');
 
 async function writeSkill(rootDir: string, name: string) {
   const skillDir = join(rootDir, name);
@@ -119,5 +123,76 @@ describe('agentSkills', () => {
     await writeConfig(tmpDir, { agent_skills: { 'gsd-planner': [] } });
     const r = await agentSkills(['gsd-planner'], tmpDir);
     expect(r.data).toBe('');
+  });
+
+  it('signals format:"text" for non-empty blocks (used by CLI dispatcher)', async () => {
+    await writeSkill(join(tmpDir, '.claude', 'skills'), 'a-skill');
+    await writeConfig(tmpDir, {
+      agent_skills: { 'gsd-planner': '.claude/skills/a-skill' },
+    });
+
+    const r = await agentSkills(['gsd-planner'], tmpDir);
+    expect(r.format).toBe('text');
+  });
+
+  it('does not signal format:"text" for empty result', async () => {
+    const r = await agentSkills(['gsd-planner'], tmpDir);
+    expect(r.format).toBeUndefined();
+  });
+});
+
+// ─── CLI stdout integration ─────────────────────────────────────────────────
+// Regression guard for the JSON-wrapping bug (#2914): the CLI must emit the
+// raw <agent_skills> block to stdout, not a JSON-quoted string.  Spawns the
+// CLI as a child process so the full dispatch path (including cli.ts format
+// handling) is exercised.
+
+describe('agentSkills CLI stdout', () => {
+  let tmpDir: string;
+
+  beforeEach(async () => {
+    tmpDir = await mkdtemp(join(tmpdir(), 'gsd-skills-cli-'));
+  });
+
+  afterEach(async () => {
+    await rm(tmpDir, { recursive: true, force: true });
+  });
+
+  it('writes raw <agent_skills> block to stdout — not JSON-wrapped', async () => {
+    const skillDir = join(tmpDir, '.claude', 'skills', 'cli-skill');
+    await mkdir(skillDir, { recursive: true });
+    await writeFile(join(skillDir, 'SKILL.md'), '# cli-skill\n');
+    await mkdir(join(tmpDir, '.planning'), { recursive: true });
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({ agent_skills: { 'gsd-planner': '.claude/skills/cli-skill' } }),
+    );
+
+    const stdout = execSync(
+      `node "${CLI}" query --project-dir "${tmpDir}" agent-skills gsd-planner`,
+      { encoding: 'utf-8' },
+    );
+
+    expect(stdout).toBe(
+      '<agent_skills>\nRead these user-configured skills:\n- @.claude/skills/cli-skill/SKILL.md\n</agent_skills>',
+    );
+  });
+
+  it('emits empty output (no JSON null) when agent type is unmapped', async () => {
+    await mkdir(join(tmpDir, '.planning'), { recursive: true });
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({ agent_skills: { 'gsd-executor': ['.claude/skills/foo'] } }),
+    );
+
+    const stdout = execSync(
+      `node "${CLI}" query --project-dir "${tmpDir}" agent-skills gsd-planner`,
+      { encoding: 'utf-8' },
+    );
+
+    // Unmapped agent → empty string → CLI falls through to JSON (""), not raw
+    // text. This is acceptable: workflows that embed an empty var are no-ops.
+    // The important invariant is that a MAPPED agent never gets JSON-wrapped.
+    expect(stdout.trim()).toBe('""');
   });
 });
--- a/sdk/src/query/skills.ts
+++ b/sdk/src/query/skills.ts
@@ -123,7 +123,9 @@ export const agentSkills: QueryHandler = async (args, projectDir) => {
   if (validEntries.length === 0) return { data: '' };
 
   const lines = validEntries.map((e) => `- @${e.ref}`).join('\n');
-  return {
-    data: `<agent_skills>\nRead these user-configured skills:\n${lines}\n</agent_skills>`,
-  };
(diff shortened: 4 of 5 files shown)


trek-e (approved):
> *This was generated by AI during PR review.*

## Adversarial Review — APPROVE

**Root cause trace:** ✅ Bug at `cli.ts:446` (unconditional JSON.stringify) → fix opts in via `format: 'text'` field on QueryResult, set by `agentSkills`, honored by dispatcher. 4 files, 3 layers, faithful to the reporter's own follow-up plan on #2914.

**Principle check:**
- **Gall's Law** ✅ — minimal additive change; default behavior unchanged for every other handler
- **Goodhart's Law** ✅ — CLI integration test asserts exact stdout bytes via `execSync`; cannot be gamed by an implementation that merely claims to be text
- **Kernighan's** ✅ — single if/else with three guarded conditions
- **Knuth's** ✅ — no optimization, slightly cheaper than the old path
- **Peter Principle** ⚠️ minor — `QueryResult` (pure data shape) now carries a CLI-consumer concern (`format`). The JSDoc acknowledges this. The alternative (CLI-side allowlist) is worse. Not blocking.

**Findings:**
- HIGH — none
- MEDIUM — `skills.test.ts:184` enshrines `expect(stdout.trim()).toBe('""')` for unmapped agent. The behavior is correct (workflow `${VAR}` interpolation passes literal `""` through to spawn prompts as 6 chars of noise) but that's a separate latent issue worth a follow-up — not a regression introduced here
- MEDIUM — `skills.ts:124` empty-branch returns `{ data: '' }` without `format`; non-empty branch sets `format: 'text'`. Inconsistent but harmless; the dispatcher's `typeof output === 'string'` guard plus `!pickField` handles it. Defensible per your own unit test naming the rationale
- LOW — `cli.ts:447` comment names the consumer in the producer; documentation only

**Secondary regression risk:**
- `--field` extraction: handled (guard `!pickField`)
- Other handlers returning strings: handled (default JSON path, opt-in only)
- `process.stdout.write` lacks trailing newline that `console.log` added — `$(...)` shell substitution strips trailing newlines anyway, so callers are unaffected. The test's exact-byte assertion locks this in
- Windows line endings: SDK builds the block with `\n` literals; portable

**Test rigor:** ✅ Vitest is correct for the SDK subpackage. Real fixture filesystems via `mkdtemp`, spawns the built CLI binary, asserts exact stdout. No `.includes()`, no regex grep on file content. Test is at the layer where the bug actually lives.

**Verdict: APPROVE.** Textbook narrow fix at the right layer. Thanks for the contribution and for the depth in the original bug report — the diagnosis was reusable as the implementation plan.

trek-e:
please update this PR with the contributing.md requirements.   in the future instruct your AI to read it and use it when creating the PR

trek-e:
> *This was generated by AI during triage.*

## Please update PR body to match the Fix template

The PR is approved and CI is green, but the body doesn't follow the [Fix PR template](https://github.com/gsd-build/get-shit-done/blob/main/.github/PULL_REQUEST_TEMPLATE/fix.md) required by [CONTRIBUTING.md](https://github.com/gsd-build/get-shit-done/blob/main/CONTRIBUTING.md#pull-request-guidelines). Per CONTRIBUTING.md: *"Using the wrong template or using the default template for a feature is a rejection reason."*

### Specific gaps

| Template requires | Current PR body |
|---|---|
| `## What was broken` | Has `## What broke` (rename) |
| `## What this fix does` | Has `## Fix` (rename) |
| `## Root cause` | **Missing** — please add a brief explanation of why the bug existed (the SDK port dropped `cmdAgentSkills` semantics) |
| `## Testing` → `### How I verified the fix` | Has `## Verification` (rename + nest) |
| `## Testing` → `### Regression test added?` checkbox | **Missing** — should be checked Yes (you added 2 CLI integration tests) |
| `## Testing` → `### Platforms tested` checkboxes | **Missing** |
| `## Testing` → `### Runtimes tested` checkboxes | **Missing** |
| `## Checklist` (7 gating items) | **Missing entirely** |
| `## Breaking changes` | **Missing** — should state "None" since `format?` is optional |

The technical content is solid and `Fixes #2914` is in place — this is purely a body-formatting update so the typed-template gate doesn't block merge. Could you update the body to follow [fix.md](https://github.com/gsd-build/get-shit-done/blob/main/.github/PULL_REQUEST_TEMPLATE/fix.md)? Once that's in, this is ready to merge.

javeroff:
@trek-e body updated to match `fix.md` — all 7 sections present, checklist included.

A few notes on the boxes I left unchecked so you have full context:

- **CHANGELOG.md** — not updated in this PR. The branch is ~10 commits behind `upstream/main` and upstream already has a `### Fixed` section under `[Unreleased]` that this branch lacks, so adding an entry now would likely conflict on merge. Happy to rebase + add a one-liner if you want it before merge — just say the word.
- **Platforms tested** — verified on Windows only.
- **Runtimes tested** — verified under Claude Code only.

Ready for your re-check whenever you have a moment.

trek-e:
please rebase on main, I'm trying to get a hot fix out this morning and this would be the last one, change logs are a must have because that's what I build release notes off of.

javeroff:
@trek-e rebased onto `upstream/main` (clean — no conflicts) and added the CHANGELOG entry under `[Unreleased] → ### Fixed`. Force-pushed; CI should pick up shortly. Ready for the hot-fix train.

PR #3105, opened 2026-05-04
fix: use #!/usr/bin/env bash in community .sh hooks for distro portability

### LINKED ISSUE #3104: Community .sh hooks use #!/bin/bash; fails on distros without /bin/bash (NixOS, minimal Alpine)
## What's happening

The three opt-in community bash hooks (`gsd-phase-boundary.sh`, `gsd-session-state.sh`, `gsd-validate-commit.sh`) ship with `#!/bin/bash` as their shebang. POSIX guarantees `/bin/sh` but not `/bin/bash`, and several distros don't ship bash at `/bin/bash` by default — most notably NixOS, minimal Alpine container images, and some hardened/container runtimes.

When a user runs the hook scripts directly (testing, manual debugging, or any future installer change that exec's them), the kernel `execve()` returns `ENOENT` because `/bin/bash` doesn't exist.

## Why this is latent in the default install path

In the current default install, hooks are wired in `settings.json` as `bash "<path>"` — bare `bash`, PATH-resolved. When `bash` is invoked with a script argument, it reads the file internally and the script's own shebang is treated as a comment. So today, on a NixOS box where `/bin/bash` doesn't exist but `bash` is on PATH (typically at `/usr/bin/bash`), hooks run fine via the wrapping invocation.

The latent bug surfaces when:

- A test runs the hook directly (e.g. `./hooks/gsd-validate-commit.sh`).
- A future change in `bin/install.js` switches to absolute-path exec (the kind of thing #2979 already had to fix for `node`).
- A user manually invokes the hook from the shell to debug it.
- A docker/CI image strips `/bin/bash` symlinks but keeps `bash` on PATH.

It also creates an inconsistency with the package's own convention: `scripts/*.sh` already use `#!/usr/bin/env bash`. Only `hooks/*.sh` were missed.

## Reproduce

On NixOS (or any distro without `/bin/bash`):

```sh
$ ls -la /bin/bash
ls: cannot access '/bin/bash': No such file or directory

$ which bash
/usr/bin/bash

$ ~/.claude/hooks/gsd-validate-commit.sh < /dev/null
zsh: ./gsd-validate-commit.sh: bad interpreter: /bin/bash: no such file or directory
```

Compare against the same script with the env-bash shebang — works fine.

## Environment

- GSD version: **1.40.0**
- Runtime: **Claude Code**
- OS: **NixOS 26.05** (also affects minimal Alpine container images)
- Node: **v24.14.0**
- Shell: **zsh** (PATH includes `/usr/bin/bash`, no `/bin/bash` symlink)

## Suggested fix

Switch the three community .sh hooks to `#!/usr/bin/env bash` to match the convention already used in `scripts/*.sh`. This is a one-line change per file with no behavioral impact on platforms that DO have `/bin/bash`. The package has a small follow-on of test/comment updates in `tests/bug-2136-sh-hook-version.test.cjs`, `tests/bug-2979-hook-absolute-node.test.cjs`, and `bin/install.js::buildHookCommand` (the comment there cites "POSIX std PATH always includes /bin" as the reason bare `bash` is OK — but the actual reason is that bare `bash` is PATH-resolved, which is what makes it portable).

I have a fork with the fix ready and the full test suite passing (6944/6949, 0 failed, 5 skipped — same skip count as upstream main on Node 24): https://github.com/otavio/get-shit-done/tree/fix/portable-bash-shebang-in-community-hooks

Happy to open the PR once this issue is confirmed (per CONTRIBUTING.md the `confirmed-bug` label is required to avoid auto-close).

otavio:
## Linked Issue

Fixes #3104

## What was broken

The three community .sh hooks (`gsd-phase-boundary.sh`, `gsd-session-state.sh`, `gsd-validate-commit.sh`) ship with `#!/bin/bash`, which fails on distros that don't ship bash at `/bin/bash` (NixOS, minimal Alpine, some container runtimes). POSIX guarantees `/bin/sh` but not `/bin/bash`.

## What this fix does

Switches the three hook shebangs to `#!/usr/bin/env bash`, matching the convention already used in `scripts/*.sh`. Updates two test assertions and a misleading comment in `bin/install.js` that cited "POSIX std PATH always includes /bin" as the rationale (the actual rationale is that bare `bash` is PATH-resolved, which is what makes it portable).

## Root cause

Two separate things came together:

1. The hook .sh files inherited the `#!/bin/bash` form from when they were first added, before `scripts/*.sh` settled on `#!/usr/bin/env bash`.
2. The bug is **latent** in the default install path because `settings.json` wires hooks as `bash "<path>"` (PATH-resolved). When `bash` is invoked with a script argument, it reads the file internally and the shebang is treated as a comment. So on NixOS today, hooks work via the wrapping invocation even though the shebang is wrong.

The latent bug surfaces if a hook is run directly: tests, manual debugging, or any future installer change that exec's the script.

### How I verified the fix

On NixOS 26.05 with Node v24.14.0:

- `npm run build:hooks`: `hooks/dist/*.sh` shebangs propagate correctly.
- `node --test tests/bug-2136-sh-hook-version.test.cjs tests/bug-2979-hook-absolute-node.test.cjs tests/bug-1817-sh-hook-guard.test.cjs tests/bug-1834-sh-hooks-installed.test.cjs tests/bug-1906-hook-relative-paths.test.cjs tests/bug-2557-gemini-local-hook-paths.test.cjs tests/bug-3017-codex-hook-absolute-node.test.cjs tests/security-scan.test.cjs tests/hooks-doc-parity.test.cjs`: 126/126 pass.
- `node scripts/run-tests.cjs` (full suite): 6944 pass / 0 fail / 5 skip.
- Manually executed each hook directly with the new shebang on a NixOS shell — all three resolve and execute (previously: `bad interpreter: /bin/bash: no such file or directory`).

### Regression test added?

- [x] Yes — `tests/bug-2136-sh-hook-version.test.cjs::version header is on line 2 (immediately after shebang)` already asserts the exact shebang on line 1; updated to expect `#!/usr/bin/env bash` so any future regression to `#!/bin/bash` will fail the test.

### Platforms tested

- [x] Linux (NixOS — the affected platform; also benchmarked behavior matches upstream main on Linux x86_64)
- [ ] macOS (no behavior change expected — `/bin/bash` exists everywhere on macOS, so the new env-bash shebang resolves identically)
- [ ] Windows (no behavior change — Git Bash + Windows install path use `bash <path>` PATH lookup; shebang is comment in this path)

### Runtimes tested

- [x] Claude Code

## Checklist

- [x] Linked issue exists and (will) carry the `confirmed-bug` label.
- [x] Single conceptual change, atomic commit.
- [x] Changeset entry under `.changeset/portable-bash-shebang-hooks.md`.
- [x] Tests updated to reflect new expectation.
- [x] Comment cleanup in `bin/install.js::buildHookCommand` aligned with reality.

## Summary by CodeRabbit

* **Documentation**
  * Clarified hook execution environment and shell portability across distributions.

* **Chores**
  * Updated hook script interpreter lines to use PATH-resolved bash for improved cross-platform compatibility.

* **Tests**
  * Adjusted tests to require the portable shebang and updated test comments to match the clarified behavior.

### DIFF
--- a/.changeset/portable-bash-shebang-hooks.md
+++ b/.changeset/portable-bash-shebang-hooks.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3105
+---
+**Community .sh hooks now use `#!/usr/bin/env bash` for cross-distro portability.** The three opt-in bash hooks (`gsd-phase-boundary.sh`, `gsd-session-state.sh`, `gsd-validate-commit.sh`) shipped with `#!/bin/bash`, which fails on distros that don't ship bash at `/bin/bash` (NixOS, minimal Alpine images, some container runtimes). POSIX guarantees `/bin/sh` but not `/bin/bash`. The fix matches the convention already used in `scripts/*.sh`. Latent in the default install path because Claude Code wires hooks as `bash <path>` from `settings.json` (PATH-resolved — the script's own shebang is read as a comment), but the bug surfaces immediately if a hook is run directly (tests, future installer changes, manual debugging). Comment in `bin/install.js::buildHookCommand` updated to clarify that the runner is PATH-resolved bare `bash`, not `/bin/bash` — POSIX std PATH guarantee was the wrong rationale.
--- a/bin/install.js
+++ b/bin/install.js
@@ -709,10 +709,16 @@ function rewriteLegacyCodexHookBlock(content, absoluteRunner) {
  */
 function buildHookCommand(configDir, hookName, opts) {
   if (!opts) opts = {};
-  // .sh hooks run under /bin/bash (POSIX std PATH always includes /bin),
-  // so bare `bash` is fine. .js hooks need the absolute node path because
-  // GUI-launched runtimes start with a minimal PATH that does not include
-  // nvm/Homebrew/Volta-installed node binaries (#2979).
+  // .sh hooks run under bare `bash` (PATH-resolved). POSIX guarantees
+  // /bin/sh but not /bin/bash, and distros like NixOS do not ship
+  // /bin/bash by default — so PATH-resolved `bash` is more portable than
+  // an absolute /bin/bash. The wrapping `bash <path>` invocation also
+  // means the script's own shebang (#!/usr/bin/env bash) is read as a
+  // comment in this code path; it only matters when the script is run
+  // directly (e.g. tests or future installer changes). .js hooks still
+  // need the absolute node path because GUI-launched runtimes start with
+  // a minimal PATH that does not include nvm/Homebrew/Volta-installed
+  // node binaries (#2979).
   const nodeRunner = resolveNodeRunner();
   const runner = hookName.endsWith('.sh') ? 'bash' : nodeRunner;
   // resolveNodeRunner returns null when process.execPath is unavailable.
--- a/hooks/gsd-phase-boundary.sh
+++ b/hooks/gsd-phase-boundary.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # gsd-hook-version: {{GSD_VERSION}}
 # gsd-phase-boundary.sh — PostToolUse hook: detect .planning/ file writes
 # Outputs a reminder when planning files are modified outside normal workflow.
--- a/hooks/gsd-session-state.sh
+++ b/hooks/gsd-session-state.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # gsd-hook-version: {{GSD_VERSION}}
 # gsd-session-state.sh — SessionStart hook: inject project state reminder
 # Outputs STATE.md head on every session start for orientation.
--- a/hooks/gsd-validate-commit.sh
+++ b/hooks/gsd-validate-commit.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # gsd-hook-version: {{GSD_VERSION}}
 # gsd-validate-commit.sh — PreToolUse hook: enforce Conventional Commits format
 # Blocks git commit commands with non-conforming messages (exit 2).
--- a/tests/bug-2136-sh-hook-version.test.cjs
+++ b/tests/bug-2136-sh-hook-version.test.cjs
@@ -100,11 +100,18 @@ describe('bug #2136 part 1: bash hook sources carry gsd-hook-version placeholder
   }
 
   test('version header is on line 2 (immediately after shebang)', () => {
-    // Placing the header immediately after #!/bin/bash ensures it is always
-    // found regardless of how much of the file is read.
+    // Placing the header immediately after the shebang ensures it is always
+    // found regardless of how much of the file is read. The shebang itself
+    // must use `#!/usr/bin/env bash` (PATH-resolved) rather than `#!/bin/bash`
+    // — POSIX guarantees /bin/sh but not /bin/bash, and distros like NixOS
+    // do not ship /bin/bash by default.
     for (const sh of SH_HOOKS) {
       const lines = fs.readFileSync(path.join(HOOKS_DIR, sh), 'utf8').split('\n');
-      assert.strictEqual(lines[0], '#!/bin/bash', `${sh} line 1 must be #!/bin/bash`);
+      assert.strictEqual(
+        lines[0],
+        '#!/usr/bin/env bash',
+        `${sh} line 1 must be "#!/usr/bin/env bash" for cross-distro portability`
+      );
       assert.ok(
         lines[1].startsWith('# gsd-hook-version:'),
         `${sh} line 2 must be the gsd-hook-version header (got: "${lines[1]}")`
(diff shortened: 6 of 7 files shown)


trek-e:
> *This was generated by AI during adversarial review.*
>
> **⏰ If the issues below are not resolved within 72 hours this PR will be closed.**

## Adversarial Review

**What it claims to do:** Switch three community `.sh` hook shebangs from `#!/bin/bash` to `#!/usr/bin/env bash` for NixOS/Alpine portability.

**Verdict: Blocked — test validity concern; misleading comment in `bin/install.js`.**

---

### 🔴 Tests are string comparisons on non-.md files

`tests/bug-2136-sh-hook-version.test.cjs` asserts `lines[0] === '#!/usr/bin/env bash'` — a string equality check on `.sh` file content. This project's test standards require that string/grep comparisons only be used when the file under test is an `.md` file (where text content *is* the product). `.sh` files are executable scripts; their behaviour should be verified by running them in a controlled environment, not by asserting their contents. Either justify the exception explicitly or rewrite as a runtime test.

### 🟡 Minor — `bin/install.js` comment carries a subtle false claim

The updated comment says `.sh` hooks "run under bare `bash` (PATH-resolved)" and implies `#!/usr/bin/env bash` is "more portable." But `buildHookCommand` generates `bash <path>` — meaning the hook's own shebang is **never read at runtime** (it is a comment when the script is invoked as `bash gsd-phase-boundary.sh`). The shebang only matters when the script is run *directly*. The comment should clarify that the portability improvement applies to direct invocation only, not GSD's own hook-launch path, so future readers don't treat it as a general portability guarantee.

trek-e (commented):
> *This review was generated by AI from a multi-agent pass (code, tests, comments, error-handling). Independent agents arrived at the same critical findings.*

## Summary

Tight, correct fix. The three community `.sh` hooks legitimately use bash-only features (`[[...]]`, `=~`, `BASH_REMATCH`, `${#VAR}`), so `#!/bin/sh` would have been wrong — `#!/usr/bin/env bash` is the right choice and aligns with the convention already enforced for `scripts/*.sh` (and documented in CONTRIBUTING.md §Shell scripts). The latent-bug analysis in the PR body is accurate: the wrapping `bash <path>` invocation in `bin/install.js::buildHookCommand` causes the script's shebang to be parsed as a comment, so this fix only matters when the hook is exec'd directly (tests, debugging, future installer changes). The comment rewrite in `buildHookCommand` correctly retracts the misleading "POSIX std PATH always includes /bin" rationale. The regression fence in `tests/bug-2136-sh-hook-version.test.cjs` is an exact-string equality check — any future drift back to `#!/bin/bash` will fail loudly.

Verified locally:
- `git grep -n '^#!.*bash$' -- '*.sh'` returns zero hits across the tree once these three hooks land — completeness is good.
- No `.cmd` / `.ps1` siblings exist for these hooks; Windows path is unaffected.
- Each script's body unambiguously requires bash (validate-commit uses `BASH_REMATCH` and `[[ =~ ]]`, phase-boundary uses `[[ == *glob* ]]`, session-state is bash-style but light).

## 🔴 Critical

None.

## 🟡 Important

None.

## 🟢 Suggestions

**1. `hooks/dist/*.sh` shebangs are stale in the repo (low impact).**
The committed dist artifacts under `hooks/dist/` still start with `#!/bin/bash` because `scripts/build-hooks.js` regenerates them from `hooks/*.sh` on `npm run build:hooks` (called via `prepublishOnly`/CI). For published `npm` consumers this is fine — the tarball is rebuilt before publish, so end users get the new shebang. But anyone running tests against the in-tree dist copy (or doing a non-build install from a git checkout) sees the old shebang. Two reasonable options: (a) commit the regenerated `hooks/dist/*.sh` in this PR so the tree is internally consistent, or (b) leave as-is and rely on the build step. Option (a) is one extra `npm run build:hooks` and `git add hooks/dist`. Confidence ~85.

**2. Regression fence covers source but not dist (low impact, follow-on).**
`bug-2136-sh-hook-version.test.cjs` asserts the exact shebang on `hooks/*.sh` only. A future refactor that bypasses `build-hooks.js` and ships dist independently could regress dist without failing this test. A two-line addition asserting the same shebang on each `hooks/dist/*.sh` (after the existing `runInstaller` step that already populates dist) would close that gap. Not blocking. Confidence ~80.

## Strengths

- **Surgical diff**: three shebang edits + one comment correction + two test-comment refreshes. No drive-by changes, no scope creep.
- **Honest changeset and PR body**: explicitly call out the latency (wrapping `bash <path>` masks the bug) — saves the next maintainer from chasing a phantom regression.
- **Test fence is exact-match** (`assert.strictEqual(lines[0], '#!/usr/bin/env bash', …)`) — superior to a `startsWith('#!')` or regex check; any drift fails immediately with a precise message.
- **Comment rewrite in `bin/install.js`** is the kind of cleanup that's easy to skip; including it here is the right call because the old comment was actively misleading future readers about why bare `bash` was safe.
- Convention alignment with `scripts/*.sh` and CONTRIBUTING.md means there's now a single shebang rule across the repo for shell scripts.

otavio:
Honestly, I don't have any problem to improve or rework the pull request. I am just a little bit confusing about what should I do. Can someone just point me to the right direction so I can learn for future contributions?

trek-e:
Superseded by #3194, which cherry-picks the fix and adds the structural-regression-guard annotation to the test.

PR #3165, opened 2026-05-06
fix(graphify): run /gsd-graphify build inline instead of spawning a sub-agent

requaos:
## Fix PR

---

## Linked Issue

Fixes #3166

> The linked issue must have the `confirmed-bug` label. If it doesn't, ask a maintainer to confirm the bug before continuing.

---

## What was broken

`/gsd-graphify build` spawned a sub-agent that ran `graphify update .` as a background bash process. With graphify v0.7+, the build split into a fast AST-extraction phase followed by a separate clustering + report-write phase. The post-extraction phase was SIGTERM'd when the sub-agent exited, leaving only the cache populated — no `graph.json`, `graph.html`, or `GRAPH_REPORT.md` artifacts were written to `.planning/graphs/`.

## What this fix does

Converts the build to run inline as a single foreground Bash call (no sub-agent), so `graphify update .`, the three artifact copies, the snapshot, and the status report all complete before the command returns.

## Root cause

Sub-agent isolation terminates any background bash left running when the agent exits. Pre-v0.7 graphify completed its work during the AST-extraction phase, which survived long enough. v0.7+ deferred clustering and report writing to a second phase that outlived the agent boundary.

### How I verified the fix

Ran `graphify update .` directly in the main shell — completed in 16 seconds with 369 MB peak memory and produced all three artifacts under `.planning/graphs/`.

### Regression test added?

- [ ] Yes — added a test that would have caught this bug
- [x] No — the fix is in the skill's instruction text (`commands/gsd/graphify.md`), not in CJS code. Runtime orchestration of a CLI tool is not covered by the existing test suite.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] Linked issue has the `confirmed-bug` label

## Breaking changes

None

### DIFF
--- a/.changeset/graphify-build-inline.md
+++ b/.changeset/graphify-build-inline.md
@@ -1,0 +1,4 @@
+---
+type: Fixed
+---
+**`/gsd-graphify build` now runs inline instead of spawning a sub-agent** — required by graphify v0.7+, which split the build into a fast AST-extraction phase followed by a separate clustering + report-write phase. The cached AST phase survived sub-agent isolation, but the post-extraction phase was being SIGTERM'd when the agent exited, leaving only the cache populated and no `graph.json` / `graph.html` / `GRAPH_REPORT.md` artifacts. The skill now runs `graphify update .`, the three artifact copies, the snapshot, and the status report as a single foreground Bash call, surviving to completion. The CLI's `graphify build` pre-flight still returns `action: "spawn_agent"` so external callers and existing tests keep working.
--- a/commands/gsd/graphify.md
+++ b/commands/gsd/graphify.md
@@ -138,7 +138,7 @@ Parse the JSON output:
 
 Display:
 
-```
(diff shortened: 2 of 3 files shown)


trek-e:
this is missing regression tests, to get the fix in the next build I'm going to redo that with the needed tests to keep from regression and validate the work.   I'll mark this as the superseded fix.

trek-e:
Closing in favor of #3169 (now merged as 41dc9bc), which landed the same inline-build fix you diagnosed. Thanks @requaos for the original root-cause analysis and the inline-build approach — that was the right call. #3169 added the structural regression test this PR opted out of, so the fix is now fenced against future re-introduction of the Task-spawn pattern.

PR #3194, opened 2026-05-06
fix(hooks): use #!/usr/bin/env bash in community .sh hooks for portability (#3104)

### LINKED ISSUE #3104: Community .sh hooks use #!/bin/bash; fails on distros without /bin/bash (NixOS, minimal Alpine)

trek-e:
## Linked Issue

Fixes #3104

Supersedes #3105 (otavio's PR — cherry-picked with test annotation fix added).

---

## What was broken

Three community `.sh` hooks (`gsd-phase-boundary.sh`, `gsd-session-state.sh`, `gsd-validate-commit.sh`) used `#!/bin/bash`, which fails on distros that do not ship bash at `/bin/bash` (NixOS, minimal Alpine, some container runtimes).

## What this fix does

Switches the three hook shebangs from `#!/bin/bash` to `#!/usr/bin/env bash`, matching the convention already used in `scripts/*.sh`. Also clarifies the `buildHookCommand` comment in `bin/install.js` to note that the shebang is read as a comment when GSD launches hooks via `bash <path>` — the portability benefit applies to direct invocation of the scripts, not the GSD hook-launch path.

## Root cause

Hook `.sh` files inherited a shebang that hardcodes `/bin/bash`, making them silently non-portable on any distro that places bash elsewhere.

## Testing

Contributor commit (otavio) cherry-picked cleanly. Added `structural-regression-guard` annotation to the shebang assertion (existing `pending-migration-to-typed-ir` annotation is for pre-existing tests in that file; new tests cannot use that category per CONTRIBUTING.md).

### Regression test added?

- [x] Yes — `tests/bug-2136-sh-hook-version.test.cjs` now asserts line 1 is `#!/usr/bin/env bash` for each hook

### Platforms tested

- [x] macOS
- [ ] Linux
- [x] N/A (shebang change; tested by source assertion)

### Runtimes tested

- [x] N/A (hook files, not runtime-specific)

---

## Checklist

- [x] Regression test added
- [x] All existing tests pass
- [x] `.changeset/portable-bash-shebang-hooks.md` added (carried from contributor PR)

## Breaking changes

None.

## Summary by CodeRabbit

* **Chores**
  * Updated community shell hook scripts to use a portable bash shebang for improved cross-platform compatibility.
  * Clarified installation documentation to note shell hooks run via PATH-resolved bash while managed JS hooks use an absolute Node runner.

* **Tests**
  * Expanded tests to validate portable bash shebangs, hook version stamping, and correct handling of managed JS hook commands and edge cases.

### DIFF
--- a/.changeset/portable-bash-shebang-hooks.md
+++ b/.changeset/portable-bash-shebang-hooks.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 3105
+pr: 3194
 ---
 **Community .sh hooks now use `#!/usr/bin/env bash` for cross-distro portability.** The three opt-in bash hooks (`gsd-phase-boundary.sh`, `gsd-session-state.sh`, `gsd-validate-commit.sh`) shipped with `#!/bin/bash`, which fails on distros that don't ship bash at `/bin/bash` (NixOS, minimal Alpine images, some container runtimes). POSIX guarantees `/bin/sh` but not `/bin/bash`. The fix matches the convention already used in `scripts/*.sh`. Latent in the default install path because Claude Code wires hooks as `bash <path>` from `settings.json` (PATH-resolved — the script's own shebang is read as a comment), but the bug surfaces immediately if a hook is run directly (tests, future installer changes, manual debugging). Comment in `bin/install.js::buildHookCommand` updated to clarify that the runner is PATH-resolved bare `bash`, not `/bin/bash` — POSIX std PATH guarantee was the wrong rationale.
--- a/.changeset/portable-bash-shebang-hooks.md
+++ b/.changeset/portable-bash-shebang-hooks.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3105
+---
+**Community .sh hooks now use `#!/usr/bin/env bash` for cross-distro portability.** The three opt-in bash hooks (`gsd-phase-boundary.sh`, `gsd-session-state.sh`, `gsd-validate-commit.sh`) shipped with `#!/bin/bash`, which fails on distros that don't ship bash at `/bin/bash` (NixOS, minimal Alpine images, some container runtimes). POSIX guarantees `/bin/sh` but not `/bin/bash`. The fix matches the convention already used in `scripts/*.sh`. Latent in the default install path because Claude Code wires hooks as `bash <path>` from `settings.json` (PATH-resolved — the script's own shebang is read as a comment), but the bug surfaces immediately if a hook is run directly (tests, future installer changes, manual debugging). Comment in `bin/install.js::buildHookCommand` updated to clarify that the runner is PATH-resolved bare `bash`, not `/bin/bash` — POSIX std PATH guarantee was the wrong rationale.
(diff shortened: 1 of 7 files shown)


trek-e:
@coderabbitai full review

Issue #2833, opened 2026-04-29
enhancement: phase-lifecycle status-line — auto-rotate STATE.md frontmatter as phase orchestrators progress (next/discussing/planning/executing/verifying)


rhettZeng107:
## Pre-submission checklist

- [x] I have confirmed this improves existing behavior — extends the statusline added in #1989
- [x] I have searched existing issues — found related ones (#1102, #1103, #1446, #1572, #612, #956, #2018) but believe this is the first proposal that addresses the systematic gap across all four phase orchestrators uniformly. If I missed a duplicate, please point me to it.
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — backed by a real 4-hour debugging session in our production project today

## Context

I'm a GSD user running a large multi-root migration project (17 phases across 5 sub-repos, currently 10/17 phases complete). Today during routine plan-phase work I hit a chain of three statusline / STATE.md issues that I think point to a real systematic gap. Sharing the experience here in case it's useful — happy to be told this is already on the roadmap or I'm missing context.

## Environment

| Item | Value |
|---|---|
| GSD package | `get-shit-done-cc@1.38.5` (via `npx gsd-sdk`) |
| `gsd_state_version` | 1.0 |
| Statusline hook | `gsd-hook-version: 1.38.6-hc-patch` (local patch on top of 1.38.6 baseline) |
| Claude Code CLI | 2.1.122 |
| Node | v24.14.0, npm 11.9.0 |
| OS | Windows 11 26H2 (running via Git Bash 2.53.0 / MINGW64) |
| Project shape | Multi-root workspace (5 sub-repos, workspace root is not a git repo); decimal phase numbers (4.1, 4.3, 4.5, 4.6, 6.5); 17 phases total in `v2.0` milestone |

## What this enhancement improves

`hooks/gsd-statusline.js` (the one introduced by #1989) **and** the four phase orchestrator workflows (`discuss-phase` / `plan-phase` / `execute-phase` / `verify-phase`) — currently the statusline displays state read from `STATE.md` frontmatter, but the frontmatter fields are not consistently maintained as the orchestrators progress. Result: the statusline drifts from reality, sometimes silently.

## Today's real debugging session — what actually happened

Three things broke in sequence over ~4 hours of work today:

**Issue 1 — `progress.percent` semantics ambiguous → severely misleading number on statusline**

[...]

PR #3241, opened 2026-05-07
fix(executor): detect subagent stalls and STATE drift on resume (#3212)

felipebianchini2006:
## Linked Issue

Fixes #3212

(Confirmed-bug label present on the issue.)

## What was broken

Executor subagents could land production commits, silently stall before the SUMMARY/STATE/ROADMAP close-out commit, and leave the framework with no detector for the stall and no safe-resume path. On resume, `check_incomplete_work` could neither distinguish "never started" from "started, committed code, stalled" nor refuse to re-dispatch over partial work.

## What this fix does

Three additive changes that close the gap without rewriting any dispatch or state core:

- **Stage A — Atomic close-out invariant (docs).** New `docs/ATOMIC-CLOSEOUT-INVARIANT.md` defines the canonical contract (production commits + `SUMMARY.md` + `STATE.md` + `ROADMAP.md` land atomically per plan) with a drift table A–D and reconciliation guidance. New `<atomic_closeout_invariant>` block in `agents/gsd-executor.md` locks the five-step sequence in the agent prompt.
- **Stage B — Safe-resume verification (Problem 1).** New `gsd-sdk query plan.consistency-check` handler (`sdk/src/query/plan-consistency-check.ts`) is read-only and classifies a plan into `consistent_*` / `drift_a..d` / `unknown` by comparing `STATE.md` against disk presence and `git log` commit presence. `get-shit-done/workflows/resume-project.md` `check_incomplete_work` step now invokes the handler before re-dispatching and surfaces drift guidance; the legacy "PLAN without SUMMARY" path is preserved as a fallback.
- **Stage C — Stall surveillance probe (Problem 2).** New `gsd-sdk query dispatch.commit-since` handler (`sdk/src/query/dispatch-surveillance.ts`) is read-only and returns the count of commits with committer-date ≥ `--since` (Unix epoch), optionally filtered by `--plan`. Stateless. No new config key. Orchestrators that can poll on a wall-clock cadence may invoke it during long `Agent()` awaits. `get-shit-done/workflows/execute-phase.md` Step 4 has an additive note documenting the probe.

The probe is **complementary** to the existing `workflow.subagent_timeout` (default 300000 ms) — it does not replace it. Both `tests/subagent-timeout.test.cjs` (13/13) and the timeout config key remain unchanged.

## Root cause

Three independent gaps compound:

1. `get-shit-done/workflows/execute-phase.md` Step 4 awaits `Agent()` synchronously with no commit-progress polling (line 1713: \"No polling — Agent blocks\").
2. `get-shit-done/workflows/resume-project.md` `<step name=\"check_incomplete_work\">` detects \"PLAN without SUMMARY\" via filesystem only — never cross-checks `git log` for production commits, so cannot tell partial-execution from never-started.
3. The four-artifact atomic close-out contract lived only as sequential prose in `agents/gsd-executor.md` with no canonical doc and no verifier.

### How I verified the fix

Ran the new test files plus the adjacent suites the user explicitly flagged as must-not-break.

| Suite | Result |
|---|---|
| `tests/atomic-closeout-invariant.test.cjs` (new) | 8/8 pass |
| `tests/plan-consistency-check.test.cjs` (new) | 6/6 pass |
| `tests/dispatch-surveillance.test.cjs` (new) | 7/7 pass |
| `sdk/src/query/plan-consistency-check.test.ts` (new) | 11/11 pass |
| `sdk/src/query/dispatch-surveillance.test.ts` (new) | 13/13 pass |
| `tests/state.test.cjs` | 101/101 pass |
| `tests/subagent-timeout.test.cjs` | 13/13 pass |
| `tests/state-prune.test.cjs` | 8/8 pass |
| `sdk/src/query/command-static-catalog-domain.test.ts` | 11/11 pass |
| `sdk/src/query/registry-assembly.test.ts` | pass |

The pre-existing `tests/skill-manifest.test.cjs` failure (about unrelated skills like `screenshot`, `transcribe`, `vercel-deploy` not being declared in the manifest) was confirmed pre-existing on `main` via `git stash` — not introduced by this PR.

### Regression test added?

- [x] **Yes** — 21 new CJS tests + 24 new SDK Vitest tests covering all drift states, argv validation, read-only contracts, invariant structure, and surveillance edge cases.

### Platforms tested

- [ ] macOS
- [x] Windows (including backslash path handling — verified during test run on Windows 11; SDK uses `path.join`/`path.sep` everywhere)
- [ ] Linux
- [ ] N/A

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other
- [ ] N/A

(Workflow file edits are runtime-agnostic — both `agents/gsd-executor.md` and the `get-shit-done/workflows/*.md` files are vendored to all runtimes via the existing install pipeline; the new SDK handlers are CLI-level so all runtimes pick them up automatically.)

---

## Checklist

- [x] Issue linked above with `Fixes #3212`
- [x] Regression test added
- [x] All existing tests pass (verified targeted suites; existing `skill-manifest` failure is pre-existing and unrelated)
- [x] `.changeset/` fragment added (`.changeset/sharp-wasps-chatter.md`, type `Fixed`, pr `3212`)

## Breaking changes

None. All changes are additive:

- Two new read-only `gsd-sdk query` handlers (`plan.consistency-check`, `dispatch.commit-since`) — neither shadows an existing alias or modifies existing handler signatures.
- `agents/gsd-executor.md` gains a new `<atomic_closeout_invariant>` block but its existing `<execution_flow>` and `<deviation_rules>` are untouched.
- `get-shit-done/workflows/resume-project.md` `check_incomplete_work` step calls the new handler but preserves its legacy filesystem-only path as a fallback when the handler returns `unknown`.
- `get-shit-done/workflows/execute-phase.md` adds a documentation note in Step 4 describing the optional probe; existing dispatch behavior is unchanged.
- `workflow.subagent_timeout` config key and its tests are untouched.

## Summary by CodeRabbit

* **New Features**
  * Opt-in plan consistency verification to detect incomplete or drifted execution.
  * Optional stall-detection probe that monitors commit progress (dispatch.commit-since) during execution waits.

* **Documentation**
  * Added Atomic Close-Out Invariant guidance and updated executor/resume workflows with recovery advice.

* **Tests**
  * End-to-end and unit tests covering consistency checks and stall-detection behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/get-shit-done/pull/3241)

### DIFF
--- a/.changeset/sharp-wasps-chatter.md
+++ b/.changeset/sharp-wasps-chatter.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3212
+---
+**Executor stall + STATE drift detection** — adds atomic close-out invariant doc, opt-in plan-consistency-check (STATE.md vs disk vs git log) used by resume-project to detect partial-execution drift, and dispatch.commit-since surveillance probe complementing subagent_timeout. All changes additive.
--- a/agents/gsd-executor.md
+++ b/agents/gsd-executor.md
@@ -135,6 +135,22 @@ For each task:
 
 </execution_flow>
 
+<atomic_closeout_invariant>
+**Invariant:** Per-task production commits, the SUMMARY.md, the STATE.md update, and the ROADMAP.md update are a single observational unit. If any one is missing for a plan, the plan is in a *drift* state and the next workflow that touches it MUST reconcile, not redo.
+
+**Sequence (do not reorder):**
+
+1. `<execute_tasks>` — per-task production commits land first.
+2. `<self_check>` — verify those commits exist before writing SUMMARY.
+3. `<summary_creation>` — write SUMMARY.md.
+4. `<state_updates>` — call `gsd-sdk query state.advance-plan`, `state.update-progress`, `roadmap.update-plan-progress`.
+5. `<final_commit>` — single commit covering SUMMARY + STATE + ROADMAP. Returning to the orchestrator before this commit lands is, by definition, a partial close-out.
+
+**Full contract, drift table, and reconciliation guidance:** see `docs/ATOMIC-CLOSEOUT-INVARIANT.md`.
+
+**Stall posture:** if you cannot complete steps 3–5 within your task budget, *do not* return success. Return a checkpoint via `checkpoint_return_format` with the per-task commit hashes already landed so the resume path can reconcile cleanly.
+</atomic_closeout_invariant>
+
 <deviation_rules>
 **While executing, you WILL discover work not in the plan.** Apply these rules automatically. Track all deviations for Summary.
 
--- a/docs/ATOMIC-CLOSEOUT-INVARIANT.md
+++ b/docs/ATOMIC-CLOSEOUT-INVARIANT.md
@@ -1,0 +1,161 @@
+# Plan Atomic Close-Out Invariant
+
+> **Status:** Documented and partially enforced as of v1.50.0 (issue
+> [#3212](https://github.com/gsd-build/get-shit-done/issues/3212)).
+> Documentation is the contract; runtime enforcement is split across
+> `agents/gsd-executor.md` (producer side) and the new `plan.consistency-check`
+> SDK handler (consumer side).
+
+When the GSD executor finishes a plan, **four artifacts must exist together**.
+This document names them, defines what "together" means, and tells you what
+to do if you find them out of sync.
+
+---
+
+## The four artifacts
+
+For every plan `{phase}-{plan}` that the executor reports as complete:
+
+| # | Artifact | Where it lives | What it proves |
+|---|---|---|---|
+| 1 | **Per-task production commits** | `git log --grep="({phase}-{plan})"` | The code changes the plan asked for actually landed |
+| 2 | **`{phase}-{plan}-SUMMARY.md`** | `.planning/phases/XX-name/` | The agent finished the plan, recorded deviations, and self-checked |
+| 3 | **`STATE.md` advanced** | `.planning/STATE.md` | The cursor (`Current Plan`, progress bar) reflects the plan as done |
+| 4 | **`ROADMAP.md` updated** | `.planning/ROADMAP.md` | Cross-phase progress table reflects the plan as done |
+
+The atomic close-out **commit** that lands artifacts 2–4 is a single `git commit`
+created by the executor's `<final_commit>` step (`agents/gsd-executor.md` lines
+659–666). It is intentionally separate from the per-task production commits so
+that resume tooling can distinguish "code landed" from "plan reported done."
+
+---
+
+## What "atomic" means here
+
+"Atomic" in this context is **observational**, not transactional. We cannot
+make four file writes and a git commit physically atomic. Instead, the
+contract is:
+
+> **If artifact 1 (production commits) exists for a plan, then artifacts 2, 3,
+> and 4 must also exist before any other GSD workflow treats that plan as
+> done.**
+
+The framework cannot prevent a partial-write at runtime (network glitch,
+agent stall, kill -9). What it **does** do — newly, post-#3212 — is detect
+the partial state on the next entry into a workflow and refuse to advance.
+
+---
+
+## Possible inconsistent states
+
+| State | Production commits? | SUMMARY.md? | STATE advanced? | ROADMAP updated? | Meaning |
+|---|---|---|---|---|---|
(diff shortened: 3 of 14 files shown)


Ari4ka (dismissed):

trek-e (changes requested):
> *This review was generated by AI from a multi-agent pass (code, tests, error-handling, contract compliance).*

## Summary

Genuinely well-designed — additive, read-only, clean classification table, no shell injection, `execFileSync` with array args throughout, `--grep -F` for literal matching. Two issues need fixing before merge: a source-grep antipattern in tests and a phase-comparison bug that silently misclassifies letter-suffix/decimal phases.

## 🔴 Important (fix before merge)

### 1. Source-grep antipattern in `tests/atomic-closeout-invariant.test.cjs`
**File:** `tests/atomic-closeout-invariant.test.cjs:28-99`

Eight assertions use `content.includes(anchor)` and `assert.match(content, /regex/)` against `ATOMIC-CLOSEOUT-INVARIANT.md` and `agents/gsd-executor.md`. The project rule requires structural parsing (TOML/JSON/frontmatter/heading tree), not source greps. The `// allow-test-rule: docs-presence-check` annotation at line 1 bypasses the lint, but `docs-presence-check` is a reason invented in this PR — it appears nowhere else in the repo — and it fully exempts eight assertions that could be parsed structurally.

**Fix:** For executor prompt anchors (`<atomic_closeout_invariant>`, `<execute_tasks>`, etc.), parse the XML-like blocks and assert on their structure. For the doc, parse the heading tree and assert on headings + table presence rather than substring matching.

### 2. Phase comparison uses string `>` instead of `comparePhaseNum` — silent misclassification for letter-suffix/decimal phases
**File:** `sdk/src/query/plan-consistency-check.ts:1060-1061`

```ts
if (curPhase > askPhase) return true;
```

`curPhase`/`askPhase` are normalized strings. String `>` works for zero-padded numerics but breaks for phases `normalizePhaseName` explicitly handles: `'12A' > '12B'` is alphabetically correct but `'12.1' > '12.10'` is wrong (string comparison), and `'10' > '9'` would also fail without zero-padding. `comparePhaseNum` is already exported from `sdk/src/query/helpers.ts:231` for exactly this purpose.

**Fix:**
```ts
import { comparePhaseNum } from './helpers.js';
// ...
if (comparePhaseNum(curPhase, askPhase) > 0) return true;
```

Projects using letter-suffix or decimal phases will get `unknown` instead of the correct `consistent_complete` or `drift_*` classification without this fix.

## Suggestions

- `dispatch-surveillance.ts:580` — the probe walks `--all` refs; in a multi-branch worktree (this repo uses `.worktrees/`, `.claude/worktrees/`) commits from unrelated branches will count as "progress." This may be intentional for the "is anything moving" signal — if so, note it in the JSDoc so callers don't misinterpret a benign commit on another branch as the dispatched agent being alive.
- `dispatch-surveillance.ts:573-574` — if a commit subject is empty, `secondSpace` is -1 and `latest_subject`/`latest_unix` silently null. Low impact but could mask a stall signal on a close-out commit with no message.
- `plan-consistency-check.ts` reads the prose `**Current Phase:**` / `**Current Plan:**` lines but not the newer `active_phase` / `next_action` frontmatter from `parseStateMd()` (#2833 lifecycle). Projects that migrate to frontmatter-only STATE.md will misclassify `consistent_complete` as `drift_d_phantom_done`. Not a blocker now but worth a follow-up before #2833 lands.

## Strengths

- Truly additive — two new read-only query handlers, no existing handler signatures changed
- No shell injection anywhere: `execFileSync` with array args, `--grep -F` literal mode, `escapeGrepFixed` strips newlines
- Argv validation throws typed `GSDError(Validation)` with specific messages, locked by CJS round-trip tests
- `%ct` (committer date) over `%at` for stall detection is the right call — author-dates lie after cherry-picks/rebases
- Classification truth table matches `ATOMIC-CLOSEOUT-INVARIANT.md` 1:1, `unknown` fallback correctly handles unlisted permutations
- `countProductionCommits` deduplication by hash handles `phase` / `normalizePhaseName(phase)` producing identical grep variants
- SDK Vitest tests cover classification/probe semantics; CJS tests cover CLI dispatch round-trip — correct split for the two test surfaces
- `resume-project.md` integration is non-destructive — legacy filesystem signal preserved as fallback, older SDK installs unaffected

PR #3246, opened 2026-05-08
fix(sdk): phase.add honors --dry-run; rejects unknown flags (#3226)

trek-e:
## Fix PR

Fixes #3226

> The linked issue must have the `confirmed-bug` label.

---

## What was broken

`gsd-sdk query phase.add "desc" --dry-run` treated `--dry-run` as the `customId` positional argument (args[1]), writing real files and modifying ROADMAP.md rather than returning a dry-run preview. Similarly, any unknown `--flag` passed to `phase.add` was silently absorbed into positional slots instead of being rejected with an error.

## What this fix does

Adds a flag parser at the top of `phaseAdd` that:
1. Iterates all args and separates `--` prefixed tokens from positional args.
2. Recognizes `--dry-run` — sets `dryRun = true` and excludes it from positional parsing.
3. Rejects any unrecognized `--flag` with a `Validation` GSDError naming the flag.
4. In dry-run mode, computes the next phase number and roadmap entry string but skips `mkdir`, `writeFile`, and `readModifyWriteRoadmapMd`; returns `{ dry_run: true, roadmap_entry }` alongside the normal result fields.

Symptom B (ROADMAP `### Phase N:` heading scan blocking number reuse) was already correctly handled by the existing `phasePattern` regex. A new regression test confirms it.

Symptom C (`--milestone` / `milestone create`) is out of scope and requires a separate enhancement issue.

## Root cause

`phaseAdd` used strict positional indexing (`args[0]`, `args[1]`) with no flag parsing. Any string in `args[1]` — including `--dry-run` — became the custom phase ID fed into the directory name.

### How I verified the fix

Six new Vitest tests in `sdk/src/query/phase-lifecycle.test.ts`:
- `--dry-run` returns JSON with `dry_run: true` and `roadmap_entry`; ROADMAP unchanged; no directory created.
- `--dry-run` after description (no customId) does not treat the flag as customId.
- `--bogus-flag` throws a GSDError whose message contains the flag name.
- Mixed `--dry-run --unknown` also throws, naming the unrecognized flag.
- `### Phase 5:` heading in ROADMAP with no on-disk dir correctly produces next = 6 (B already fixed).

All 57 `phase-lifecycle.test.ts` tests pass. Full SDK test suite passes.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None

## Summary by CodeRabbit

* **Bug Fixes**
  * `gsd-sdk query phase.add --dry-run` now truly performs a dry run (no file or directory writes) and returns computed phase metadata.
  * Unknown `--flag` arguments to `gsd-sdk query phase.add` are now rejected with a validation error naming the offending flag.

* **Tests**
  * Added tests covering dry-run behavior, argument/flag validation, argument ordering, and correct phase-number computation from existing roadmap content.

### DIFF
--- a/.changeset/nimble-deer-chatter.md
+++ b/.changeset/nimble-deer-chatter.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3246
+---
+**`gsd-sdk query phase.add --dry-run` is now honored** — previously absorbed into the description text and writing real files. Unknown `--flag` arguments now return a validation error instead of silent fallthrough.
--- a/sdk/src/query/phase-lifecycle.test.ts
+++ b/sdk/src/query/phase-lifecycle.test.ts
@@ -550,6 +550,47 @@ describe('phaseAdd', () => {
     // Must detect Phase 5 from ### heading → next = 6, not 1
     expect(data.phase_number).toBe(6);
   });
+
+  // ── Concurrent phase.add: no duplicate IDs (CR finding) ────────────────
+  it('concurrent phase.add calls produce distinct sequential phase numbers', async () => {
+    const { phaseAdd } = await import('./phase-lifecycle.js');
+    await setupTestProject(tmpDir, {
+      phases: ['09-foundation', '10-read-only-queries'],
+    });
+
+    // Fire two phase.add calls simultaneously. If computation happens outside
+    // the lock both will observe maxPhase=10 and claim newPhaseId=11 — collision.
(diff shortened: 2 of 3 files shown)


trek-e:
@coderabbitai full review

PR #3249, opened 2026-05-08
fix(install): require persistent SDK reachability before reporting ready (#3231)

trek-e:
## Fix PR

Fixes #3231

---

## Linked Issue

Fixes #3231

---

## What was broken

`npx get-shit-done-cc@latest` prints `✓ GSD SDK ready (sdk/dist/cli.js)` on Linux but never creates a persistent `gsd-sdk` shim, so every subsequent workflow call fails with `gsd-sdk: command not found`. A stale legacy symlink pointing at the deprecated `gsd-tools.cjs` binary also causes the installer to skip materializing a modern shim.

## What this fix does

1. **Filters transient `_npx` PATH segments** before the initial reachability check. The npx-injected `~/.npm/_npx/<hash>/node_modules/.bin` directory is ephemeral and not reachable from the user's interactive shell — a `gsd-sdk` found there must not count as "on PATH".

2. **Adds `isLegacyGsdSdkShim()`** — sniffs the first 512 bytes of a candidate binary for the `@deprecated` + `gsd-tools.cjs` fingerprint. `isGsdSdkOnPath()` now rejects legacy shims so the installer falls through to self-link with the modern SDK.

3. **Exports `filterNpxFromPath()` and `isLegacyGsdSdkShim()`** for test injection and future callers.

The `getUserShellPath() === null` case (POSIX, `$SHELL` unset or rc timeout) is safe after this fix: `onPath` now reflects the filtered persistent PATH check, which is the best available invariant.

## Root cause

Two independent root causes:

- `isGsdSdkOnPath()` (zero-arg, line 9441) read `process.env.PATH` which npm/npx pollutes with transient `_npx/<hash>/.bin` entries. On Linux, `gsd-sdk` lands there as a secondary bin during `npx` execution — causing a hit that skips `trySelfLinkGsdSdk` entirely. Then `getUserShellPath()` returns null (no `$SHELL`), the cross-shell guard is short-circuited, and the false `✓` is printed.

- A stale legacy symlink at e.g. `~/.local/bin/gsd-sdk → <path>/gsd-tools.cjs` passes the file+exec check but the binary lacks the `query` registry. `isGsdSdkOnPath()` returned true, skipping self-link with the modern SDK.

### How I verified the fix

Three regression tests in `tests/bug-3231-false-gsd-sdk-ready-linux.test.cjs`:

1. **Transient npx PATH + null login-shell PATH** — installer must NOT print "GSD SDK ready" when the only `gsd-sdk` entry is in an `_npx` transient dir.
2. **Stale legacy symlink** — `isLegacyGsdSdkShim()` detects the `@deprecated gsd-tools.cjs` marker; installer replaces stale shim before claiming success.
3. **Clean install** — self-linking into a persistent PATH dir still prints success.

All 105 install-related tests pass (including existing `bug-2775`, `bug-3020`, `bug-3011`, `bug-2678`, `bug-2829`, `sdk-no-sdk-guard` suites).

### Regression test added?

### Platforms tested

- [x] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None — the false-success message is replaced with the existing actionable warning (already shown when `trySelfLinkGsdSdk` fails). No API surface changes for callers that pass an explicit `pathString` to `isGsdSdkOnPath()`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Installer no longer reports "GSD SDK ready" when only a transient package-manager PATH entry or an outdated shim is present; now validates persistent PATH, replaces stale shims, and warns when login-shell PATH probing fails.

* **Tests**
  * Added regression tests covering transient PATH false-positives, legacy shim detection/replacement, and successful persistent installs.

* **Chores**
  * Per-process hook staging introduced and ignore patterns updated to match per-PID staging directories.

### DIFF
--- a/.changeset/quick-voles-sprint.md
+++ b/.changeset/quick-voles-sprint.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3249
+---
+**`✓ GSD SDK ready` no longer prints when no persistent `gsd-sdk` shim exists** — the installer now requires durable reachability (not just transient npx PATH) and replaces stale legacy symlinks pointing at deprecated `gsd-tools.cjs`. Falls back to an actionable warning when login-shell PATH probing fails.
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,8 @@ commands.html
 # Build artifacts (committed to npm, not git)
 hooks/dist/
 
-# Atomic-write staging dir used by scripts/build-hooks.js (see comment there)
-hooks/.dist-staging/
+# Per-process atomic-write staging dirs used by scripts/build-hooks.js (see comment there)
+hooks/.dist-staging-*/
 
 # Coverage artifacts
 coverage/
--- a/bin/install.js
+++ b/bin/install.js
@@ -9437,8 +9437,14 @@ function installSdkIfNeeded(opts) {
   // strictly weaker invariant than the one workflows depend on
   // (`command -v gsd-sdk` resolving), and led to a false ✓ in npx-cache
   // installs (issue #2775).
+  //
+  // #3231: strip transient npx-injected PATH segments before checking. The
+  // installer subprocess PATH includes `~/.npm/_npx/<hash>/node_modules/.bin`
+  // which is ephemeral — it is NOT reachable from the user's interactive
+  // shell. A gsd-sdk found there must NOT count as "on PATH".
   const shimSrc = path.resolve(__dirname, 'gsd-sdk.js');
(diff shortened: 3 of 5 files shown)


trek-e:
@coderabbitai full review

trek-e:
@coderabbitai full review

trek-e:
@coderabbitai full review

Issue #2790, opened 2026-04-27
refactor: consolidate 86 gsd-* skills to ~58 — eliminate redundant entries, fold flags, delete dead skills


trek-e:
## Context

Follow-up to #2789 (description verbosity) and #2762 (minimal install). This issue addresses the *structural* source of skill count bloat: the skill surface was designed for direct invocation, but GSD's own routing meta-skills (`gsd-do`, `gsd-next`, `gsd-progress`) obsolete that assumption. If routing works, users don't need 86 named entry points — they need ~15 coarse capabilities and a router.

**The core contradiction:** `gsd-do`, `gsd-next`, and `gsd-progress` exist precisely because users can't navigate 86 skills. But those 86 skills are in the system prompt *because* they're skills. The routing meta-skills are proof the granularity is wrong.

**Design principle going forward:** If a new skill's function can be expressed as `gsd-existing --flag`, it's a flag, not a skill.

---

### Group 1 — Idea/Task Capture: 5 skills → 1 (`gsd-capture`)

All five are "I have a thought, persist it somewhere." The destination (note vs todo vs backlog vs seed) is a routing decision inside the skill.

| Absorbed skill | Becomes |
|---|---|
| `add-todo` | default mode |
| `note` | `--note` mode |
| `add-backlog` | `--backlog` mode |
| `plant-seed` | `--seed` mode |
| `check-todos` | `--list` mode |

**Saves 4 slots.**

---

### Group 2 — Phase CRUD: 4 skills → 1 (`gsd-phase`)

These are CRUD on a ROADMAP.md row. The append-vs-decimal decision (add vs insert) is internal routing, not a user-facing distinction.

| Absorbed skill | Becomes |
|---|---|
| `add-phase` | default (append) |
| `insert-phase` | `--insert` mode |
| `remove-phase` | `--remove` mode |
| `edit-phase` | `--edit` mode |

**Saves 3 slots.**

---

### Group 3 — Configuration: 4 skills → 1 (`gsd-config`)

All mutate GSD configuration. The basic/advanced split is artificial — a power user who needs advanced settings will find a flag.

| Absorbed skill | Becomes |
|---|---|
| `settings` | default |
| `settings-advanced` | `--advanced` flag |
| `settings-integrations` | `--integrations` flag |
| `set-profile` | `--profile` flag |

**Saves 3 slots.**

---

### Group 4 — Workspace Management: 3 skills → 1 (`gsd-workspace`)

[...]

PR #3250, opened 2026-05-08
fix(capture): restore one-shot --seed contract (#3236)

trek-e:
## Fix PR

---

## Linked Issue

Fixes #3236

## What was broken

`/gsd-capture --seed <idea>` blocked on a mandatory three-question questionnaire (Trigger / Why / Scope) before writing the seed file. Users capturing a stream of ideas lost their flow waiting to answer questions between writes.

## What this fix does

Restores the one-shot capture contract: `plant-seed.md` now writes the seed file immediately with sensible defaults (`trigger_when: when relevant`, `scope: unknown`), then runs optional breadcrumb collection. The Trigger / Why / Scope enrichment questions are preserved as an optional `--enrich` path that runs **after** the file is written, not before.

## Root cause

During skill consolidation (PR #2790 absorbed `/gsd-plant-seed` into `/gsd-capture --seed`), the `plant-seed.md` workflow gained a `gather_context` step positioned at index 2 (before `write_seed` at index 5). This converted an optional enrichment step into a mandatory pre-capture gate. Additionally, all step names used underscores instead of hyphens, violating the CONTEXT.md step-naming rule.

### How I verified the fix

Added `tests/bug-3236-capture-seed-one-shot.test.cjs` — 9 assertions that:
- Confirm `write-seed` step exists with a hyphenated name
- Assert `write-seed` appears in the step order before any Trigger / Why / Scope `AskUserQuestion` call
- Assert `write-seed` appears before any `enrich-seed` / `gather-context` step
- Confirm the workflow documents defaults for `trigger_when` and `scope`
- Confirm the optional enrichment path is still present

All 9 tests were RED before the fix, GREEN after.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux
- [x] N/A (not platform-specific — workflow markdown is platform-independent)

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] N/A (workflow markdown is runtime-agnostic)

---

## Checklist

- [x] Issue linked above with `Fixes #3236`
- [x] `.changeset/` fragment added if this is a user-facing fix (`npm run changeset -- --type Fixed --pr <NNN> --body "..."`) — pending PR number

## Breaking changes

None — this restores the prior one-shot contract. Seeds written with the new workflow are identical in schema to those written before the regression; only the `trigger_when` and `scope` fields now have explicit defaults instead of being mandatory inputs.

## Summary by CodeRabbit

* **Bug Fixes**
  * gsd-capture --seed is one-shot again; Trigger/Why/Scope are now optional and defaulted to sensible values to skip repetitive questionnaires.

* **New Features**
  * Seeds are written immediately with placeholder sections; breadcrumbs are auto-collected afterward. Added an enrich shortcut (--enrich) and a text-mode fallback for enrichment on non-interactive runtimes.

* **Tests**
  * New tests validate one-shot capture, defaults, step ordering, and optional enrichment paths.

### DIFF
--- a/.changeset/sturdy-sloths-hum.md
+++ b/.changeset/sturdy-sloths-hum.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3250
+---
+**`/gsd-capture --seed <idea>` is one-shot again** — Trigger / Why / Scope are optional inputs with sensible defaults instead of a mandatory pre-capture questionnaire. Users capturing a stream of ideas no longer get blocked between writes.
--- a/get-shit-done/workflows/plant-seed.md
+++ b/get-shit-done/workflows/plant-seed.md
@@ -20,7 +20,7 @@ later. The file is never gated behind questions.
 Parse `$ARGUMENTS` for the idea summary.
 
 If empty, ask:
-```
+```text
 What's the idea? (one sentence)
 ```
 
@@ -113,7 +113,7 @@ gsd-sdk query commit "docs: plant seed — {$IDEA}" --files .planning/seeds/SEED
 </step>
 
 <step name="confirm">
-```
+```text
 ✅ Seed planted: SEED-{PADDED}
 
 "{$IDEA}"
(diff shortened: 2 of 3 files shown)


trek-e:
@coderabbitai full review

Issue #3017, opened 2026-05-02
Codex SessionStart hook still emits bare node after #3002 absolute-node fix


hanzckernel:
## GSD Version

Current `main` after PR #3002: `de25400` (`get-shit-done-cc@1.39.0-rc.4`).

## Runtime

Codex

## Operating System

macOS

## Node.js Version

v24.15.0

## Shell

/bin/zsh

## Installation Method

Fresh install from current repository `main` in an isolated temp HOME:

```sh
HOME=<TMP>/home node /tmp/gsd-main-y4yUYX/bin/install.js --codex --global
```

## What happened?

PR #3002 fixed #2979 for settings.json-based managed JS hooks by emitting an absolute Node runner instead of bare `node`. That fix is effective for Gemini and Antigravity, but the Codex install path still bypasses it.

Codex has a separate inline `config.toml` hook block in `bin/install.js`, and current `main` still writes:

```toml
# GSD Hooks
[[hooks.SessionStart]]

[[hooks.SessionStart.hooks]]
type = "command"
command = "node <HOME>/.codex/hooks/gsd-check-update.js"
```

In a GUI/minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) on this machine, that managed hook still fails with exit 127:

```text
/bin/sh: node: command not found
```

So #2979 is only partially fixed after #3002: Gemini and Antigravity fresh installs are fixed, but Codex still has the same bare-node executor failure.

## What did you expect?

Codex-managed JS hooks should follow the same invariant implemented in #3002:

1. Emit an absolute, quoted Node runner for the Codex `gsd-check-update.js` SessionStart hook.
2. If Node cannot be resolved, skip Codex JS hook registration and warn instead of writing a broken hook command.
3. Reinstall/update should rewrite exact legacy Codex managed bare-node hooks to the canonical absolute-runner form.
4. Preserve user-owned Codex hooks/config entries.

## Steps to reproduce

These steps use isolated directories and do not touch real `~/.codex`.

```sh
export REPO=/tmp/gsd-main-y4yUYX
export TMP=$(mktemp -d)
mkdir -p "$TMP/home" "$TMP/work"

HOME="$TMP/home" node "$REPO/bin/install.js" --codex --global

grep -A5 'GSD Hooks' "$TMP/home/.codex/config.toml"

CMD=$(python3 - <<'PY'
import os, pathlib, re
home = pathlib.Path(os.environ['TMP']) / 'home'
[...]

PR #3254, opened 2026-05-08
fix(install): accept Codex TOML floats; idempotent rollback (#3245)

trek-e:
## Fix PR

---

## Linked Issue

Fixes #3245

---

## What was broken

`get-shit-done-cc --codex --global` (v1.41.0+) fails on any Codex `config.toml` containing TOML floats (e.g. `tool_timeout_sec = 20.0`). The hand-rolled TOML parser in `bin/install.js` explicitly rejected floats, then called `restoreCodexSnapshot()` — but that only reverted `config.toml`, leaving skills/, agents/, and VERSION written from this session on disk (hybrid state).

## What this fix does

Two fixes shipped together (they're in the same failure path):

1. **`parseTomlValue` now accepts TOML 1.0 floats.** Decimal, exponent, signed, and underscore-separator forms (`20.0`, `1.5e-3`, `+1.0`, `1_000.0`) are all parsed as JS `Number`. Date/time separators (`-`, `:`, `T`, `Z`) are still rejected. This is the inverse of the prior behavior, which was correct for #2760 CR4 finding 3's concern ("don't silently truncate `0.5` to `0`") but wrong for Codex — Codex CLI's serde schema requires `f64` for `tool_timeout_sec` / `startup_timeout_sec`, so integers are what Codex actually rejects.

2. **`restoreCodexSnapshot` is extended into a unified idempotent rollback.** Pre-install state is captured before the first Codex-specific write (skills/). On any post-install failure the rollback reverts: `config.toml`, `skills/gsd-*` directories, `agents/gsd-*.{md,toml}` files, `get-shit-done/VERSION`, and orphaned atomic-write temp files. Non-`gsd-*` user content is untouched. The rollback is safe to call multiple times and before any snapshots are captured.

## Root cause

`parseTomlValue` matched an integer prefix, then rejected the next character if it was `.`, `e`, or `E` — citing #2760 CR4 finding 3 ("prevent silent integer-prefix truncation of floats"). This was correct as a truncation guard but overcorrected: Codex's serde config schema requires `f64` (float) for timeout fields and fails with "invalid type: integer" if they're integers. The validator fired on the user's config before any GSD content was merged, so any pre-existing float in the user's file caused an abort with a misleading "unsupported TOML value" error.

The partial rollback defect was latent since #2760: `restoreCodexSnapshot` was defined after skills/agents/VERSION were already written, and only knew about `config.toml`.

### How I verified the fix

- `tests/bug-3245-codex-toml-floats.test.cjs` — 18 new regression tests:
  - `parseTomlToObject` accepts all TOML float forms (`20.0`, `60.0`, `1e10`, `1.5e-3`, `+1.0`, `-0.5`, `1_000.0`)
  - Integers still parse correctly; date/time/hex literals still rejected
  - `validateCodexConfigSchema` passes a config with `tool_timeout_sec = 20.0`
  - Full install completes without throwing when `config.toml` contains floats
  - Post-install config retains float value as numeric `20`
  - Validation failure via `__codexSchemaValidator` seam rolls back skills/, agents/, VERSION
  - Rollback is idempotent and safe when no prior content exists
  - Non-`gsd-*` user skills survive rollback
  - Orphaned atomic-write temp files are cleaned up
- `tests/bug-2760-codex-install-defensive.test.cjs` — updated the one test that explicitly rejected floats (the truncation-guard test), replacing it with a test that verifies floats are now properly parsed (not truncated). All 29 tests pass.
- Full `npm test`: 7973/7975 pass; the 2 failures are pre-existing local environment issues (Homebrew Node.js symlink path mismatch in `#3017` e2e test, unrelated to this change).

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Codex (this is Codex-only)

---

## Checklist

- [ ] `.changeset/` fragment added if this is a user-facing fix (`npm run changeset -- --type Fixed --pr <NNN> --body "..."`) — or `no-changelog` label applied

## Breaking changes

None

## Summary by CodeRabbit

* **Bug Fixes**
  * TOML floating-point values (e.g., tool_timeout_sec = 20.0, exponent/signed/underscore forms) are now accepted.
  * Installation rollback now restores/removes config, skills, agents, VERSION, and cleans installer temp files on validation failures.
  * Rollback is idempotent and avoids deleting non-installer user data.

* **Tests**
  * Added regression tests for TOML floats, install round-trip, robust/idempotent rollback, and frontmatter extraction edge cases.

### DIFF
--- a/.changeset/fierce-birds-wake.md
+++ b/.changeset/fierce-birds-wake.md
@@ -2,4 +2,4 @@
 type: Fixed
 pr: 3254
 ---
-**`get-shit-done-cc --codex` no longer rejects valid TOML floats** — `tool_timeout_sec = 20.0` (which Codex CLI's serde schema actually requires) is now preserved instead of triggering a half-rolled-back install. Rollback also covers skills/, agents/, and VERSION on validation failure.
+**`get-shit-done-cc --codex` no longer rejects valid TOML floats** — `tool_timeout_sec = 20.0` (which Codex CLI's serde schema actually requires) is now preserved instead of triggering a half-rolled-back install. On any post-install validation failure, rollback now covers all five mutation surfaces: `skills/` (gsd-* skill dirs), `agents/` (gsd-*.md/.toml files), `VERSION`, `config.toml`, and any orphaned atomic-write temp files left by an aborted write.
--- a/.changeset/fierce-birds-wake.md
+++ b/.changeset/fierce-birds-wake.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3254
+---
+**`get-shit-done-cc --codex` no longer rejects valid TOML floats** — `tool_timeout_sec = 20.0` (which Codex CLI's serde schema actually requires) is now preserved instead of triggering a half-rolled-back install. Rollback also covers skills/, agents/, and VERSION on validation failure.
--- a/bin/install.js
+++ b/bin/install.js
@@ -7460,14 +7460,25 @@ function install(isGlobal, runtime = 'claude') {
   //
   // Snapshot contents:
   //   codexPreInstallSkillNames  — Set of gsd-* skill dir names that existed
+  //   codexPreInstallSkillContents — Map<skillName, Map<relPath, Buffer>> of
+  //       the full file tree of each pre-existing gsd-* skill dir, so that
+  //       overwritten dirs can be fully restored on rollback (not just removed).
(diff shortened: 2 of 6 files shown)


trek-e:
@coderabbitai full review

PR #3272, opened 2026-05-09
fix(3259): non-mutating --help guard for native query handlers

### LINKED ISSUE #3259: bug(sdk): milestone.complete --help executes mutation instead of rendering help
## Summary

`gsd-sdk query milestone.complete --help` does not render contextual help. The SDK native handler treats `--help` as the milestone version and executes the mutation.

## Reproduction

From a checkout with the built SDK:

```bash
node sdk/dist/cli.js query milestone.complete --help
```

Observed output:

```json
{
  "version": "--help",
  "name": "--help",
  "milestones_updated": true
}
```

In my local repro this also wrote `.planning/MILESTONES.md` with a shipped milestone named `--help --help`. I removed that local artifact after confirming the behavior.

## Expected

`--help` / `-h` should be non-mutating and render contextual help for `milestone.complete`, or at minimum be rejected before the mutation handler writes anything.

## Notes from source inspection

- `sdk/src/cli.ts` intentionally forwards `--help` inside `queryArgv` for contextual handler help (#3019).
- `milestone.complete` is registered as a native SDK handler in `sdk/src/query/command-static-catalog-domain.ts`, so it does not fall back to the CJS help path.
- `sdk/src/query/phase-lifecycle.ts` `milestoneComplete` reads `args[0]` as `version` and does not guard help flags.
- `sdk/src/query/query-dispatch.ts` dispatches matched native handlers directly, so the handler receives `--help` as a positional arg.

## Suggested fix

Add a non-mutating help guard for native query handlers, especially mutating commands, before calling the handler. At minimum, `milestoneComplete` should reject or render help when `args[0]` is `--help` / `-h`. A regression test should assert that `gsd-sdk query milestone.complete --help` does not create/update `.planning/MILESTONES.md`.

trek-e:
## Fix PR

---

## Linked Issue

Fixes #3259

> The linked issue must have the `confirmed-bug` label. If it doesn't, ask a maintainer to confirm the bug before continuing.

---

## What was broken

`gsd-sdk query milestone.complete --help` executed the mutation instead of rendering help: it treated `--help` as a milestone version, wrote `--help` entries to `.planning/MILESTONES.md`, archived ROADMAP/REQUIREMENTS into `.planning/milestones/--help/`, and rewrote `STATE.md`. Every native mutating handler that reads positional args without sanitization was vulnerable to the same class of bug.

## What this fix does

**Dispatcher-level guard** (`sdk/src/query/query-dispatch.ts`): Before invoking a matched native handler, if the args contain `--help` or `-h` AND the matched command is marked `mutation: true` in the command manifest, dispatch short-circuits to a non-mutating JSON stub `{ "help": "Usage: gsd-sdk query <command> [args...]" }`. Mutating handlers are not help-aware by default (fail-closed).

**Handler-level defense-in-depth** (`sdk/src/query/phase-lifecycle.ts`): `milestoneComplete` now explicitly rejects `--help` and `-h` as the version argument (throwing `GSDError` with `ErrorClassification.Validation`) before any disk write, regardless of whether the dispatcher guard intercepted first.

The `cli.ts` #3019 contract (forwarding `--help` into queryArgv for contextual help) is preserved unchanged. Unknown commands with `--help` still fall through to the CJS fallback.

## Root cause

The `cli.ts` #3019 contract forwards `--help` into `queryArgv` so handlers can render contextual help. The dispatcher at `query-dispatch.ts` passed args straight to native handlers with no interception. `milestoneComplete` read `args[0]` as the milestone version with only an empty-string guard — no check for flag values like `--help`.

### How I verified the fix

- `milestone.complete --help` via `runQueryDispatch` returns a non-mutating JSON stub and leaves all `.planning/` files unmodified (digest comparison of mtimes before/after).
- `milestone.complete -h` same.
- Registry-driven test: iterates every registered native mutating command (`COMMAND_MUTATION_SET`) and asserts `.planning/` is byte-identical (mtime digest) before/after calling with `--help`.
- Handler-level: `milestoneComplete(['--help'], tmpDir)` throws `GSDError` with `ErrorClassification.Validation` before STATE.md or MILESTONES.md are modified.
- Handler-level: same for `['-h']`.
- Preservation of #3019: `unknown-cmd --help` still falls through to CJS fallback and returns the script's stdout.
- Non-mutating handlers (e.g. `state.json`) with `--help` in args pass through to the handler unchanged.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] `.changeset/` fragment added if this is a user-facing fix (`npm run changeset -- --type Fixed --pr <NNN> --body "..."`) — or `no-changelog` label applied

## Breaking changes

None

## Summary by CodeRabbit

* **Bug Fixes**
  * Mutating commands with `--help` or `-h` now return help output without performing any writes.
  * Handlers now reject help flags when they would be misinterpreted as argument values.

* **Tests**
  * Added coverage asserting help flags do not cause filesystem mutations and that validation errors are raised.

* **Documentation**
  * Added a changeset documenting the fix.

### DIFF
--- a/.changeset/calm-herons-wake.md
+++ b/.changeset/calm-herons-wake.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3272
+---
+**`gsd-sdk query milestone.complete --help` (and all mutating query handlers) no longer execute mutations** — the dispatcher now short-circuits to a non-mutating help stub when `--help`/`-h` appears in args for any native mutating handler (dispatcher-level guard, fail-closed by default). `milestoneComplete` also rejects `--help`/`-h` as a version value before any disk write (handler-level defense-in-depth).
--- a/sdk/src/query/phase-lifecycle.test.ts
+++ b/sdk/src/query/phase-lifecycle.test.ts
@@ -1505,6 +1505,68 @@ describe('phasesArchive', () => {
   });
 });
 
+// ─── milestoneComplete help-flag defense (#3259) ────────────────────────────
+
+describe('milestoneComplete help-flag defense', () => {
+  it('rejects --help as a version value with GSDError before any disk write', async () => {
+    const { milestoneComplete } = await import('./phase-lifecycle.js');
+    const { GSDError, ErrorClassification } = await import('../errors.js');
+    await setupTestProject(tmpDir);
+
+    // Capture pre-invocation filesystem state
+    const planningDir = join(tmpDir, '.planning');
+    const milestonesPath = join(planningDir, 'MILESTONES.md');
+    const statePath = join(planningDir, 'STATE.md');
+    const preStateStat = await import('node:fs').then((m) => m.statSync(statePath));
+    const milestonesExistedBefore = existsSync(milestonesPath);
+
+    let thrown: unknown;
+    try {
+      await milestoneComplete(['--help'], tmpDir);
+    } catch (e) {
+      thrown = e;
+    }
+
+    expect(thrown).toBeInstanceOf(GSDError);
+    const err = thrown as InstanceType<typeof GSDError>;
+    expect(err.classification).toBe(ErrorClassification.Validation);
+    expect(err.message).toContain('--help');
+
+    // Assert no files were written
+    const postStateStat = await import('node:fs').then((m) => m.statSync(statePath));
+    expect(postStateStat.mtimeMs).toBe(preStateStat.mtimeMs);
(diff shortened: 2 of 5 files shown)


trek-e:
@coderabbitai full review

PR #3276, opened 2026-05-09
fix(3266): preserve wave 0 and bucket plans by depends_on DAG in phase-plan-index

### LINKED ISSUE #3266: SDK phase-plan-index: wave grouping ignores depends_on edges; collapses wave: 0 into wave 1
## Summary

`gsd-sdk query phase-plan-index <phase>` produces a wave-grouping that violates two invariants:

1. **Wave 0 collapses into Wave 1.** A plan with frontmatter `wave: 0` is reported with `"wave": 1` and bucketed into the `"1"` array of `waves`.
2. **`depends_on` edges are not honored when building wave buckets.** Two plans where `B.depends_on: [A]` are placed in the *same* wave bucket if their frontmatter `wave:` happens to match (or both get collapsed into wave 1 from wave 0/1).

The downstream effect is that `/gsd-execute-phase` will attempt to spawn the dependent plan in parallel with the dependency, which only happens to be safe if there is also a `files_modified` overlap (the workflow's `intra-wave files_modified overlap check` is a coincidental safety net, not the right one — file overlap is sufficient but not necessary for "must not parallelize").

## Repro

In `lead-db` Phase 21 (`adopt-enrichasin-py-for-sp-api-from-amz-os-rename-enrichasin`):

| plan | frontmatter `wave:` | frontmatter `depends_on:` | SDK-reported `wave` | SDK bucket |
|------|---------------------|---------------------------|---------------------|------------|
| `21-01` | `0` | `[]` | `1` | `waves["1"]` |
| `21-02` | `1` | `[21-01]` | `1` | `waves["1"]` |
| `21-02b` | `1` | `[21-01]` | `1` | `waves["1"]` |
| `21-03` | `2` | (chain) | `2` | `waves["2"]` |
| `21-04` | `3` | (chain) | `3` | `waves["3"]` |

```bash
$ gsd-sdk query phase-plan-index 21 | jq '.waves'
{
  "1": ["21-01", "21-02", "21-02b"],
  "2": ["21-03"],
  ...
}
```

But `21-02` and `21-02b` both have `depends_on: [21-01]`, **and** `21-02b` reads a file (`21-V-SPIKE-RESULT.md`) that `21-01` *produces*. Putting all three in the same wave is wrong on two grounds: the explicit dependency edge, and the implicit data-flow edge (which the file-overlap heuristic cannot detect — `21-V-SPIKE-RESULT.md` is in 21-01's outputs and 21-02b's *reads*, not in 21-02b's `files_modified`).

## Expected

```jsonc
{
  "0": ["21-01"],          // honor frontmatter wave: 0
  "1": ["21-02", "21-02b"], // dependents of 21-01, no inter-edges
  "2": ["21-03"],
  ...
}
```

Or, if collapsing wave 0 is intentional, then dependency-aware grouping should still split it:

```jsonc
{
  "1": ["21-01"],
  "2": ["21-02", "21-02b"],
  "3": ["21-03"],
  ...
}
```

The minimum fix is **dependency-aware**: build the wave buckets via topological-level grouping over `depends_on`, not just `wave:` frontmatter. The frontmatter `wave:` should be advisory; the dependency graph is the source of truth.

## Why this matters

`/gsd-execute-phase` relies on `phase-plan-index.waves` to decide which plans run in parallel. The current behavior:

- Spawns the dependency and its dependents simultaneously in worktree-isolated executors.
- The dependent agent reads files that don't exist yet (because the dependency hasn't committed them), or reads stale versions.
- The orchestrator's `intra-wave files_modified overlap check` does *not* catch this — it only flags shared *write* targets, not write→read data flow.

I caught this manually during a Phase 21 run by reading frontmatter and renumbering the waves locally. An orchestrator following the SDK output blindly would have parallelized 21-01/21-02/21-02b and corrupted 21-02b.

## Suggested fix

In whatever module builds `phase-plan-index.waves`:

1. Parse `depends_on` from each plan's frontmatter.
2. Build a DAG; assign topological levels (source nodes = level 0).
3. Use the level as the wave number.
4. Treat frontmatter `wave:` as a *check* (warn if it disagrees with the computed level) rather than the source of truth.

Optional follow-up: surface a `<plan>.reads_from: [...]` frontmatter field so cross-plan data-flow edges (like `21-V-SPIKE-RESULT.md`) can be made explicit and pulled into the same DAG.

## Environment

- gsd-sdk v1.41.0 (`@gsd-build/sdk`)
- Caught while running `/gsd-execute-phase 21` against `lead-db` repo
- Workflow: `~/.claude/get-shit-done/workflows/execute-phase.md`

trek-e:
## Fix PR

---

## Linked Issue

Fixes #3266

---

## What was broken

`phase-plan-index` had two cooperating defects:

1. **Wave 0 collapse** — `parseInt(String(fm.wave), 10) || 1` coerced the parsed value `0` to `1` (JS falsy default), making it impossible to declare `wave: 0` plans.
2. **`depends_on` ignored** — Wave bucketing read only the `wave:` frontmatter field. `PlanInfo` had no `depends_on` field. The `phase-runner.ts` downstream consumed `planIndex.waves` directly via `Promise.allSettled`, so dependents were parallelized with their dependencies.

## What this fix does

Replaces the falsy-default `|| 1` with a `Number.isNaN` guard so `wave: 0` is preserved. Replaces pure-`wave:`-bucketing with **topological-level grouping** over `depends_on` using Kahn's algorithm: source nodes (no in-phase deps) go in the lowest level; each plan's level = max(deps' levels) + 1. The lowest bucket key is `"0"` when any plan declares `wave: 0`, otherwise `"1"` (preserving existing 1-indexed behavior). Declared `wave:` that disagrees with computed level emits a non-fatal `warnings[]` array on the result. Cycle detection throws `GSDError` (Execution classification) naming the cycle nodes.

## Root cause

The `|| 1` falsy default predates JS `Number.isNaN`. The `depends_on` field from plan frontmatter was parsed by `verify.ts` but never read by the wave builder. `PlanInfo` had no `depends_on` field.

### How I verified the fix

Added regression tests in both the vitest suite (`sdk/src/query/phase.test.ts`) and the CJS node:test suite (`tests/phase.test.cjs`). Confirmed red→green transition: the new tests failed against the old code and passed after the fix.

Ran `npm test` — 8449 pass, 8 pre-existing failures (all pre-date this PR, confirmed by running against `origin/main` unstashed).

Also ran `npm run lint:tests` — 0 violations.

### Regression test added?

- [x] Yes — added tests for:
  - wave 0 round-trip: plan with `wave: 0` lands in `waves["0"]` with `PlanInfo.wave === 0`
  - DAG topological grouping: `B.depends_on: [A]` → B lands in a later bucket
  - declared-vs-computed mismatch surfaces a `warnings[]` array
  - cycle detection throws `GSDError` naming the cycle nodes
  - `PlanInfo.depends_on` is populated from frontmatter

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] N/A (not runtime-specific — this is a core plan-index query function)

---

## Checklist

- [x] Issue linked above with `Fixes #3266` — **PR will be auto-closed if missing**
- [x] `.changeset/` fragment added if this is a user-facing fix — will add after PR number is known

## Breaking changes

None — `PlanInfo` gains a new required field `depends_on: string[]` (previously absent from the type but the runtime now populates it). `PhasePlanIndex` gains optional `warnings?: string[]`. Both are additive.

Plans that declared `wave: 2` with no `depends_on` previously landed in `waves["2"]`; they now land in `waves["1"]` with a warning, because topological analysis shows no dependency that would place them later. This is the corrected behavior per the issue.

## Summary by CodeRabbit

* **New Features**
  * Plans are bucketed by declared dependencies so dependent plans run after their prerequisites.
  * Each plan now exposes its declared dependencies for visibility.

* **Bug Fixes**
  * Fixed wave-collapsing that could merge wave 0 into wave 1.

* **Improvements**
  * Emit warnings when a plan’s declared wave disagrees with its dependency-derived wave; cycles are reported as errors.

### DIFF
--- a/.changeset/noble-otters-hop.md
+++ b/.changeset/noble-otters-hop.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3276
+---
+phase-plan-index no longer collapses wave 0 to wave 1, and now buckets plans using their depends_on DAG so dependents run after their dependencies rather than in the same parallel wave
--- a/get-shit-done/bin/lib/phase.cjs
+++ b/get-shit-done/bin/lib/phase.cjs
@@ -384,6 +384,9 @@ function cmdPhasePlanIndex(cwd, phase, raw) {
   // Build a map from plan ID → raw plan for fast lookup.
   // Deps that reference plans outside this phase are treated as external and ignored.
   const planMap = new Map(rawPlans.map(p => [p.id, p]));
+  // Secondary index: canonical prefix → full plan ID, so depends_on: ['03-01'] resolves
+  // to '03-01-auth-hardening-PLAN.md'-derived ID '03-01-auth-hardening' (k015).
+  const canonicalToId = new Map(rawPlans.map(p => [extractCanonicalPlanId(p.id), p.id]));
 
   // Kahn's algorithm — compute in-degree and adjacency for in-phase deps only.
   const level = new Map();
@@ -394,9 +397,11 @@ function cmdPhasePlanIndex(cwd, phase, raw) {
     if (!inDeg.has(p.id)) inDeg.set(p.id, 0);
     if (!adj.has(p.id)) adj.set(p.id, []);
     for (const dep of p.dependsOn) {
-      if (!planMap.has(dep)) continue; // external dep — ignore
-      if (!adj.has(dep)) adj.set(dep, []);
-      adj.get(dep).push(p.id);
+      // Accept both full-stem ('03-01-auth-hardening') and canonical-prefix ('03-01') forms.
+      const resolvedDep = planMap.has(dep) ? dep : canonicalToId.get(dep);
+      if (!resolvedDep) continue; // external dep — ignore
+      if (!adj.has(resolvedDep)) adj.set(resolvedDep, []);
+      adj.get(resolvedDep).push(p.id);
       inDeg.set(p.id, (inDeg.get(p.id) ?? 0) + 1);
     }
   }
--- a/get-shit-done/bin/lib/phase.cjs
+++ b/get-shit-done/bin/lib/phase.cjs
@@ -322,10 +322,9 @@ function cmdPhasePlanIndex(cwd, phase, raw) {
     })
   );
 
-  const plans = [];
-  const waves = {};
-  const incomplete = [];
-  let hasCheckpoints = false;
+  // ── Pass 1: parse each plan file ─────────────────────────────────────────
+
+  const rawPlans = [];
 
   for (const planFile of planFiles) {
     const planId = planFile.replace('-PLAN.md', '').replace('PLAN.md', '');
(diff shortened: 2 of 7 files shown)


trek-e:
@coderabbitai full review

trek-e:
@coderabbitai full review

PR #3289, opened 2026-05-09
fix(install): allow codex hooks.state.<key> as regular table (#3285)

### LINKED ISSUE #3285: Codex install fails when config.toml contains hooks.state entries
﻿## Summary

Installing or updating `get-shit-done-cc` for Codex fails during post-write schema validation when an existing Codex `config.toml` contains `hooks.state` entries that store trusted hook state.

This appears to be a validator/compatibility bug rather than an invalid Codex config.

## Environment

- GSD package: `get-shit-done-cc@1.41.1`
- Codex CLI: `0.130.0`
- Node.js: `v22.22.2`
- OS: Windows 11
- Install target: Codex global install (`~/.codex`)

## Repro

1. Use Codex long enough that `~/.codex/config.toml` contains a `hooks.state` section for previously trusted hooks.
2. Run:

```powershell
npx -y --package=get-shit-done-cc@latest -- get-shit-done-cc --codex --global
```

3. Let the installer write the Codex config.

## Example pre-existing Codex config shape

```toml
[hooks.state]

[hooks.state.'<project>/.codex/hooks.json:pre_tool_use:0:0']
enabled = true
trusted_hash = "sha256:..."
```

## Expected

The installer should preserve or tolerate valid Codex `hooks.state` data and complete successfully.

## Actual

The install aborts with:

```text
post-write Codex schema validation failed: bare [hooks.state.<project>/.codex/hooks.json:pre_tool_use:0:0] table is invalid in current Codex schema (expected [[hooks.state.<project>/.codex/hooks.json:pre_tool_use:0:0]] array-of-tables). Restored pre-install backup.
```

## Workaround

If I temporarily remove the entire `hooks.state` block from `~/.codex/config.toml`, the same install command succeeds.

However, that also removes Codex's remembered hook trust state, so after the install Codex prompts to review/trust the hook again.

## Why this looks like a GSD-side validator bug

The failing entries are not event hook definitions like `[[hooks.SessionStart]]`. They are Codex-managed hook trust state under `hooks.state`.

It looks like the post-write validator is treating every `hooks.*` table as an event hook that must be expressed as an array-of-tables, which breaks on `hooks.state`.

## Optional follow-up

After the install succeeds, GSD also leaves a deprecated `codex_hooks = true` feature flag in `[features]`. That is non-blocking, but the schema-validation failure above is the main installation blocker.

trek-e:
## Linked Issue

Fixes #3285

## What was broken

`get-shit-done-cc --codex --global` failed with a post-write schema validation error when the existing `~/.codex/config.toml` contained `hooks.state` entries — the persistent hook-trust state that Codex CLI 0.130.0+ writes to track which hooks have been approved.

## What this fix does

Adds a carve-out in `validateCodexConfigSchema` for the `hooks.state` and `hooks.state.*` namespaces, allowing them to appear as regular (single-bracket) tables. Event-handler paths like `hooks.SessionStart` still require the array-of-tables (`[[hooks.SessionStart]]`) shape.

## Root cause

The validator was written when the only `hooks.*` sub-keys were Codex event names (SessionStart, etc.), which do require AoT shape. Codex CLI 0.130.0 added a `hooks.state` namespace for per-hook trust persistence that uses regular-table shape. The validator's AoT-required rule matched every `hooks.*` path without exception, so any existing trust-state entry caused the post-write validation to fail and roll back the install.

Two sites needed fixing:
1. Section-header loop: `!section.array && section.path.startsWith('hooks.')` — now excludes `hooks.state` and `hooks.state.*`.
2. Parsed-object loop over `Object.entries(parsed.hooks)` — now skips the `state` key (which is an object, not an array of event-handler tables).

### How I verified the fix

Added `tests/bug-3285-codex-hooks-state-allowed.test.cjs` with 8 tests:
- Unit tests driving `validateCodexConfigSchema` directly with bare `[hooks.state]`, deep `[hooks.state.'<key>']`, mixed state+SessionStart configs, and multiple sub-keys.
- Regression guard confirming `[hooks.SessionStart]` (bare, not AoT) still fails.
- Integration tests running the full install pipeline against a pre-populated config with trust entries, asserting no throw and structural preservation of `hooks.state` in the output.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `get-shit-done-cc --codex` to accept valid hook state configuration entries that were previously rejected during schema validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/get-shit-done/pull/3289)

### DIFF
--- a/.changeset/mellow-lynx-forage.md
+++ b/.changeset/mellow-lynx-forage.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3289
+---
+**`get-shit-done-cc --codex` no longer rejects valid Codex `hooks.state` trust-persistence entries** — the schema validator was over-classifying every `hooks.*` table as an event-handler array-of-tables, breaking installs against Codex CLI 0.130.0+ where `hooks.state.<project>/...` stores per-hook trust state. Regular-table shape is now accepted for `hooks.state.*` while `hooks.<EVENT>` still requires AoT.
--- a/bin/install.js
+++ b/bin/install.js
@@ -3977,7 +3977,13 @@ function validateCodexConfigSchema(content) {
       };
     }
 
-    if (!section.array && section.path.startsWith('hooks.')) {
+    // hooks.state.* is Codex's persistent hook-trust namespace (added in
+    // Codex CLI 0.130.0). It uses regular-table shape, NOT array-of-tables.
+    // Any section whose path is exactly `hooks.state` or starts with
+    // `hooks.state.` must be allowed as a bare table. All other hooks.*
+    // paths (event handlers like hooks.SessionStart) still require AoT.
+    if (!section.array && section.path.startsWith('hooks.') &&
+        section.path !== 'hooks.state' && !section.path.startsWith('hooks.state.')) {
       return {
         ok: false,
         reason: `bare [${section.path}] table is invalid in current Codex schema (expected [[${section.path}]] array-of-tables)`,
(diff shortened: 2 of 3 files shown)


trek-e:
@coderabbitai review

Issue #3287, opened 2026-05-09
gsd-sdk: phase directory naming inconsistent — `project_code` prefix applied by `phase.add`/`phase.insert` but bypassed by `/gsd-discuss-phase` and `/gsd-plan-phase` first-touch creation paths


0xdhx:
## TL;DR

When a project has `project_code` set in `.planning/config.json`, GSD's phase-creation paths disagree on directory naming. `gsd-sdk query phase.add` and `gsd-sdk query phase.insert` (called from `/gsd-phase` and `/gsd-phase --insert`) honor the prefix and create `<CODE>-NN-slug/` directories. But the first-touch creation path used by `/gsd-discuss-phase` (and likely `/gsd-plan-phase`) — which materializes `.planning/phases/NN-slug/` when the discuss/plan workflow writes its first artifact — creates the directory without the prefix. The result: a project with `project_code: "XR"` accumulates a two-headed naming convention (`01-foundation/` from discuss-phase first-touch alongside `XR-02.1-spike/` from `phase.insert`), defeating the cross-project-namespace benefit `project_code` was added to provide.

**Fix:** route all phase-directory creation through a single shared helper (e.g., `getPhaseDirName(phaseNum, slug, config)` already implicit in `cmdPhaseAdd`/`cmdPhaseInsert`), or have `init.phase-op` apply the prefix when materializing the directory if not present.

### 1. `project_code` prefix applied inconsistently across phase-creation paths

**What happens:**

- Project has `"project_code": "XR"` in `.planning/config.json` (introduced upstream by [commit 1f349657](https://github.com/gsd-build/get-shit-done/commit/1f349657), 2026-03-26, closes #1019, intent: *"disambiguates phases across multiple GSD projects in the same session"*).
- `gsd-sdk query phase.add "<desc>"` → creates `.planning/phases/XR-NN-<slug>/` (prefix applied).
- `gsd-sdk query phase.insert <after> "<desc>"` → creates `.planning/phases/XR-NN.M-<slug>/` (prefix applied).
- `/gsd-discuss-phase <N>` (or whichever workflow first writes a CONTEXT.md / PLAN.md into a phase dir) → creates `.planning/phases/NN-<slug>/` (prefix **NOT** applied).

The `.planning/phases/` tree ends up containing both shapes:

```
.planning/phases/
├── 01-phase-branch-workflow-foundation/      ← created via /gsd-discuss-phase first-touch (NO prefix)
[...]

PR #3539, opened 2026-05-15
fix(state): padding-tolerant phase-ID matching across affected verbs (#3537)

bshiggins:
## Linked Issue

Fixes #3537

> The linked issue carries the `confirmed-bug` label (applied 2026-05-14).

## What was broken

Phase-state verbs — `roadmap get-phase`, `roadmap analyze`, `roadmap annotate-dependencies`, `phase complete`, `phase next-decimal`, `phase insert`, and the shared `getRoadmapPhaseInternal` helper in `core.cjs` — silently no-op'd on projects whose `ROADMAP.md` uses un-padded phase numbers (`### Phase 2.7:`) when callers passed the padded form (`02.7`). Skills that resolve a phase and pass its padded `phase_number` to state verbs would receive success-shaped JSON while the regex never matched.

## What this fix does

Wires `phaseMarkdownRegexSource()` — the padding-tolerant regex-source helper added in 1.42.1 (#3287) — across the remaining affected sites. Promotes the helper from `roadmap.cjs` to `core.cjs` so all three lib files share one definition without circular imports.

## Root cause

1.42.1 added `phaseMarkdownRegexSource()` in `roadmap.cjs` and wired it into `cmdRoadmapUpdatePlanProgress` only. The other sites still constructed phase-number regex sources with raw `escapeRegex(phaseNum)` (or `0*${escapeRegex(...)}` literals that tolerate extra padding but not missing). A padded caller `02.7` produced `Phase\s+02\.7:` and never matched un-padded ROADMAP prose `### Phase 2.7:`.

Wired sites in this PR:

- `core.cjs:1079` — `getRoadmapPhaseInternal` (replaces the `isNumeric`-branched literal that left decimal phases broken)
- `roadmap.cjs:151` — `cmdRoadmapGetPhase` (`escapedPhase` → `searchPhaseInContent`)
- `roadmap.cjs:252` — `cmdRoadmapAnalyze` per-phase checkbox detection
- `roadmap.cjs:515` — `cmdRoadmapAnnotateDependencies` phase-header lookup
- `phase.cjs:174` — `cmdPhaseNextDecimal` decimal scan (drops the now-redundant literal `0*` since the helper prepends it)
- `phase.cjs:724` — `cmdPhaseInsert` decimal scan (same drop)
- `phase.cjs:1034`, `:1040`, `:1097` — `cmdPhaseComplete` checkbox + table-row + REQUIREMENTS section

`cmdRoadmapUpdatePlanProgress` already used the helper at `roadmap.cjs:356` and is unchanged.

### How I verified the fix

New regression test `tests/bug-3537-padded-id-against-unpadded-roadmap.test.cjs`:

- Helper-export sanity test (confirms `phaseMarkdownRegexSource` is reachable from `core.cjs`).
- Parity-style behavioral tests per affected verb via `runGsdTools` from `tests/helpers.cjs`, asserting padded and un-padded callers produce the same result against un-padded ROADMAP fixtures.
- One mismatch-fixture test for `cmdRoadmapAnalyze` (un-padded prose + padded checklist) — a different surface from caller-padding but covered by the same regex fix at `roadmap.cjs:252`.

**TDD verified empirically:** stashed the lib changes, ran the new test against unpatched main — confirmed **RED** on 5 of 8 tests (helper export, `get-phase`, `annotate-dependencies`, `analyze`, `phase complete`). Restored the fix, re-ran — all 8 GREEN. The 3 always-GREEN tests are regression guards for sites where the literal `0*` already tolerated padding (`cmdRoadmapUpdatePlanProgress`, `cmdPhaseNextDecimal`, `cmdPhaseInsert` with integer bases) but the helper-relocation could regress.

### Regression test added?

### Platforms tested

- [x] macOS (local `npm test`: **9213 passing / 0 failing**)
- [ ] Windows (not exercised locally — covered by CI matrix)
- [x] Linux (covered by CI matrix: Ubuntu × Node 22, 24)
- [ ] N/A

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] N/A — purely SDK-internal regex-source change

## Checklist

- [x] Issue linked above with `Fixes #3537`
- [x] Regression test added
- [x] All existing tests pass (`npm test` — 9213/9213)
- [x] `.changeset/` fragment added (`fix-3537-padding-tolerant-phase-id-matching.md`, type Fixed)

## Breaking changes

None. `phaseMarkdownRegexSource` was not previously exported from `roadmap.cjs`'s `module.exports`, so promoting the definition to `core.cjs` and importing it back doesn't change any external API surface.

---

**Maintainer flexibility:** if you'd prefer to keep `phaseMarkdownRegexSource` defined in `roadmap.cjs` rather than promote it to `core.cjs`, the alternative is to keep the def in `roadmap.cjs`, export it, and import it into `core.cjs` + `phase.cjs`. Both shapes are equivalent for correctness; happy to refactor on request.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed phase identification to consistently match padded phase identifiers (e.g., `02.7`) against unpadded phase references in roadmap documentation, ensuring all roadmap and phase-related commands work correctly regardless of zero-padding format.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/get-shit-done/pull/3539)

### DIFF
--- a/.changeset/fix-3537-padding-tolerant-phase-id-matching.md
+++ b/.changeset/fix-3537-padding-tolerant-phase-id-matching.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3539
+---
+**Padded phase IDs no longer silently no-op on un-padded ROADMAP prose** — `phaseMarkdownRegexSource()` (added in 1.42.1) was wired into only 1 of 8 affected sites; the rest used raw `escapeRegex(phaseNum)` and silently failed when skills resolved a padded `phase_number` like `02.7` and passed it to state verbs against un-padded prose like `### Phase 2.7:`. The helper is now wired across `cmdRoadmapGetPhase`, `cmdRoadmapAnalyze`, `cmdRoadmapAnnotateDependencies`, `cmdPhaseComplete`, `cmdPhaseNextDecimal`, `cmdPhaseInsert`, and `getRoadmapPhaseInternal`; promoted from `roadmap.cjs` to `core.cjs` so the three lib files share one definition without a circular import. Closes #3537.
--- a/get-shit-done/bin/lib/core.cjs
+++ b/get-shit-done/bin/lib/core.cjs
@@ -726,6 +726,30 @@ function normalizePhaseName(phase) {
   return str;
 }
 
+/**
+ * Build a regex-source fragment that matches a phase number in ROADMAP.md /
+ * STATE.md prose regardless of zero-padding. ROADMAP text is human-authored and
+ * usually un-padded ("Phase 2.6"); phase dirs and SDK-normalized IDs are padded
+ * ("CK-02.6", "02.6"). Callers pass either form — this tolerates both:
+ *   "2.6" | "02.6" | "CK-02.6"  -> "0*2\.6"
+ *   "3"   | "03"                 -> "0*3"
+ *   "12A"                        -> "0*12A"   (case handled by /i at call sites)
+ *   "PROJ-42"                    -> "0*42"    (project-code prefix stripped, like normalizePhaseName)
+ *   non-numeric custom IDs       -> escaped passthrough
+ *
+ * Lives in core.cjs so phase.cjs / roadmap.cjs / core.cjs all share one helper.
+ */
+function phaseMarkdownRegexSource(phaseNum) {
+  const stripped = String(phaseNum).replace(/^[A-Z]{1,6}-(?=\d)/i, '');
+  const match = stripped.match(/^0*(\d+)([A-Z])?((?:\.\d+)*)$/i);
+  if (!match) return escapeRegex(phaseNum);
+
+  const integer = match[1].replace(/^0+/, '') || '0';
+  const letter = match[2] ? escapeRegex(match[2]) : '';
+  const decimal = match[3] ? escapeRegex(match[3]) : '';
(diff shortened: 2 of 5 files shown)


bshiggins:
Closing as duplicate of #3538 — your fix landed ~8 minutes before this PR opened, and we took the same approach (helper promoted to `core.cjs`, wired across the 8 sites, parity-style regression test). Nothing here that's not already in main.

Two small wrinkles that may or may not warrant a follow-up:

1. **Helper-export regression guard.** This PR included a sanity test confirming `phaseMarkdownRegexSource` is exported from `core.cjs` (catches an accidental revert of the relocation). The merged test verifies wiring at each call site but doesn't directly assert the export location.

2. **`cmdRoadmapAnalyze` mismatch fixture.** `roadmap.cjs:252` is a slightly different surface than the others — the regex is built from prose-form `phaseNum` and matches against the summary checklist, so the bug manifests when one ROADMAP has un-padded prose + padded checklist (rather than the caller-padding direction the other tests exercise). Happy to open a small follow-up if you'd like that case explicitly guarded.

Otherwise the bug is fixed and the regression test catches the pattern. Thanks for the quick triage and turnaround.

PR #3611, opened 2026-05-16
fix(3587)(security): argv-based subprocess for check.ship-ready

trek-e:
## Linked Issue

Fixes #3587

The linked issue has the \`confirmed-bug\` label.

---

## What was broken

\`gsd-sdk query check.ship-ready <phase>\` was vulnerable to shell-command injection through git branch names. Git accepts refnames containing shell metacharacters (\`;\`, \`\$\`, backticks, etc.), and \`sdk/src/query/check-ship-ready.ts:50\` interpolated the current branch name into a shell-string passed to \`execSync\`:

\`\`\`ts
runSyncSafe(\`git config --get branch.\${current_branch}.merge\`, cwd)
// → execSync('git config --get branch.foo;touch\${IFS}INJ;bar.merge')
// → /bin/sh parses three commands; middle one creates INJ in the project dir
\`\`\`

A user who checks out a maliciously-named branch (or pulls one named that way from a malicious remote) and then runs \`gsd-sdk query check.ship-ready\` triggers arbitrary code execution.

## What this fix does

Replaces every subprocess call in the module with argv-based \`execFileSync\`. The shell is never invoked, so branch names — even ones containing metacharacters — are passed as a single argv element and treated as opaque data.

- \`runSyncSafe(cmd: string, cwd)\` → \`runArgvSafe(file, args: readonly string[], cwd)\` — pure argv via \`execFileSync\`.
- 7 subprocess sites converted: \`git status --porcelain\`, \`git rev-parse --abbrev-ref HEAD\`, \`git config --get branch.<name>.merge\` (the interpolation site), \`git rev-parse --verify main\`, \`git remote\`, \`gh --version\`, \`which gh\`.

## Root cause

\`runSyncSafe(cmd: string, cwd)\` used \`execSync\`, which spawns \`/bin/sh -c <cmd>\`. The shell parses the argument as shell syntax, so any \`;\`, \`\$()\`, or backtick in the interpolated branch name breaks out of git's argv into shell command syntax. Argv-based \`execFileSync\` doesn't invoke a shell — the kernel exec call receives the args verbatim — closing the entire class of injection sites in one helper change.

### How I verified the fix

- **Manually reproduced the exploit** on git 2.53.0: refname \`foo;touch\${IFS}INJ;bar\` is accepted by \`git check-ref-format\` and by \`git checkout -b\`. Running the pre-fix \`execSync\` against that branch name creates an \`INJ\` file in the project dir.
- **TDD regression test** (\`sdk/src/query/check-ship-ready.test.ts\`): creates a real temp git repo, checks out the proven exploit branch, runs \`checkShipReady\`, and asserts the sentinel file does NOT exist. Verified the test FAILED on the unfixed code (\`expected true to be false\` — INJ file got created) before applying the fix. After the fix, all 10 tests pass.
- **Additional coverage**: positive round-trip test (branch with \`\$\`, \`(\`, \`)\` metacharacters appears verbatim in \`current_branch\`), and a gh-probe contract guard against future regressions.

### Regression test added?

- [x] Yes — three tests: exploit-blocked (negative proof), metacharacter round-trip (positive proof), gh-probe contract guard. The exploit-blocked test was verified to fail without the fix.
- [ ] No — explain why:

### Platforms tested

- [x] macOS — full suite via gsd-test-local + SDK vitest unit suite
- [ ] Windows (including backslash path handling) — argv execution is platform-agnostic; \`execFileSync\` quoting on Windows is handled by Node, not us
- [x] Linux — full suite via remote Docker (gsd-test-both)

### Runtimes tested

- [x] Claude Code — handler is runtime-agnostic; the change is in the SDK query layer
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Codex
- [ ] Other: ___
- [ ] N/A — handler runs in the SDK process under any runtime

---

## Checklist

- [x] Issue linked above with \`Fixes #NNN\` — **PR will be auto-closed if missing**
- [x] Linked issue has the \`confirmed-bug\` label
- [x] Fix is scoped to the reported bug — no unrelated changes (only check-ship-ready.ts + its test file + changeset)
- [x] Regression test added that proves the original exploit no longer fires
- [x] All existing tests pass — SDK unit suite 1,869/1,869; full root suite via \`gsd-test-both\` 10,676/10,676 Mac + 10,676/10,676 Linux Docker, zero diff
- [x] \`.changeset/\` fragment added — \`proud-sloths-rally.md\`, type \`Security\`, PR 3587

## Breaking changes

None. Behavior contract is identical: same inputs produce the same outputs. The change is purely in the subprocess invocation mechanism — argv instead of shell-string — and only matters when an attacker controls the branch name. Normal branch names produce identical results.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a security issue where maliciously named git branches could lead to shell command injection; branch names are now handled safely as opaque data.

* **Tests**
  * Added regression tests to ensure branch names with shell metacharacters cannot trigger command execution and to guard against future regressions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/get-shit-done/pull/3611?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/proud-sloths-rally.md
+++ b/.changeset/proud-sloths-rally.md
@@ -1,0 +1,5 @@
+---
+type: Security
+pr: 3587
+---
+**Fixed shell-injection in `check.ship-ready`** — a maliciously-named git branch (e.g. `foo;touch${IFS}INJ;bar`) could execute arbitrary shell commands when `gsd-sdk query check.ship-ready <phase>` ran from the repository. `sdk/src/query/check-ship-ready.ts` interpolated the current branch name into a shell-string `git config --get branch.${current_branch}.merge` and ran it via `execSync`. Every subprocess call in the module now uses argv-based `execFileSync` — the shell is never invoked, branch names are passed as opaque data, and no interpolation site exists.
--- a/sdk/src/query/check-ship-ready.test.ts
+++ b/sdk/src/query/check-ship-ready.test.ts
@@ -3,10 +3,30 @@
  */
 
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
-import { mkdir, writeFile, rm } from 'node:fs/promises';
+import { mkdir, writeFile, rm, stat as fsStat } from 'node:fs/promises';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
+import { execFileSync } from 'node:child_process';
 import { checkShipReady } from './check-ship-ready.js';
+
+/**
+ * Initialize a fresh git repo in `dir` with one empty initial commit on
+ * branch `main`. Uses argv-based execFileSync so the test harness itself
+ * never shell-interpolates anything. Returns silently if git is unavailable
+ * on the host; callers check for that and skip.
+ */
+function initGitRepoOrSkip(dir: string): boolean {
+  try {
+    execFileSync('git', ['init', '--initial-branch=main', '--quiet'], { cwd: dir, stdio: ['pipe', 'pipe', 'pipe'] });
+    execFileSync('git', ['config', 'user.email', 'test@example.com'], { cwd: dir, stdio: ['pipe', 'pipe', 'pipe'] });
+    execFileSync('git', ['config', 'user.name', 'Test'], { cwd: dir, stdio: ['pipe', 'pipe', 'pipe'] });
+    execFileSync('git', ['config', 'commit.gpgsign', 'false'], { cwd: dir, stdio: ['pipe', 'pipe', 'pipe'] });
+    execFileSync('git', ['commit', '--allow-empty', '-m', 'init', '--quiet'], { cwd: dir, stdio: ['pipe', 'pipe', 'pipe'] });
+    return true;
+  } catch {
+    return false;
+  }
+}
 
 describe('checkShipReady', () => {
   let projectDir: string;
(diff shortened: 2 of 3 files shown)


trek-e:
## Codex review — all in-scope findings addressed in ac34ab2c

Codex returned 2 MED + 5 LOW. Three LOW are out-of-scope (pre-existing test code or SDK convention disagreements with root-helper conventions); the rest are fixed.

| Severity | Site | Finding | Resolution |
|----------|------|---------|------------|
| MED | gh test sensitivity | Test only asserted the boolean shape of \`gh_available\`; a future change that re-introduced shell-string \`execSync\` for the gh path would pass undetected. | ac34ab2c: added an \`architectural-invariant\` structural test that reads \`check-ship-ready.ts\` and asserts no \`execSync(\` calls, no \`spawnSync … shell:true\`, \`execFileSync\` is the only imported child_process primitive, and every options object explicitly pins \`shell: false\`. This is the canonical CONTRIBUTING.md pattern for invariants that behavioral tests can't observe (the defect is the *presence* of the shell parser, not its output). |
| MED | execFileSync options | Default is already \`shell: false\` but the option wasn't pinned explicitly — leaving room for a future options-spread to flip it, and not protecting Windows where a \`git.cmd\` shim could route through cmd.exe. | ac34ab2c: pinned \`shell: false\` explicitly in the runArgvSafe options object with a comment explaining why. |
| LOW | exploit-blocked test skip | Silently \`return\`ed when git rejected the payload refname on a stricter platform — coverage loss looked like a pass. | ac34ab2c: replaced with vitest's \`ctx.skip()\` so reduced coverage shows up in the CI skip count. |
| LOW (out of scope) | \`try/finally\` at line 79 | Pre-existing test code from before this PR. | One-concern-per-PR rule — not changing in a fix-only PR for #3587. |
| LOW (out of scope) | \`createTempGitProject\` helper | That helper lives in \`tests/helpers.cjs\` (root, node:test world). | SDK tests use vitest with their own ad-hoc tmpdir pattern; matching SDK convention. A cross-runtime helper migration is a separate task. |
| LOW (out of scope) | shared \`cleanup\` helper | Same — root vs SDK divergence. | Not changing in this PR. |

Validation: 11/11 new tests pass (added the architectural-invariant test); SDK unit suite 1,870/1,870; full root suite via \`gsd-test-both\` 10,676/10,676 on Mac AND 10,676/10,676 on Linux Docker, zero cross-platform diff.

> *This was generated by AI during PR review remediation.*

PR #3634, opened 2026-05-16
test(3595): filesystem fault-injection for platformWriteSync atomic-write seam

trek-e:
## Linked Issue

Closes #3595

The linked issue has the `approved-enhancement` label.

---

## What this enhancement improves

Adds filesystem fault-injection coverage that `CONTRIBUTING.md` §"QA Matrix Requirements / Filesystem writes and installers" and `TEST-EXAMPLES.md` §"Filesystem Fault Injection" prescribe. Tests target the **canonical atomic-write seam** (`platformWriteSync` in `get-shit-done/bin/lib/shell-command-projection.cjs`) — the function every CJS config/state/generated-artifact writer routes through. Coverage at that seam transitively covers every write path that uses it.

## Before / After

**Before:**

- `platformWriteSync`'s fault-recovery logic (rename failure → fallback to direct write, with tmp cleanup) was implicit — no test exercised any of the error paths.
- No coverage for behavior on read-only/non-existent parent dirs, target-is-directory collisions, paths with special characters, or symlink-pointed targets.
- The "fall back to direct write" branch's error-handling semantics were undocumented (it silently swallows the original tmp-write error and surfaces only the fallback's error if both fail).

**After:**

13 new tests in `tests/feat-3595-fs-fault-injection-atomic-write.test.cjs`. Each uses `node:test`'s `mock.method()` against the real `fs` seam, restores mocks via `t.after()` so failures don't leak, and asserts on observable post-conditions (file existence, content, presence/absence of orphan tmp files, propagated `err.code`) — never on prose.

Fault matrix coverage:

| Fault | Test |
|---|---|
| Happy path baseline | `platformWriteSync happy path writes content atomically (baseline for fault tests)` |
| Rename fails (EXDEV cross-device) | `platformWriteSync recovers when renameSync fails` |
| Tmp write fails (ENOSPC) | `platformWriteSync falls back when initial tmp writeFileSync fails` |
| Both writes fail (ENOSPC + EACCES) | `platformWriteSync propagates the FALLBACK error when both tmp and fallback writes fail` |
| mkdirSync fails (EACCES) | `platformWriteSync propagates mkdirSync failure unchanged` |
| Target path is an existing directory | `platformWriteSync against a target path that is an existing directory fails cleanly` |
| Paths with spaces / Unicode / tabs / newlines | `platformWriteSync handles paths with spaces, unicode, and newline characters` |
| 25 sequential writes leave 0 tmp orphans | `platformWriteSync never leaks a tmp file after a successful happy-path write` |
| `platformEnsureDir` idempotent | `platformEnsureDir is idempotent on an existing directory` |
| `platformEnsureDir` propagates EACCES | `platformEnsureDir propagates EACCES when parent dir is unwritable` |
| Symlink replacement (SAFETY) | `platformWriteSync REPLACES a symlink with a regular file rather than following it` |
| Broken symlink replacement | `platformWriteSync against a broken symlink replaces it with the intended file` |
| Concurrent-write collision (EBUSY mid-flight) | `platformWriteSync survives a concurrent collision on the same target path` |

**Symlink safety (security-relevant):**

The atomic-write pattern's `renameSync(tmp, target)` REPLACES a symlink at `target` with the regular file from `tmp` — it does NOT follow the symlink. A test pins this so a planted symlink in `.planning/` pointing at `~/.ssh/authorized_keys` (or similar) is not clobbered. Any future refactor to `fs.writeFileSync(targetPath, ...)` (which DOES follow symlinks) becomes a visible test regression.

**Two known-open gaps deliberately NOT fixed in this PR** (each warrants its own focused issue):

1. `platformWriteSync`'s fallback path swallows the original tmp-write error. When both writes fail, operators only see the fallback's error. Tests pin the current behavior.
2. `platformWriteSync`'s `mkdirSync(dirname, { recursive: true })` error escapes without context-wrapping. Tests pin the current code propagation.

## How it was implemented

- Single new test file (`feat-3595-fs-fault-injection-atomic-write.test.cjs`).
- Each test uses its own scratch `mkdtempSync` directory + `t.after(() => fs.rmSync(...))` cleanup. No shared state across tests.
- Mocks use `mock.method(fs, 'writeFileSync', ...)`, `mock.method(fs, 'renameSync', ...)`, `mock.method(fs, 'mkdirSync', ...)`. All restored via `t.after(() => mock.restore())`.
- The orphan-tmp detector uses a strict regex `/\.tmp\.\d+$/` against `readdirSync` entries so a test that happens to write a real `*.tmp.123` file doesn't trigger a false positive.
- The double-failure test pins the CURRENT error surfaced (the fallback's `EACCES`) so flipping it to a chained `.cause` is a one-line assertion change the day the fix lands.

### How I verified the enhancement works

- `node --test tests/feat-3595-fs-fault-injection-atomic-write.test.cjs` → **13 tests, 13 pass, 0 fail**.
- `node --test tests/feat-3595-*.test.cjs tests/bug-2774-worktree-cleanup-workspace-safety.test.cjs tests/state.test.cjs tests/config.test.cjs` → **192 tests, 192 pass, 0 fail** — pre-existing fs-adjacent suites unaffected.
- `node scripts/lint-no-source-grep.cjs` → **545 test files checked, 0 violations**.

### Platforms tested

- [x] macOS — full local matrix
- [ ] Windows (including backslash path handling) — symlink-related tests are explicitly gated `if (process.platform === 'win32') return;` so the suite is cross-platform-safe
- [ ] Linux

### Runtimes tested

- [x] N/A (filesystem-layer test coverage — runtime-agnostic)

---

## Scope confirmation

- [x] The implementation matches the scope approved in the linked issue — no additions or removals
- [x] If scope changed during implementation, I updated the issue and got re-approval before continuing

The issue called out a broad fault matrix across `bin/install.js`, `worktree-safety.cjs`, `shell-command-projection.cjs`, config/state/artifact writers, and install/uninstall flows. This PR delivers coverage at the SHARED atomic-write seam, which is the highest-leverage layer (every writer above transitively benefits). Per-flow deepening for `bin/install.js` install/uninstall sequences and for `worktree-safety.cjs` lifecycle faults is mechanical follow-up that the pattern landed here unlocks.

---

## Checklist

- [x] Issue linked above with `Closes #3595`
- [x] Linked issue has the `approved-enhancement` label
- [x] Changes are scoped to the approved enhancement — test-only PR, zero production code changes
- [x] All existing tests pass (192/192 across adjacent fs/state/config/worktree suites)
- [x] New or updated tests cover the enhanced behavior
- [x] `.changeset/3595-fs-fault-injection-atomic-write.md` added (`Added` — new test surface)
- [x] Documentation updated if behavior or output changed (per-test inline docstrings document the contract being pinned)

## Breaking changes

None. Test-only PR.

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive fault-injection test suite for filesystem atomic-write behavior, covering multiple failure scenarios, edge cases (paths with special characters and whitespace), symlink replacement and broken-symlink handling, concurrency/collision cases, directory creation and collision scenarios, cleanup invariants, and error-propagation checks. The suite also records current behavior gaps to be addressed in a future fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/get-shit-done/pull/3634?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/3595-fs-fault-injection-atomic-write.md
+++ b/.changeset/3595-fs-fault-injection-atomic-write.md
@@ -1,0 +1,5 @@
+---
+"get-shit-done-cc": patch
+---
+
+**Added: filesystem fault-injection coverage for `platformWriteSync`.** New tests in `tests/feat-3595-fs-fault-injection-atomic-write.test.cjs` use `node:test`'s `mock.method()` to inject `ENOSPC`, `EACCES`, `EXDEV`, `EBUSY`, and `EISDIR` against the shared atomic-write seam in `get-shit-done/bin/lib/shell-command-projection.cjs`. Covers rename-failure fallback, double-failure error propagation, mkdir failure, target-is-directory collision, paths with spaces/Unicode/newlines, symlink-replacement safety (writer does NOT follow symlinks), broken-symlink handling, and concurrent-write collision. Documents two pre-existing behavior gaps (fallback error swallows original cause; mkdir failure escapes unhandled) as pinned current behavior, ready for the future fix to flip the assertion.
--- a/tests/feat-3595-fs-fault-injection-atomic-write.test.cjs
+++ b/tests/feat-3595-fs-fault-injection-atomic-write.test.cjs
@@ -1,0 +1,409 @@
+/**
+ * Filesystem fault-injection coverage for the canonical atomic-write
+ * seam (#3595).
+ *
+ * `platformWriteSync` in `get-shit-done/bin/lib/shell-command-projection.cjs`
+ * is the shared seam every config/state/generated-artifact writer in
+ * the CJS layer routes through. Its contract:
+ *
+ *   1. mkdirSync(dirname, { recursive: true }) — ensure parent exists.
+ *   2. writeFileSync(<tmpPath>, content) — write to a sibling tmpfile
+ *      named `<filePath>.tmp.<pid>`.
+ *   3. renameSync(<tmpPath>, filePath) — atomic publish.
+ *   4. On any error in steps 2-3: unlinkSync(<tmpPath>) (best-effort)
+ *      then writeFileSync(filePath, content) directly as a fallback.
+ *
+ * Per CONTRIBUTING.md §"QA Matrix Requirements / Filesystem writes and
+ * installers" the tests below use `mock.method()` against the real `fs`
+ * seam to drive each fault mode, restore mocks with `t.after()`, and
+ * assert on observable post-conditions (file existence, content,
+ * presence/absence of orphan tmp files, propagated error code) — not
+ * on prose.
+ *
+ * Pre-existing behavior gaps surfaced and PINNED (not fixed in this
+ * PR — #3595 is test coverage, fixes belong in separate issues):
+ *
+ *   - The "fall back to direct write" branch silently swallows the
+ *     ORIGINAL error from the tmp+rename path. If the fallback ALSO
+ *     fails, the operator only sees the fallback's error — not the
+ *     original cause. Tests document this.
+ *   - On EACCES against the parent directory mkdir, the error
+ *     escapes (no try/catch around mkdirSync). Pinned.
+ */
+
+'use strict';
+
+const { test, mock } = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const path = require('node:path');
+const os = require('node:os');
+
+const {
+  platformWriteSync,
+  platformEnsureDir,
+} = require('../get-shit-done/bin/lib/shell-command-projection.cjs');
+
+/**
+ * Create a fresh real-fs scratch dir per test so no two faults share
+ * state. Returns the directory; caller must clean up.
+ */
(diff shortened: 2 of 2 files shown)


trek-e:
Addressed both Codex nitpicks (commit `ade78301`):

1. **Win32 symlink gates** — replaced `if (process.platform === 'win32') return` with `t.skip('symlinks on Win32 need admin'); return` in both the symlink-replace and broken-symlink tests. CI lanes that don't have the prerequisite now report SKIPPED rather than PASS-with-zero-assertions.

2. **Rename mock delegation** — captured `const originalRename = fs.renameSync` before installing the mock and replaced the bogus `fs.renameSync.wrapped` reference with `originalRename.call(fs, src, dest)`. Now matches the ENOSPC test's pattern and actually delegates if the defensive second-call branch ever fires.

Tests still 13/13 pass on macOS (skipped=0 because not on Win32). Verified by running the suite locally.

PR #3637, opened 2026-05-16
fix(3632): report each ts sibling independently in lint-shared-module-handsync

### LINKED ISSUE #3632: fix(lint-shared-module-handsync): early-exit on basename match misses unauthorized siblings
## Summary

`scripts/lint-shared-module-handsync.cjs` (lines 227–233) exits its inner loop with `continue` the moment ANY allowlisted `(cjs, ts)` cooperating-sibling pair matches the current candidate. If two sibling pairs share the same basename — e.g. `sdk/src/<name>.ts` AND `sdk/src/query/<name>.ts` — and only one of them is on the allowlist, the linter silently passes the unallowlisted pair through.

CI is supposed to fail on new hand-synced drift; this gap means new unauthorized siblings can slip in undetected as long as they share a basename with an allowlisted pair.

## Reproduction

1. Pick an allowlisted basename in `scripts/shared-module-handsync-allowlist.json` (e.g. one of the documented cooperating siblings).
2. Add a NEW `<basename>.cjs` ↔ `<basename>.ts` pair at a DIFFERENT path (e.g. `sdk/src/query/<basename>.ts` when the allowlisted version is at `sdk/src/<basename>.ts`).
3. Run `node scripts/lint-shared-module-handsync.cjs`.
4. Observe: lint exits 0 even though the new unallowlisted pair exists.

## Root cause

```js
// scripts/lint-shared-module-handsync.cjs:227-233 (approximate)
for (const candidate of pairs) {
  let matchedCooperating = false;
  for (const allow of allowlist) {
    if (allow.basename === candidate.basename) {
      matchedCooperating = true;
      continue;  // ← exits the inner loop on FIRST basename match
    }
  }
  if (matchedCooperating) continue;
  // ... reports drift
}
```

The match check uses basename equality, not full-path equality. The early `continue` means the second/third candidate sharing that basename never gets evaluated.

## Fix

Compare allowlist entries against the candidate's full `(cjsPath, tsPath)` tuple, not just the basename, before deciding to skip.

## Acceptance criteria

- [ ] Multi-sibling fixture: two `(cjs, ts)` pairs share a basename; only one is allowlisted; lint reports the other.
- [ ] Regression test pinning the multi-sibling fixture so the early-exit pattern cannot return.
- [ ] Existing single-sibling cases remain green (the documented 22 cooperating siblings in the current allowlist).

## Discovery

Surfaced by `/codex:review` on PR #3627 against a stale base that incidentally included the #3577 merged diff. The finding is on already-merged code (origin/main).

## Privacy

No PII — finding is purely about code paths in this repository.

trek-e:
## Linked Issue

Fixes #3632

> The linked issue has `confirmed-bug` label.

---

## What was broken

`scripts/lint-shared-module-handsync.cjs` collected all ts candidates for a given cjs and ran `cooperatingPairs.has(...)` inside `.some()`. As soon as **any** `(cjs, ts)` pair matched the allowlist, the candidate was skipped wholesale — so when a single cjs had two ts siblings sharing the same basename (e.g. `sdk/src/foo.ts` and `sdk/src/query/foo.ts`) and **only one** pair was allowlisted, the unallowlisted sibling silently passed CI.

> Note: issue body's pseudocode referenced a basename-only `===` check; the actual short-circuit is the `.some()` at lines 227–229. The user-visible failure mode (unauthorized siblings escape lint) matches the issue exactly.

The bug was on `feat/3575-enforcement-hardening` (PR #3577, open) — not on `origin/main` as the issue body asserted. This PR targets that branch.

## What changed

`scripts/lint-shared-module-handsync.cjs` now iterates each ts candidate for a given cjs and classifies it independently:

- pair in `cooperatingSiblings` → silent pass
- pair in `migrateMeBacklog` → warning (still passes CI)
- otherwise → error (fails CI)

A cjs with mixed-status siblings now produces both: a clean pass for the allowlisted pair and an error for the unauthorized one.

## Regression test

Added `reports unallowlisted ts sibling when another ts sibling for the same cjs IS allowlisted (#3632)` to `tests/lint-shared-module-handsync.test.cjs`. The test:

1. Creates a fixture with `bin/lib/multi-sibling.cjs`, `sdk/src/multi-sibling.ts`, and `sdk/src/query/multi-sibling.ts`.
2. Allowlists only the `sdk/src/<name>.ts` pair.
3. Asserts the `sdk/src/query/<name>.ts` sibling appears in `payload.errors[].tsPaths` and the allowlisted one does NOT.
4. Verified red on the prior code (`status === 0` instead of 1) and green after the fix.

All 11 tests in the file pass; real-tree lint on `feat/3575-enforcement-hardening` is unchanged (22 cooperating siblings, 0 errors, 0 backlog pairs).

## Risk

Behavior change is strictly: previously-silent-because-one-sibling-was-allowlisted unauthorized pairs now correctly fail lint. The intended invariant of the linter ("every (cjs, ts) pair must be explicitly classified") is restored. No existing allowlisted pair changes status.

> *This PR description was assisted by AI during issue triage and fix authoring.*


trek-e:
Per maintainer policy: PRs should target `main`, not feature branches. This PR's base is `feat/3575-enforcement-hardening`, which violates that policy.

Independent verification on issue #3632 confirms the bug lives only inside PR #3577's own diff — `scripts/lint-shared-module-handsync.cjs` does not exist on `main`. There is no main-targeting fix to write here.

Recommendation: close this PR and fold the fix (the per-sibling classification rewrite in `scripts/lint-shared-module-handsync.cjs` + the regression test in `tests/lint-shared-module-handsync.test.cjs`) into PR #3577 itself, so #3577 merges to `main` with the bug already fixed.

PR #3639, opened 2026-05-16
fix(3631): family routers forward --raw to SDK bridge as mode:'raw'

### LINKED ISSUE #3631: fix(routers): SDK dispatch path drops --raw flag in family routers (regression from #3577)
## Summary

SDK dispatch path in `get-shit-done/bin/lib/phase-command-router.cjs:51` calls `output(result.data)` without forwarding the `raw` flag. Shell consumers that rely on raw-mode contracts (e.g. `gsd-tools phase next-decimal --raw` printing just the scalar token) now receive a JSON-wrapped object when the SDK dispatch path fires.

## Reproduction

```bash
# Pre-#3577 (CJS path): prints just "1.1"
# Post-#3577 (SDK path): prints { "phase": "1.1", ... } as JSON
gsd-tools phase next-decimal --raw 1
```

The same pattern (`output(result.data)` without `raw`) appears in the other SDK-bridge family routers (`get-shit-done/bin/lib/*-command-router.cjs` introduced/touched by #3577 / #3558 / #3548).

## Root cause

When the family router dispatches to the SDK bridge (PR #3577, Phase 6 of #3524), the success path forwards `result.data` to `output()` but drops the second positional `raw` argument that the CJS-only path used to pass. `output(data, raw, ...)` in `get-shit-done/bin/lib/core.cjs` treats `raw === true` as "print as-is, no JSON wrap" — the SDK call site never opts in.

## Acceptance criteria

- [ ] Every `*-command-router.cjs` SDK dispatch path that forwards a result to `output()` passes the `raw` flag through verbatim.
- [ ] Behavioral regression test: `gsd-tools phase next-decimal --raw <N>` produces the raw scalar form, not JSON, regardless of whether dispatch goes through CJS or SDK.
- [ ] Same coverage for at least one other family that exposes `--raw`.

## Discovery

Surfaced by `/codex:review` on PR #3627 against a stale base that incidentally included the #3577 merged diff. The finding is on already-merged code (origin/main), not on #3627's added test files.

## Privacy

No PII — finding is purely about code paths in this repository.

trek-e:
## Linked Issue

Fixes #3631

> The linked issue has `confirmed-bug` label.

---

## What was broken

#3577 routed every family subcommand through the SDK bridge with a hardcoded `mode: 'json'`. Under `--raw`, the bridge returned the typed JSON IR and each router called `output(result.data)` — bypassing `output()`'s rawValue branch. Shell consumers expecting scalar tokens (`gsd-tools phase next-decimal --raw 1` → `1.1`) received the JSON-stringified IR instead.

The bug lives on `feat/3575-enforcement-hardening` (PR #3577, open) — not on `origin/main` as the issue body asserted. This PR targets that branch so the regression doesn't ship.

## What changed

**Routers (`*-command-router.cjs`).** Every SDK dispatch path (init, phase, phases, roadmap, state, validate, verify) now:

1. Requests `mode: raw ? 'raw' : 'json'` from the bridge.
2. Under `--raw`, calls `output(null, true, stringForm)` so the rawValue branch of `output(result, raw, rawValue)` lands the scalar string on stdout without re-stringification.
3. Under default mode, keeps the existing `output(result.data)` JSON behaviour.

The `state-command-router` had a partial fix via `dispatchViaSdk(... rawFormatter)`. The fallthrough `output(result.data)` when no rawFormatter was present is the same regression and was patched to use the rawValue branch under `--raw`.

**SDK runtime bridge worker.** Wired `formatNativeRaw` to `formatQueryRawOutput` so `mode: 'raw'` requests through the sync bridge actually run the SDK's raw-projection registry rather than the generic `toRaw` fallback.

**Raw projection registry.** Extended `formatQueryRawOutput` for the two commands in the issue acceptance criteria:

- `phase.next-decimal` → `data.next`
- `roadmap.get-phase` → `data.section`

Other registered projections (`state.load`, `commit`, `config-set`, `state.begin-phase`) are unaffected.

## Regression test

Added `tests/bug-3631-router-raw-flag.test.cjs`:

1. `gsd-tools phase next-decimal --raw 1` emits a scalar phase token (e.g. `1.1` / `01.1`) — verified to NOT match `^\{`.
2. `gsd-tools roadmap get-phase --raw 2` emits the section text containing the `Phase 2:` heading — verified to NOT match `^\{`.

Both tests start RED on the prior code (JSON output) and GREEN after the fix.

## Adjacent fixes

`tests/phases-command-router.test.cjs` and `tests/roadmap-command-router.test.cjs` mock the CJS-side `phase`/`milestone`/`roadmap` handlers but routing now prefers the SDK bridge (sdk/dist is built by `pretest`). The mocks were bypassed and the SDK side failed because the test cwd had no `.planning/` fixture. Setting `GSD_WORKSTREAM` in `before()/after()` deterministically routes through the CJS handlers the tests were written against — without weakening their assertions.

## Acceptance criteria mapping

- [x] Every `*-command-router.cjs` SDK dispatch path that forwards a result to `output()` passes the `raw` flag through.
- [x] Behavioural regression test: `gsd-tools phase next-decimal --raw <N>` produces the raw scalar form, not JSON, regardless of dispatch path.
- [x] Same coverage for at least one other family that exposes `--raw` (`roadmap get-phase --raw`).

> *This PR description was assisted by AI during issue triage and fix authoring.*

### DIFF
--- a/.changeset/fix-3631-sdk-raw-flag-routers.md
+++ b/.changeset/fix-3631-sdk-raw-flag-routers.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3631
+---
+**SDK dispatch path in family routers now honours `--raw`** — `phase next-decimal --raw`, `roadmap get-phase --raw`, and other family-router commands that route through the SDK bridge now emit the same scalar string the CJS path emitted before #3577. Routers request `mode: 'raw'` from the bridge under `--raw`; the sync-bridge worker wires `formatNativeRaw` to `formatQueryRawOutput` so the bridge returns the per-command projection. Routers then pass the formatted string through `output()`'s rawValue branch instead of JSON-stringifying it.
--- a/get-shit-done/bin/lib/init-command-router.cjs
+++ b/get-shit-done/bin/lib/init-command-router.cjs
@@ -31,11 +31,12 @@ function routeInitCommand({ init, args, cwd, raw, parseNamedArgs, error }) {
         registryArgs,
         legacyCommand: 'init',
         legacyArgs,
-        // Always request typed JSON from the bridge; CJS `output(data, raw)` handles
-        // user-facing rendering. Passing `mode: 'raw'` would make the bridge
-        // pre-render result.data to a JSON string that the CJS output path then
-        // double-stringifies (returning a JSON string of a JSON string).
-        mode: 'json',
+        // #3631: under --raw, request mode:'raw' so the bridge runs the SDK's
+        // raw projection (formatQueryRawOutput) and returns the scalar string
+        // CJS callers used to print. We then bypass output()'s JSON-stringify
+        // path by passing rawValue (the third positional). With mode:'json',
+        // output() emits the JSON IR as before.
+        mode: raw ? 'raw' : 'json',
         projectDir: cwd,
       });
       if (!result.ok) {
@@ -44,7 +45,11 @@ function routeInitCommand({ init, args, cwd, raw, parseNamedArgs, error }) {
           : `init ${registryCommand} failed (${result.errorKind})`);
         return;
       }
-      output(result.data);
+      if (raw) {
+        output(null, true, typeof result.data === 'string' ? result.data : String(result.data ?? ''));
+      } else {
+        output(result.data);
+      }
     };
   }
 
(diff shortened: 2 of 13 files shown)


trek-e:
Per maintainer policy: PRs should target `main`, not feature branches. This PR's base is `feat/3575-enforcement-hardening`, which violates that policy.

Independent verification on issue #3631 confirms the bug lives only inside PR #3577's own diff — on `main`, `gsd-tools phase next-decimal --raw 1` already emits the correct scalar output, and the SDK-dispatch path that drops `--raw` does not exist there. There is no main-targeting fix to write here.

Recommendation: close this PR and fold the fix (router `mode: raw` forwarding + `formatQueryRawOutput` projections + the regression test) into PR #3577 itself, so #3577 merges to `main` with the bug already fixed.

PR #3658, opened 2026-05-17
Match `gsd-sdk query commit` in graphify auto-update hook (#3653)

cristianuibar:
## Linked Issue

Fixes #3653

## What was broken

The PostToolUse Bash hook `hooks/gsd-graphify-update.sh` substring-matched `tool_input.command` against direct shell git ops to detect HEAD-advancing commits and trigger a graphify rebuild. `gsd-sdk query commit` invokes git via `spawnSync('git', ['commit', ...])` — direct argv, never via a shell — so the literal `git commit` substring never appears in the Bash tool's command string. Gate 2 short-circuited with `exit 0` on every SDK-issued commit, including the `phase.complete`-following commit that closes every phase. `.planning/graphs/` silently drifted stale.

## What this fix does

Extends Gate 2's case statement to also accept `*"gsd-sdk query commit"*` — the user-facing invocation the Bash tool actually sees in `tool_input.command`.

```bash
case "$COMMAND" in
  *"git commit"*|*"git merge"*|*"git pull"*|*"git rebase --continue"*|*"git cherry-pick"*) ;;
  *"gsd-sdk query commit"*) ;;
  *) exit 0 ;;
esac
```

## Root cause

The hook's contract was authored against the only documented commit path at the time (bare `git commit`). The SDK introduced `gsd-sdk query commit` later as the canonical workflow-side commit interface, invoking git via `spawnSync` for safe argv passing — which is correct for the SDK but invisible to hooks that scrape `tool_input.command`. No side-channel was emitted at the same time, so every consumer hook degraded silently.

## Scope

Only `sdk/src/query/commit.ts` advances HEAD on the main repo. Other SDK verbs in the `gsd-sdk query` namespace (`phase.complete`, `roadmap.update-plan-progress`, `state.begin-phase`) mutate `.md` files but do not invoke git themselves — they rely on a subsequent `gsd-sdk query commit` to land the change. Matching them would cause a 1:1 false-positive rebuild per state mutation. The new negative tests lock that decision in.

`commitToSubrepo()` (commit.ts:305–382) also uses `spawnSync('git', …)` but operates inside a submodule and does not move the main-repo HEAD that graphify tracks — out of scope.

## Rejected alternatives

- **SDK-side stdout marker (issue's Fix A).** Current hook reads only `tool_input.command`, not `tool_response.stdout`; emitting a marker would require widening what the hook reads AND adding a new output line — strictly more change for the same coverage the narrow matcher delivers.
- **Shell-invoke the SDK commit (issue's Fix C).** Loses `spawnSync`'s safe argv passing for commit messages and file paths; gain is purely cosmetic now that the hook matches the user-facing invocation directly.

### How I verified the fix

- `node --test tests/feat-3347-graphify-auto-update-hook.test.cjs` — 22/22 pass, including the new positive case for `gsd-sdk query commit` and three negative cases for non-HEAD-advancing SDK verbs.
- `node --test tests/enh-3170-graphify-commit-staleness.test.cjs tests/bug-3579-graphify-hook-publish.test.cjs tests/feat-3347-graphify-auto-update-config.test.cjs` — 26/26 pass.
- `node scripts/build-hooks.js` regenerates dist; `diff hooks/gsd-graphify-update.sh hooks/dist/gsd-graphify-update.sh` empty.
- `node scripts/lint-docs-required.cjs` → `ok_no_triggering_fragments` (changeset frontmatter passes parser).
- Manual fire-drill from repo root:
  ```bash
  echo '{"tool_name":"Bash","tool_input":{"command":"gsd-sdk query commit \"docs: probe\" --files README.md"}}' \
    | bash hooks/gsd-graphify-update.sh; echo "exit=$?"
  ```
  Exits 0 and reaches Gate 2 (no longer short-circuits).

### Regression test added?

- [x] Yes — added a positive matcher case for `gsd-sdk query commit` and three negative matcher cases (`gsd-sdk query phase.complete`, `gsd-sdk query roadmap.update-plan-progress`, `gsd-sdk query state.begin-phase`) in `tests/feat-3347-graphify-auto-update-hook.test.cjs`.

### Platforms tested

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] N/A

(CI covers Ubuntu 22 / 24 and macOS-latest.)

### Runtimes tested

- [x] Claude Code (the hook's primary runtime — PostToolUse on the Bash tool)
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] N/A

## Checklist

- [x] Issue linked above with `Fixes #3653`
- [x] Regression test added
- [x] All existing tests pass (`npm test` adjacent suites green)
- [x] `.changeset/3653-graphify-hook-sdk-commit-visibility.md` added (`type: Fixed`, `pr: 3658`)

## Breaking changes

None. The hook only matches one additional command shape; all existing matchers and bail-paths are unchanged. Opt-in (`graphify.enabled` AND `graphify.auto_update` both `true`) is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Opt-in Graphify auto-update setting and bundled hook that refreshes graph build status after recognized commit actions.

* **Bug Fixes**
  * Hook now correctly triggers for SDK-issued commits (e.g., `gsd-sdk query commit`), preventing stale graph data and avoiding prefix-collision matches.

* **Tests**
  * Added regression tests covering SDK commit forms and ensuring other query verbs do not trigger updates.

* **Documentation**
  * Added release notes describing auto-update behavior and UI setting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/get-shit-done/pull/3658?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cristianuibar:
Tri-CLI cross-review summary for the maintainer:

**Codex CLI** (`codex review --base origin/main`) — `LGTM`. Verified the substring catches every realistic Bash-tool invocation shape (`npx`, full path, env-var prefix, `command`, line continuation). Cross-checked `sdk/src/query/` and confirmed the only HEAD-advancing verbs are `commit` and `commit-to-subrepo` (the latter is also caught by the substring as a side effect). False-positive profile unchanged from existing matchers; Gates 3–8 still absorb the cost. Two NITs: PR text says "only the SDK `commit` verb" (technically also `commit-to-subrepo`), and an explicit test for `commit-to-subrepo` was not added (covered transitively by the substring). Both non-blocking.

**Gemini CLI** (`gemini -p ... --approval-mode plan`) — `LGTM`. Same independent finding on the matcher being whitespace/quote-sensitive (e.g. won't match `"gsd-sdk" query commit` or doubled spaces) — consistent with the existing `*"git commit"*` matcher, accepted as a simplicity tradeoff. Confirmed the substring catches `commit-to-subrepo` as a bonus. Test structure clean, no flakiness vectors.

**Grok CLI** — skipped. Local auth token returned `AuthorizationRequired`; only interactive logins (`grok login --oauth` or `--device-auth`) recover it, no env-var path. Coverage from the other two reviewers + CodeRabbit was deemed sufficient.

**CodeRabbit** — re-review now `pass` after the changeset frontmatter fix (`type: Bug Fix` → `type: Fixed`, `pr: 3658` added in 0b5899da).

All 13 CI checks pass (docs-lint, changeset-lint, security, size-check, lint-tests, test on ubuntu-22/24 + macos-latest, plus all template/branch policies). Ready to merge.

cristianuibar:
Re: the earlier `gsd-pr-template-policy` warning comment from 02:44:29Z — that fired against the original PR body. Since the body has been rewritten to the Fix template, the policy now passes:

- CI check `Pull request template format` on the latest commit (`0b5899da`): `success` ([run](https://github.com/gsd-build/get-shit-done/actions/runs/25979681104/job/76365977015))
- Local validator run: `{"valid":true,"action":"pass","trusted":true,"authorAssociation":"CONTRIBUTOR","template":"fix","reason":"PR body uses a typed pull request template.","missingHeadings":[]}`

The bot comment is stale — the workflow only posts on warn/close and doesn't auto-minimize on subsequent pass. I don't have permission to delete or `MinimizeComment` it; flagging here so it's obvious in the timeline.

trek-e:

## Review: APPROVE

Tight, well-scoped fix. Verified:

**Shell-script correctness**
- `$COMMAND` is double-quoted in the `case` expression; the new `*"gsd-sdk query commit"*` arm is a literal-substring glob with no expansion of user content into eval/exec contexts.
- Lock-file / PID logic on the hook's gate path is untouched.
- Hook fail-opens (exits 0) on every gate miss — documented intent (`"Never blocks the user-facing tool call"`); no regression.

**False-positive surface (non-blocking observation)**
- The substring `gsd-sdk query commit` will also match a hypothetical sibling verb that starts with `commit` (e.g., `commit-status`). Today no such verb exists and the negative tests pin the three non-HEAD sibling verbs. Worth a note only.

**Test quality (CONTRIBUTING.md compliance)**
- Tests assert on the **structured outcome** (existence of `.planning/graphs/.last-build-status.json`), not on stdout text or source-grep.
- Each test creates an isolated tmp git repo with config and registers `t.after(cleanup)`. No vacuous truth, no swallowed exceptions, no source-grep.
- Positive case adds one matcher, three negative cases lock in non-dispatch for sibling SDK verbs.

**CodeRabbit status**
- 1 actionable: changeset `type: Bug Fix` → `type: Fixed`. **Applied** in `0b5899d`. No outstanding CR items.

**CI**
- 13/13 checks COMPLETED, 1 SUCCESS rollup. Mergeable.

The bug analysis in the PR body is accurate — `spawnSync('git', [...])` from the SDK never surfaces `git commit` as a substring of `tool_input.command`, so Gate 2 was silently dropping every SDK-issued commit. The narrow matcher addition is the minimum viable fix; rejected alternatives (stdout marker, shell-invoke) are justified.

Ready to merge.

trek-e:

## Codex Adversarial Review — verdict: needs-attention

Supplementing my prior `APPROVE`. Codex found a real over-match I flagged at low confidence — and confirmed it's not hypothetical:

### Finding [medium] — `gsd-sdk query commit` substring also matches `commit-to-subrepo`

`hooks/gsd-graphify-update.sh` dispatches when `tool_input.command` contains the raw substring `gsd-sdk query commit`. This is not token-bounded and **will also match the existing command `gsd-sdk query commit-to-subrepo`** (confirmed: defined in `sdk/src/query/command-static-catalog-domain.ts`, `sdk/src/query/commit.ts`, `sdk/src/query/command-aliases.generated.ts`, `get-shit-done/bin/lib/commands.cjs`, and documented in `get-shit-done/references/git-integration.md` and `get-shit-done/references/git-planning-commit.md`).

When `commit-to-subrepo` runs, the hook writes `status=running` and detaches a graph rebuild without proving the current repo HEAD actually advanced — the subrepo-only flow does not move outer HEAD. Result: spurious rebuilds, misleading status churn, and potential lock-timing interference with real updates.

Also: any shell text that merely *contains* the phrase (commits with that prose in the message, comments in scripts under `tool_input.command` echo, etc.) would fire too.

### Required changes

1. **Tighten Gate 2 matching to token boundaries.** Don't use raw substring containment. The cleanest path is parsing the command and checking for an exact `gsd-sdk query commit` argv sequence — or at minimum, anchor with surrounding word boundaries that reject `commit-to-subrepo`, `commit-status`, etc.
2. **Add a negative regression test for `gsd-sdk query commit-to-subrepo`** explicitly asserting the hook does NOT dispatch. The current 3 sibling-verb tests cover `phase.complete`, `roadmap.update-plan-progress`, `state.begin-phase` — but `commit-to-subrepo` is the real footgun and is not pinned.
3. **(Optional, defense-in-depth)**: gate dispatch on verified HEAD movement (or a confirmed-success signal from the commit command's output) before writing `status=running` / spawning rebuild. This makes the hook robust to future matcher mistakes.

### Verdict revision

Reverting from `approved to merge` to `needs-changes`. The fix is conceptually right but the matcher is too loose against the actual current command surface. Once the matcher is tightened and the `commit-to-subrepo` negative test is in place, this is mergeable.

cristianuibar:
@trek-e review changes applied

trek-e:

## Re-review after changes applied

### Verdict: approve

The prior CR's required changes are addressed. The matcher is now correctly bounded and the `commit-to-subrepo` negative test is in place. CI is 19/19 green.

---

### Code review

- **Gate 2 matcher is correctly tightened.** `*"gsd-sdk query commit"|*"gsd-sdk query commit "*)` accepts bare `gsd-sdk query commit` (no trailing args) and `gsd-sdk query commit <args>` (space-delimited args), while excluding `commit-to-subrepo` and `commit-status` — verified with live `bash -c 'case ...'` tests.
- **Shell safety unchanged.** `$COMMAND` is never eval'd or expanded into an exec context; it's only used as the `case` operand. No injection surface introduced.
- **Changeset frontmatter corrected.** `type: Fixed` with `pr: 3658` is present; docs-lint passes (`ok_no_triggering_fragments`).

---

### Test coverage

- Positive cases: `gsd-sdk query commit "..." --files ...` and `npx gsd-sdk query commit ...` both dispatch. Added with this PR.
- Negative — `commit-to-subrepo`: new explicit test asserts no dispatch, locking the collision fix in. This was the specific gap called out in prior CR.
- Negative — sibling verbs: `phase.complete`, `roadmap.update-plan-progress`, `state.begin-phase` all pinned from initial commit.
- Tests assert on structured outcome (`.last-build-status.json` existence), not stdout text — CONTRIBUTING.md compliant.
- Coverage is behavioral and regression-safe; the original bug cannot recur without a test catching it.

---

### Error handling

- Hook continues to fail-open (exit 0) on every gate miss, including malformed JSON input. Unchanged from original.
- No new try/catch surfaces; `node -e` parsing in the hook has the same `|| true` / `|| exit 0` guards as before.

---

### Rubber-duck — root cause vs symptom

**Root cause:** SDK's `commit.ts` invokes git via `spawnSync('git', [...args])` — direct argv, no shell — so the literal string `git commit` never appears in `tool_input.command`. The hook's Gate 2 could never match SDK-initiated commits.

**Symptom:** `.planning/graphs/` silently drifted stale at the end of every phase; the `phase.complete`-following commit (always SDK-issued) always missed the trigger.

**Fix addresses:** both root cause and symptom. The added case arm targets the only user-facing command string Claude Code's Bash tool ever carries for an SDK commit. It doesn't change the SDK internals (which is correct — Fix C was properly rejected), it adds the missing observation point at the hook layer.

**Sibling coverage:** `phase.complete`, `roadmap.update-plan-progress`, `state.begin-phase` are intentionally excluded because they don't invoke git themselves; they depend on a subsequent `gsd-sdk query commit` call, which is now matched. This is correct and the negative tests lock it in.

---

### Options evaluation

**Alternatives considered:** (a) stdout marker from SDK (issue Fix A) — more change for same coverage, requires widening hook to read `tool_response.stdout`; (b) `execSync` shell-invoke (Fix C) — loses safe argv passing; (c) widen case to all `gsd-sdk query *` verbs — causes spurious rebuilds per state mutation.

**Coverage of issue scope:** complete for the stated bug. The issue also proposed a durable stdout-marker approach (Fix A) as the long-term solution; this PR takes the minimal-change hook-side path (Fix B) which is acceptable as the primary fix. Fix A can follow in a separate enhancement.

**Chosen approach correctness:** the two-pattern split (`*"gsd-sdk query commit"` | `*"gsd-sdk query commit "*)`) is the idiomatic bash-case way to match a word at string end or word followed by space. It's minimal and precise.

---

### Unresolved CR threads addressed

- **[CodeRabbit] `type: Bug Fix` → `type: Fixed` in changeset frontmatter:** resolved in commit `0b5899d`. ✅
- **[trek-e] Codex finding — `commit-to-subrepo` prefix collision:** resolved in commit `840752c`. Matcher changed to two-pattern bounded form; negative test for `commit-to-subrepo` added. ✅

No unresolved CR threads remain.

---

### Codex adversarial pass

**Finding 1 [minor]:** Tab-separated or double-spaced invocations (e.g. `gsd-sdk query commit\t--message x`) don't match either arm — confirmed by shell test. In practice Claude Code's Bash tool always emits single-space-separated command strings, so this is not a realistic failure path. If a future runtime normalizes whitespace differently, the fix would be a `[[ =~ ]]` regex check. Not blocking.

**Finding 2 [minor]:** Commands whose text merely *contains* the phrase (e.g. `echo gsd-sdk query commit`) trigger a spurious rebuild. This is also true of the existing `*"git commit"*` arms and is inherent to the substring-match design. Gate 5 (default-branch check) and Gate 8 (PID lock) bound the cost. Not blocking.

Codex confirmed: `commit-to-subrepo` exclusion is correct; no critical or major issues found.

---

### Recommended labels

- add: `review: approved`
- remove: `review: changes requested`, `needs changes`

Issue #3055, opened 2026-05-03
bug: top-level branching_strategy in .planning/config.json silently becomes "none" — execute-phase skips branch creation


wdi-dave-roberts:
### GSD Version

User-scope tools at `~/.claude/get-shit-done/VERSION` → **1.39.1**.
SDK CLI (`which gsd-sdk` → `~/.npm/_npx/.../get-shit-done-cc/sdk/dist/cli.js`) ships bundled tools at **1.38.5**. Both versions reproduce.

### Runtime / OS / Node / Shell

Claude Code 2.1.x / WSL2 Ubuntu (Linux 6.6 x86_64) / Node v22.22.2 / zsh.

### Installation Method

Updated from a previous version via `npx get-shit-done-cc --claude --global`.

### What happened?

`.planning/config.json` declares `branching_strategy: "phase"` at the top level. The validator flags it as unknown, the value is dropped, and `loadConfig` falls back to the `none` default. Every workflow that goes through `loadConfig` (including `/gsd-execute-phase`'s `handle_branching` step) sees `branching_strategy: "none"` and skips branch creation, so phase commits land on the operator's current branch (here: `main`) instead of `gsd/phase-{N}-{slug}`.

The contradiction: `gsd-sdk query config-get branching_strategy` returns `"phase"` (it reads `config.json` directly without validation, see `config.cjs:389-410`), so the operator has no easy way to notice the value the workflow actually consumed is `"none"`.

The single-line warning *does* fire on stderr:
```
gsd-tools: warning: unknown config key(s) in .planning/config.json: branching_strategy — these will be ignored
```
But it scrolls past inside long `gsd-sdk` outputs, and the workflow continues regardless.

This is structurally identical to #2638 (top-level `sub_repos` written then warned as unknown), and is the latest in a class of recurring config-schema-drift bugs (#1535, #1542, #2047/#2052, #2687/#2706, #2798/#2816, #2653/#2670).

### What did you expect?

Either of:

1. **Self-healing migration** (preferred, mirrors #2655's fix for `sub_repos`): on `loadConfig`, if `parsed.branching_strategy` is set and `parsed.git?.branching_strategy` is unset, move the value into `parsed.git.branching_strategy`, persist to disk, and continue with the correct strategy. Same shape as the `multiRepo → planning.sub_repos` migration already in `core.cjs`.

[...]

Issue #2798, opened 2026-04-28
context_window missing from VALID_CONFIG_KEYS — /gsd-settings-advanced cannot set it


luanngosodadigital:
## Summary

`/gsd-settings-advanced` Section 6 prompts the user to set `context_window`, but `gsd-sdk query config-set context_window <n>` rejects the write because `context_window` isn't in the `VALID_CONFIG_KEYS` allowlist. The documented happy path is unreachable.

**Version:** `get-shit-done-cc@1.38.1`

## Repro

```
$ gsd-sdk query config-set context_window 1000000
Error: Unknown config key: "context_window". Did you mean: context?
```

## Expected

Write succeeds. The workflow at `get-shit-done/workflows/settings-advanced.md` Section 6 documents `context_window` as a settable runtime knob, and the SDK already reads it elsewhere.

## Evidence

1. **Workflow doc says it's settable:** `get-shit-done/workflows/settings-advanced.md` — Section 6 prompts for `context_window` and routes the write through `gsd-sdk query config-set`.
2. **SDK reads the key:** `sdk/dist/query/init.js:190` (`config.context_window ?? 200000`) and `sdk/dist/query/validate.js:575` (validates as positive integer, recommends `200000` or `1000000`).
3. **Allowlist omits it:** `sdk/dist/query/config-mutation.js` (~line 55–82) — `VALID_CONFIG_KEYS` includes peer top-level keys (`response_language`, `search_gitignored`, `model_profile`, `context`, `project_code`, etc.) but not `context_window`.

So three independent sites agree this should be a settable key — the validator is the only outlier.

## Bonus issue: misleading typo suggestion

The error suggests `Did you mean: context?`, but `context` is a different key — a mode selector with values `dev|research|review`, unrelated to context window size. A user who follows the suggestion gets `Invalid context value '1000000'. Valid values: dev, research, review`.

Suggest adding `context_window` to `CONFIG_KEY_SUGGESTIONS` (or seeding the LCP fallback so it doesn't collide with `context`).

## Proposed fix

Add `'context_window'` to `VALID_CONFIG_KEYS` in `sdk/src/query/config-mutation.ts` (also mirrored in `get-shit-done/bin/lib/config.cjs` for the CJS path). One-line change.

## Workaround

Edit `.planning/config.json` directly to add `"context_window": 1000000` at the top level. The SDK reads it correctly; only the writer rejects it.

luanngosodadigital:
[...]

Issue #2687, opened 2026-04-25
bug: loadConfig warns on `review` and other dynamic-pattern containers in .planning/config.json despite valid contents


j2h4u:
### GSD Version

`upstream/main` at `cd057255` (`package.json` version `1.38.2`).

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

v24.14.1

### Shell

/bin/bash

### Installation Method

Updated from a previous version

### What happened?

`loadConfig` in `get-shit-done/bin/lib/core.cjs` warns on every read of `.planning/config.json` that contains a `review.models.<cli>` entry — even though that key is documented, has a dynamic-pattern validator in `bin/lib/config-schema.cjs:77`, and is the canonical write path used by `/gsd-settings-integrations`, `docs/CLI-TOOLS.md`, and `workflows/settings-integrations.md`.

```text
gsd-tools: warning: unknown config key(s) in .planning/config.json: review — these will be ignored
```

The keys are **not** ignored at runtime — `workflows/review.md:155-158` reads them via `gsd-sdk query config-get review.models.<cli>` successfully. The warning is a false positive, fires once per `loadConfig` call, and reaches users on every `gsd-tools.cjs init phase-op` and similar reads.

#### Root cause

`core.cjs` builds a `KNOWN_TOP_LEVEL` Set used to detect unknown top-level keys (`get-shit-done/bin/lib/core.cjs:340-355`):

```js
const { VALID_CONFIG_KEYS } = require('./config.cjs');
const KNOWN_TOP_LEVEL = new Set([
  // Extract top-level key names from dot-notation paths (e.g., 'workflow.research' → 'workflow')
  ...[...VALID_CONFIG_KEYS].map(k => k.split('.')[0]),
  // Section containers that hold nested sub-keys
  'git', 'workflow', 'planning', 'hooks', 'features',
  // Internal keys loadConfig reads but config-set doesn't expose
  'model_overrides', 'agent_skills', 'context_window', 'resolve_model_ids', 'claude_md_path',
  // Deprecated keys (still accepted for migration, not in config-set)
  'depth', 'multiRepo',
]);
```

This set is derived from static `VALID_CONFIG_KEYS` plus a hard-coded section-container list (`'git', 'workflow', 'planning', 'hooks', 'features'`). It does **not** consult `DYNAMIC_KEY_PATTERNS` from `config-schema.cjs:75-86`, which currently registers five dynamic namespaces:

[...]

Issue #2653, opened 2026-04-24
bug: SDK config-set rejects documented config keys accepted by CJS config-set


mlmeehan:
### GSD Version

`main` at `a72bebb` (`package.json` version `1.38.2`).

A related drift also exists in npm `get-shit-done-cc@1.38.3`, but the exact key counts differ because the published package does not contain the current `config-schema.cjs` split.

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

Reproduced with Node v25.9.0 and Node v22.22.2. The bug is Node-agnostic.

### Shell

/bin/zsh

### Installation Method

Cloned the repo and ran `npm -g install` from the `sdk` subdirectory (not a match for any of the template's dropdown options). The bug is install-method-agnostic — it is in the SDK source itself.

### What happened?

The SDK config mutation handler has its own hardcoded config allowlist, dynamic-key regexes, suggestion map, and value validators in:

```text
sdk/src/query/config-mutation.ts
```

The CJS path uses the authoritative schema in:

```text
get-shit-done/bin/lib/config-schema.cjs
get-shit-done/bin/lib/config.cjs
```

On `main` at `a72bebb`, those two paths have drifted. As a result, documented commands using `gsd-sdk query config-set` reject valid config keys that the CJS path accepts and that the runtime reads once present in `.planning/config.json`.

#### Drift on `main` (`a72bebb`)

| Dimension | CJS source | SDK source | Delta |
|-----------|------------|------------|-------|
| Exact config keys | `config-schema.cjs` `VALID_CONFIG_KEYS` | `config-mutation.ts` `VALID_CONFIG_KEYS` | **28 keys missing in SDK**, **1 SDK-only key** |
| Dynamic key patterns | `config-schema.cjs` `DYNAMIC_KEY_PATTERNS` | inline regexes in `config-mutation.ts` | **3 CJS patterns missing in SDK** |
| Suggestions | `config.cjs` `CONFIG_KEY_SUGGESTIONS` | `config-mutation.ts` `CONFIG_KEY_SUGGESTIONS` | **4 CJS suggestions missing in SDK** |
| Value validators | `config.cjs` `cmdConfigSet` | `config-mutation.ts` `configSet` | SDK only validates `context`; CJS also validates drift and post-planning-gap keys |

#### Exact keys missing from SDK on `main`

```text
claude_md_assembly.mode
claude_md_path
context_window
graphify.build_timeout
graphify.enabled
intel.enabled
[...]

PR #3771, opened 2026-05-20
fix(3749): port strategy-branch creation from CJS to SDK commit handler

### LINKED ISSUE #3749: bug: PR #1279 strategy-branch-before-commit fix never ported to SDK commit handler — pre-execution commits still land on wrong branch
### GSD Version

1.42.3

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.15.0

### Shell

/bin/zsh

### Installation Method

npm install -g get-shit-done-cc

### What happened?

PR #1279 (merged 2026-03-21, closed #1278) added strategy-branch creation to
cmdCommit() in get-shit-done/bin/lib/commands.cjs so pre-execution workflows
(/gsd:discuss-phase, /gsd:plan-phase, etc.) would create the configured
phase/milestone branch before their first commit, instead of landing those
commits on main.

That fix only patched the CJS path. The SDK port at sdk/dist/query/commit.js —
which is what `gsd-sdk query commit` actually dispatches to in production —
has no branching logic at all. So the bug fully reproduces on the latest
release with default config: any pre-execution commit (artifacts from
discuss-phase, plan-phase, etc.) lands on whatever branch the user happens to
be on (typically main) rather than on the configured phase/milestone branch.

Root cause: gsd-sdk's `bin/gsd-sdk.js` shim spawn-execs sdk/dist/cli.js, which
dispatches `query commit` to sdk/dist/query/commit.js#commit. That handler
does parse-flags → sanitize → `git add` → `git diff --cached` → `git commit`,
with no equivalent of commands.cjs:282–314 (the
`if (config.branching_strategy && ...) { ... git checkout -b ... }` block from
#1279). A header comment at sdk/dist/query/commit.js:4 says "Ported from
get-shit-done/bin/lib/commands.cjs (cmdCommit, ...)" — the port predates or
missed #1279.

This is the same SDK↔CJS drift pattern as #3523 (which went the other
direction: SDK got #3116, CJS missed it). The branching-strategy feature is
particularly bitten because two independent fixes have to be kept in sync and
only one half ever lands.

### What did you expect?

With `git.branching_strategy: "phase"` in .planning/config.json, the first
pre-execution commit for phase N should:

1. Detect from the staged file paths that the commit belongs to phase N
2. Look up the phase via findPhaseInternal(cwd, N) → get phase_slug
3. Create or switch to the templated branch
   (default: gsd/phase-{phase}-{slug})
4. Commit on that branch — not on the branch the user happened to start from

This is exactly what the CJS path does at commands.cjs:282–314, courtesy of
PR #1279.

### Steps to reproduce

End-to-end repro through the real /gsd-discuss-phase workflow, on a fresh
project, using GSD defaults — no template overrides.

1. Initialise a project on main:

   mkdir repro && cd repro
   git init -b main
   echo "# repro" > README.md
   git add README.md && git commit -m "init"
   gsd-sdk init "demo project"

2. Enable phase branching with default template:

   .planning/config.json:
   {
     "mode": "yolo",
     "commit_docs": true,
     "git": {
       "branching_strategy": "phase"
     }
   }

3. Add a minimal ROADMAP.md so discuss-phase can resolve the phase:

   .planning/ROADMAP.md (excerpt):
     ## Phases
     - [ ] **Phase 1: Pig Latin CLI** - Toy CLI

     ### Phase 1: Pig Latin CLI
     **Goal**: Build a CLI that converts text to Pig Latin
     **Plans**: TBD

   git add .planning/ && git commit -m "chore: scaffold project"
   git branch         # confirm: only `main` exists

4. Run /gsd-discuss-phase 1 in your runtime (Claude Code, etc.) and answer the
   gray-area questions normally.

5. discuss-phase runs its git_commit step:

   gsd-sdk query commit "docs(01): capture phase context" \
     --files .planning/phases/01-pig-latin-cli/01-CONTEXT.md \
             .planning/phases/01-pig-latin-cli/01-DISCUSSION-LOG.md

   Expected: branch `gsd/phase-01-pig-latin-cli` is created (the default
             phase_branch_template) and the commit lands there.
   Actual:   handler returns `{"committed": true, "hash": "..."}` and the
             commit lands on main. `git branch -a` still shows only `main`.

### Error output / logs

```shell
# discuss-phase's git_commit step (the SDK handler call):
$ gsd-sdk query commit "docs(01): capture phase context" \
    --files .planning/phases/01-pig-latin-cli/01-CONTEXT.md \
            .planning/phases/01-pig-latin-cli/01-DISCUSSION-LOG.md
{
  "committed": true,
  "hash": "124978e",
  "message": "docs(01): capture phase context",
  "files": [
    ".planning/phases/01-pig-latin-cli/01-CONTEXT.md",
    ".planning/phases/01-pig-latin-cli/01-DISCUSSION-LOG.md"
  ]
}

$ git branch --show-current
main                                                  ← still on main

$ git branch -a
* main                                                ← only branch exists

$ git reflog -n 3
124978e HEAD@{0}: commit: docs(01): capture phase context
83f7f06 HEAD@{1}: commit: chore: scaffold project
22bf03e HEAD@{2}: commit: chore: enable branching
```

### GSD Configuration

```json
{
  "mode": "yolo",
  "commit_docs": true,
  "git": {
    "branching_strategy": "phase"
  }
}
```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

### Workaround (if any)

Manually `git checkout -b <expected-branch>` before running discuss-phase /
plan-phase. This works but defeats the purpose of branching_strategy: the
feature exists precisely so the user doesn't have to do that.

### Additional context

Status: fix prepared and tested locally; will PR once labeled `confirmed-bug`,
per CONTRIBUTING.md's issue-first rule.

Related issues (same SDK↔CJS drift surface):
- #1278 — original bug (closed by #1279)
- #1279 — partial fix (CJS only) — this is the fix that never crossed over
- #3055 — earlier nested-config variant
- #3116 — fixed top-level branching_strategy migration in SDK
- #3523 — sister drift: SDK got #3116, CJS didn't
- #3730 — recent execute-phase branching regression

Verification this is the live path, not CJS:

- `gsd-sdk` symlink → `bin/gsd-sdk.js` → `spawn node sdk/dist/cli.js`
- The discuss-phase / plan-phase workflows exclusively call
  `gsd-sdk query commit ...`, never the CJS bin
- `loadConfig` in both paths correctly resolves nested
  `git.branching_strategy` to `"phase"` (not the bug)
- `findPhaseInternal('01')` returns valid phase data (verified)
- The phase-detection regex matches the discuss-phase file path layout
  (verified)
- The CJS `cmdCommit` block at commands.cjs:282–314 would create the branch
  correctly *if it were invoked* — it just isn't, because production goes
  through the SDK

Scope note: `/gsd-execute-phase` is **unaffected** because its orchestrator
does `git checkout -b` directly (visible as
`Switched to a new branch 'gsd/phase-01-pig-latin-cli'` in its output, before
any `gsd-sdk query commit` call). Only the pre-execution commit path
(discuss-phase, plan-phase) — which relies on the SDK commit handler to
create the branch as a side-effect of the first commit — is broken. This is
why the discuss-phase artifacts end up permanently on main: the subsequent
execute-phase branches off main at the polluted HEAD.

### Privacy Checklist

agni21:
## Linked Issue

Fixes #3749

> Issue has the `confirmed-bug` label and a maintainer triage comment confirming the gap and proposing exactly this fix shape.

## What was broken

`gsd-sdk query commit` (the production dispatch path used by `/gsd-discuss-phase`, `/gsd-plan-phase`, etc.) did not honor `git.branching_strategy: phase | milestone` for the *first* pre-execution commit. Planning artifacts landed on whatever branch the user happened to be on (typically `main`), polluting it before `execute-phase` branched off.

## What this fix does

Ports the strategy-branch-before-commit logic from CJS `bin/lib/commands.cjs::cmdCommit` (PR #1279) to the SDK handler at `sdk/src/query/commit.ts`. A new `ensureStrategyBranch()` helper runs after pathspec resolution and before the `git add` loop, so the configured phase/milestone branch is created (or switched to) before staging.

## Root cause

PR #1279 fixed the bug only in the CJS `cmdCommit` path. The SDK port at `sdk/src/query/commit.ts` was created earlier (`feat(sdk): Phase 1 typed query foundation` #2118) and was never updated when #1279 landed. Once `gsd-sdk query commit` became the production path for `/gsd-discuss-phase` and `/gsd-plan-phase`, the bug surfaced again — but only on the SDK path.

This is the same drift class catalogued in ADR-3524 and prior incidents (#1535, #1542, #2047/#2052, #2638/#2655, #2653/#2670, #2687/#2706, #2798/#2816, #3055/#3116, #3523).

### How I verified the fix

1. **Unit/regression** — added 4 tests under `commit branching strategy (#1279, #3749)` in `sdk/src/query/commit.test.ts` covering phase strategy, milestone strategy, decimal phase numbers (`45.14-foo`), and the `branching_strategy=none` no-op. All pass: `npx vitest run src/query/commit.test.ts` → 30/30.
2. **Full root suite** — `npm test` (canonical CI suite, `node:test`) → 6182/6182 pass.
3. **End-to-end repro on a clean project** — scaffolded a minimal repro project with `git.branching_strategy: phase`, ran `gsd-sdk query commit` from `main` with a `.planning/phases/01-pig-latin-cli/CONTEXT.md` file. Without the fix: commit lands on `main`. With the fix: `gsd/phase-01-pig-latin-cli` branch is created and the commit lands there; `main` HEAD unchanged.
4. **`/gsd-discuss-phase` flow** — drove the fix through the actual workflow end-to-end; CONTEXT.md committed onto the strategy branch, `main` clean.

### Regression test added?

- [x] Yes — 4 tests under `'commit branching strategy (#1279, #3749)'` in `sdk/src/query/commit.test.ts`
- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

## Checklist

- [x] Regression test added
- [x] All existing tests pass (`npm test` → 6182/6182)
- [x] `.changeset/` fragment added (`.changeset/clever-otters-rest.md`, type Fixed, pr 3770 — will update if PR number differs)

## Breaking changes

None. The fix only kicks in when `git.branching_strategy` is `phase` or `milestone`. The `none`/unset path is a no-op.

### DIFF
--- a/.changeset/clever-otters-rest.md
+++ b/.changeset/clever-otters-rest.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 3770
+pr: 3771
 ---
 Pre-execution commits via `gsd-sdk query commit` (used by /gsd-discuss-phase, /gsd-plan-phase) now honor `git.branching_strategy`. Ports PR #1279's strategy-branch-before-commit logic from the CJS `cmdCommit` path to the SDK commit handler so planning artifacts land on the configured phase/milestone branch instead of `main`.
--- a/.changeset/clever-otters-rest.md
+++ b/.changeset/clever-otters-rest.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3770
+---
+Pre-execution commits via `gsd-sdk query commit` (used by /gsd-discuss-phase, /gsd-plan-phase) now honor `git.branching_strategy`. Ports PR #1279's strategy-branch-before-commit logic from the CJS `cmdCommit` path to the SDK commit handler so planning artifacts land on the configured phase/milestone branch instead of `main`.
--- a/sdk/src/query/commit.test.ts
+++ b/sdk/src/query/commit.test.ts
@@ -483,3 +483,129 @@ describe('commit --respect-staged (#3522)', () => {
     expect(diff).toContain('// modified in working tree');
   });
 });
+
+// ─── branching strategy (#1279, #3749) ─────────────────────────────────────
+
+describe('commit branching strategy (#1279, #3749)', () => {
+  function currentBranch(cwd: string): string {
+    return execSync('git rev-parse --abbrev-ref HEAD', { cwd, encoding: 'utf-8' }).trim();
+  }
+
+  beforeEach(() => {
+    // Pin the initial branch name so the assertion does not depend on the
+    // local `init.defaultBranch` git config (defaults vary across machines).
+    execSync('git checkout -b main', { cwd: tmpDir, stdio: 'pipe' });
+    execSync('git config user.email "test@test.com"', { cwd: tmpDir, stdio: 'pipe' });
+    execSync('git config user.name "Test User"', { cwd: tmpDir, stdio: 'pipe' });
+  });
+
+  async function seedInitialCommit(): Promise<void> {
+    // Seed an initial commit so HEAD is resolvable when the handler runs
+    // `git rev-parse --abbrev-ref HEAD` before staging.
+    await writeFile(join(tmpDir, 'README.md'), '# repro');
+    execSync('git add README.md', { cwd: tmpDir, stdio: 'pipe' });
+    execSync('git commit -m "init"', { cwd: tmpDir, stdio: 'pipe' });
+  }
+
+  it('creates strategy branch before first commit when branching_strategy is phase', async () => {
+    await seedInitialCommit();
+    const { commit } = await import('./commit.js');
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({
+        commit_docs: true,
+        git: { branching_strategy: 'phase', phase_branch_template: 'gsd/phase-{phase}-{slug}' },
+      }),
+    );
+    await mkdir(join(tmpDir, '.planning', 'phases', '01-setup'), { recursive: true });
+    await writeFile(
+      join(tmpDir, '.planning', 'ROADMAP.md'),
+      '# Roadmap\n\n## Phase 1: Setup\nGoal: Initial setup\n',
+    );
+    await writeFile(join(tmpDir, '.planning', 'phases', '01-setup', '01-CONTEXT.md'), '# Context\n');
+
+    const result = await commit(
+      ['docs(01): add context', '--files', '.planning/phases/01-setup/01-CONTEXT.md'],
+      tmpDir,
+    );
+    expect((result.data as { committed: boolean }).committed).toBe(true);
+    expect(currentBranch(tmpDir)).toBe('gsd/phase-01-setup');
+  });
+
+  it('creates strategy branch before first commit when branching_strategy is milestone', async () => {
+    await seedInitialCommit();
+    const { commit } = await import('./commit.js');
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({
+        commit_docs: true,
+        git: {
+          branching_strategy: 'milestone',
+          milestone_branch_template: 'gsd/{milestone}-{slug}',
+        },
+      }),
+    );
+    // getMilestoneInfo() reads ROADMAP.md for the version and name.
+    await writeFile(
+      join(tmpDir, '.planning', 'ROADMAP.md'),
+      '## v1.0: Initial Release\n\n### Phase 1: Setup\n',
+    );
+    await writeFile(join(tmpDir, '.planning', 'test-context.md'), '# Context\n');
+
+    const result = await commit(
+      ['docs: add context', '--files', '.planning/test-context.md'],
+      tmpDir,
+    );
+    expect((result.data as { committed: boolean }).committed).toBe(true);
+    expect(currentBranch(tmpDir)).toBe('gsd/v1.0-initial-release');
+  });
+
+  it('captures decimal phase numbers in the strategy branch name', async () => {
+    await seedInitialCommit();
+    const { commit } = await import('./commit.js');
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({
+        commit_docs: true,
+        git: { branching_strategy: 'phase', phase_branch_template: 'gsd/phase-{phase}-{slug}' },
+      }),
+    );
+    await mkdir(join(tmpDir, '.planning', 'phases', '45.14-golden-capture'), { recursive: true });
+    await writeFile(
+      join(tmpDir, '.planning', 'ROADMAP.md'),
+      '# Roadmap\n\n## Phase 45.14: Golden Capture\nGoal: Capture golden standard\n',
+    );
+    await writeFile(
+      join(tmpDir, '.planning', 'phases', '45.14-golden-capture', '45.14-CONTEXT.md'),
+      '# Context\n',
+    );
+
+    const result = await commit(
+      [
+        'docs(45.14): add context',
+        '--files',
+        '.planning/phases/45.14-golden-capture/45.14-CONTEXT.md',
+      ],
+      tmpDir,
+    );
+    expect((result.data as { committed: boolean }).committed).toBe(true);
+    expect(currentBranch(tmpDir)).toBe('gsd/phase-45.14-golden-capture');
+  });
+
+  it('does not switch branches when branching_strategy is none', async () => {
+    await seedInitialCommit();
+    const { commit } = await import('./commit.js');
+    await writeFile(
+      join(tmpDir, '.planning', 'config.json'),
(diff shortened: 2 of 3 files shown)


trek-e:
*Generated by combined pr-review-toolkit + codex adversarial review*

## Critical Issues

None.

## Important Issues

**[I1] Phase number extraction regex is too broad and can match non-phase filenames.**
In `ensureStrategyBranch`, the phase number is extracted from the joined file paths with:
```ts
const phaseMatch = files.join(' ').match(/(\d+(?:\.\d+)*)-/);
```
This regex matches the **first** sequence of digits-with-dots followed by a hyphen found anywhere across all file path strings joined with spaces. If a path contains a version number, date, or numbered directory that isn't a phase (e.g., `docs/v1.2-notes.md`, `packages/node18-compat/file.ts`), a wrong phase will be extracted, causing an incorrect branch creation. The CJS path (`cmdCommit`) in `commands.cjs:285-320` likely uses a more specific pattern (e.g., path segment relative to `.planning/phases/`). Using the `.planning/phases/<phase>-` prefix from the path is more reliable.

**[I2] `findPhase` API surface may not match expectations.**
The helper calls `findPhase([phaseNum], projectDir, workstream)` and casts the result data to `{ found: boolean; phase_number: string | null; phase_slug: string | null }`. If `findPhase` returns a different shape (e.g., `phase` instead of `phase_number`, or `slug` instead of `phase_slug`), the branch template substitution silently produces `gsd/phase-null-phase` instead of failing visibly. The test covers the happy path but not a case where `findPhase` returns `found: true` with a null `phase_slug` — the fallback `|| 'phase'` masks it.

**[I3] Tests use `beforeEach` to create a `main` branch on the shared `tmpDir` but `tmpDir` initialization happens in the outer describe scope.**
The `beforeEach` runs `git checkout -b main` on a bare repo that may already have a `main` branch from a prior test in the outer suite. If two tests in `commit branching strategy` both attempt `checkout -b main`, the second will fail with "branch already exists." The `beforeEach` should either use a fresh tmpDir per test or use `git checkout main || git checkout -b main`.

**[I4] The `checkout -b branchName` fallback swallows the error silently.**
When `checkout -b branchName` fails with exit code ≠ 0 (branch already exists), the code does `execGit(projectDir, ['checkout', branchName])` — but if that also fails (e.g., dirty working tree), the error is also swallowed. This means a branch-switch failure leaves the user on the wrong branch with no indication, and the subsequent commit lands in the wrong place. At minimum, log the failure at debug level.

## Suggestions

**[S1] Missing test: `branching_strategy: none` explicitly set.**
The PR notes this as a tested case in the PR body ("and the `branching_strategy=none` no-op"), but it is not visible in the diff. Confirm this test case exists.

**[S2] `getMilestoneInfo` return type assumption.** The code accesses `milestone.version` and `milestone.name` without checking if `getMilestoneInfo` can return `null` vs an object with `version: null`. The current null check `if (!milestone || !milestone.version) return` handles this, but `milestone.name` is used without checking — `(milestone.name || 'milestone')` handles the fallback correctly, which is good.

**[S3] `config.git.phase_branch_template` access without null guard.**  
After `strategy === 'phase'` is confirmed, the code accesses `config.git.phase_branch_template` directly. If `config.git` exists but `phase_branch_template` is undefined (user set `branching_strategy: phase` but omitted the template), this would produce `undefined` as the branch name, which `replace` calls on `undefined` would throw. Add a guard: `if (!config.git?.phase_branch_template) return;`

## Strengths

- Excellent PR body with root-cause tracing, ADR citation, full test matrix, and end-to-end verification.
- The porting approach (CJS → SDK parity) is the right fix for ADR-3524 class drift.
- `ensureStrategyBranch` is correctly placed before staging (not after), so the branch is present for `git add`.
- Error-swallowing design intentionally chosen to prevent config problems from blocking commits — the design rationale is well documented.
- Changeset fragment is present and correctly formatted.
- 4 tests covering phase/milestone/decimal/none strategies.

## Recommended Action

Request changes on I1 (regex fragility), I3 (potential test interference on shared tmpDir), and I4 (silent branch-switch failure). I2 and S3 are worth addressing but lower priority. The core approach is sound and the PR is well-written.

agni21:
Thanks for the review @trek-e — and apologies for the duplicate work. I just realized #3763 was merged ten minutes before your review here, fixing the same bug with a more thorough decomposition (typed-IR helpers, multi-phase rejection, empty-template guard) than my port.

Closing this in favor of the merged fix in #3763. The review feedback (regex breadth I1, branch-switch silent-failure I4, template guard S3) is preserved for posterity should anyone hit edges of the merged implementation later.

Lesson learned on my end: check `gh pr list --search "<issue#>"` for in-flight PRs against the same issue even when the issue is still OPEN.

PR #3786, opened 2026-05-21
fix(phase): case-insensitive depends_on resolution

### LINKED ISSUE #3785: depends_on resolution is case-sensitive; lowercase reference to uppercase-suffix plan ID silently drops DAG edge
### GSD Version

1.42.3 (verified against `main` at 4798391)

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v22.x

### Shell

/bin/zsh

### Installation Method

### What happened?

`phase.plan-index` uses strict-case `Map.has()` lookups when resolving `depends_on` references back to plan IDs in the topological-sort DAG (`sdk/src/query/phase.ts`, around the `planMap` / `canonicalToId` / `shortFormToId` lookups). When a plan declares `depends_on: [05c-01]` (lowercase) and the sibling plan on disk is `05C-01-PLAN.md` (uppercase letter-suffix phase), the lookup misses on all three maps, the edge is silently dropped, and the dependent plan collapses into wave 1. The only signal is a misleading

> `Plan 05C-02: declared wave: 2 but depends_on DAG places it in wave 1`

warning that points at the wave declaration rather than at the case mismatch. The actual problem (a typo'd case) is never named.

### What did you expect?

`depends_on: [05c-01]` should resolve to the sibling plan `05C-01` since plan-id case-folding does not introduce any collisions in practice. The plan-id namespace is `NN`, `NN-NN`, or `NN-NN-slug` with an optional letter suffix on the phase segment; a single phase directory cannot host two plans whose IDs differ only in case (canonical-prefix indexing would collide first), and POSIX filesystems treat them as distinct files anyway.

At minimum, if strict-case behavior is intentional, the warning should explicitly name the case mismatch instead of pointing at the wave declaration.

### Steps to reproduce

1. Create a phase directory `.planning/phases/05C-letter-suffix/`
2. Write `05C-01-PLAN.md` with frontmatter `phase: 05C, plan: 01, wave: 1, depends_on: []`
3. Write `05C-02-PLAN.md` with frontmatter `phase: 05C, plan: 02, wave: 2, depends_on: [05c-01]` (lowercase)
4. Run `gsd-sdk query phase.plan-index 05C` (or equivalent CLI invocation)
5. Observe: `waves` puts `05C-02` in wave 1, and warnings include `declared wave: 2 but depends_on DAG places it in wave 1` instead of an unresolved-reference warning naming the case mismatch.

### Error output / logs

```
warnings:
  - "Plan 05C-02: declared wave: 2 but depends_on DAG places it in wave 1"
waves:
  "1": ["05C-01", "05C-02"]
```

(No `unresolved depends_on reference '05c-01'` warning is emitted, because `unresolvedDeps` is only populated when none of the three lookup maps hit — and they all hit on the uppercase form when the dep is uppercase. The lowercase-on-uppercase mismatch is invisible to the unresolved-reference path.)

### Impact

### Additional context

Real-world repro: plans authored by the `gsd-planner` agent occasionally lowercase letter-suffix phase IDs even though the filenames on disk use uppercase. The doubled warning signature (wave-mismatch + silently-dropped edge) is hard to debug from the user side because nothing names the case mismatch directly.

Fix is one-file (case-insensitive map keys + case-folded lookup, with the canonical-cased plan ID preserved as map value so output still uses on-disk casing). PR forthcoming and references this issue.

jnuyens:
## Linked Issue

Fixes #3785

## What was broken

`phase.plan-index` resolved `depends_on` references with strict-case `Map.has()` lookups, so `depends_on: [05c-01]` did not match a sibling plan whose on-disk ID was `05C-01`. The dropped edge collapsed the dependent plan into wave 1 and surfaced only a misleading "declared wave: N but depends_on DAG places it in wave 1" warning.

## What this fix does

Plan-id lookup maps (`planMap`, `canonicalToId`, `shortFormToId`) now key on the lowercased plan ID, and `dep` is lowercased at lookup time. Resolved deps return the canonical-cased plan ID stored as the map value, so wave output and warning messages still use on-disk casing.

## Root cause

The three lookup maps were keyed off the raw plan ID exactly as derived from the filename, and `Map.has(dep)` is case-sensitive. The `unresolvedDeps` path is only populated when all three maps miss, so a lowercase-on-uppercase mismatch landed silently in the dropped-edge state rather than surfacing as an unresolved-reference warning. Plans authored by the `gsd-planner` agent occasionally lowercase letter-suffix phase IDs even though filenames on disk use uppercase, which is how this surfaced in practice.

### How I verified the fix

New `phase.test.ts` case: a `05C` phase with sibling plans `05C-01-PLAN.md` and `05C-02-PLAN.md` where `05C-02` declares `depends_on: [05c-01]` (lowercase canonical-prefix form). The test asserts wave 1 contains `05C-01`, wave 2 contains `05C-02`, and `warnings` is empty (no unresolved-reference warning, no wave-mismatch warning).

Patch also verified in a downstream Claude Code plugin packaging of GSD (jnuyens/gsd-plugin@v2.43.5) where the same byte-identical fix passes 30/30 in `phase.test.ts`.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] Linked issue has the `confirmed-bug` label (issue #3785 just filed, awaiting maintainer triage)
- [x] Fix is scoped to the reported bug, no unrelated changes included
- [x] Regression test added
- [x] All existing tests pass (`phase.test.ts` 30/30 in downstream verification; upstream npm test not run because the fork was cloned shallow without deps installed)
- [ ] `.changeset/` fragment added if this is a user-facing fix, or `no-changelog` label applied (please advise; I can add either, this is the first time I have contributed here)

## Breaking changes

None. Plan-id namespace is `NN`, `NN-NN`, or `NN-NN-slug` with an optional letter suffix on the phase segment. A single phase directory cannot host two plans whose IDs differ only in case (canonical-prefix indexing would collide first), and POSIX filesystems treat them as distinct files anyway. Case-folding introduces no new collision surface, and the canonical-cased plan ID is preserved in map values so wave output and warning messages remain byte-identical for all previously-resolving cases.

### DIFF
--- a/get-shit-done/bin/lib/phase.cjs
+++ b/get-shit-done/bin/lib/phase.cjs
@@ -446,10 +446,13 @@ function cmdPhasePlanIndex(cwd, phase, raw) {
 
   // Build a map from plan ID → raw plan for fast lookup.
   // Deps that reference plans outside this phase are treated as external and ignored.
-  const planMap = new Map(rawPlans.map(p => [p.id, p]));
+  // Plan-id lookups are case-insensitive: map keys are lowercased and dep
+  // lookups are lowercased at query time so that depends_on: ['05c-01'] resolves
+  // to a plan with on-disk ID '05C-01-foo'.
+  const planMap = new Map(rawPlans.map(p => [p.id.toLowerCase(), p]));
   // Secondary index: canonical prefix → full plan ID, so depends_on: ['03-01'] resolves
   // to '03-01-auth-hardening-PLAN.md'-derived ID '03-01-auth-hardening' (k015).
-  const canonicalToId = new Map(rawPlans.map(p => [extractCanonicalPlanId(p.id), p.id]));
+  const canonicalToId = new Map(rawPlans.map(p => [extractCanonicalPlanId(p.id).toLowerCase(), p.id]));
 
   // Kahn's algorithm — compute in-degree and adjacency for in-phase deps only.
   const level = new Map();
@@ -461,7 +464,9 @@ function cmdPhasePlanIndex(cwd, phase, raw) {
     if (!adj.has(p.id)) adj.set(p.id, []);
     for (const dep of p.dependsOn) {
       // Accept both full-stem ('03-01-auth-hardening') and canonical-prefix ('03-01') forms.
-      const resolvedDep = planMap.has(dep) ? dep : canonicalToId.get(dep);
+      // Lookups are case-insensitive.
+      const depKey = dep.toLowerCase();
+      const resolvedDep = planMap.has(depKey) ? planMap.get(depKey).id : canonicalToId.get(depKey);
       if (!resolvedDep) continue; // external dep — ignore
       if (!adj.has(resolvedDep)) adj.set(resolvedDep, []);
       adj.get(resolvedDep).push(p.id);
--- a/sdk/src/query/phase.test.ts
+++ b/sdk/src/query/phase.test.ts
@@ -648,4 +648,38 @@ describe('phasePlanIndex', () => {
     // A clear, dedicated warning naming the unresolved reference must surface.
     expect(warnings.some(w => /unresolved/i.test(w) && w.includes('does-not-exist'))).toBe(true);
   });
+
+  it('case-insensitive depends_on: lowercase dep resolves to uppercase-suffix plan ID', async () => {
+    const phase05c = join(tmpDir, '.planning', 'phases', '05C-letter-suffix');
+    await mkdir(phase05c, { recursive: true });
+    await writeFile(join(phase05c, '05C-01-PLAN.md'), [
+      '---',
+      'phase: 05C',
+      'plan: 01',
+      'wave: 1',
+      'autonomous: true',
+      'depends_on: []',
+      '---',
+      '<objective>Plan A with uppercase suffix.</objective>',
+    ].join('\n'));
+    await writeFile(join(phase05c, '05C-02-PLAN.md'), [
+      '---',
+      'phase: 05C',
+      'plan: 02',
+      'wave: 2',
+      'autonomous: true',
+      'depends_on: [05c-01]',
+      '---',
+      '<objective>Plan B references plan A via lowercase canonical-prefix form.</objective>',
+    ].join('\n'));
+
+    const result = await phasePlanIndex(['05C'], tmpDir);
+    const data = result.data as Record<string, unknown>;
+    const waves = data.waves as Record<string, string[]>;
+    const warnings = (data.warnings as string[] | undefined) ?? [];
+
+    expect(waves['1']).toEqual(['05C-01']);
+    expect(waves['2']).toEqual(['05C-02']);
+    expect(warnings).toEqual([]);
+  });
 });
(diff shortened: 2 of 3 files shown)


trek-e:
*Generated by combined pr-review-toolkit + codex adversarial review*

## Critical Issues

None.

## Important Issues

**[I1] Missing `.changeset/` fragment — author requests guidance.**
The checklist item `[ ] .changeset/ fragment added if this is a user-facing fix, or no-changelog label applied` is unchecked and the author explicitly asks for guidance. This fix is user-facing (incorrect wave assignment / misleading warnings silently dropped for plans in letter-suffix phases such as `05C`). A `.changeset/` fragment with `type: Fixed` is required per CONTRIBUTING.md L140–L153. Add it or apply the `no-changelog` label if maintainer determines this is infra-only.

**[I2] CJS path (`phase.cjs`) does not update `shortFormToId` to lowercase.**
The SDK path (`phase.ts`) correctly lowercases the `shortForm` key at line `const shortForm = canonical.slice(lastDash + 1).toLowerCase()`. The CJS path (`phase.cjs`) has the analogous change for `planMap` and `canonicalToId` but the diff does not show the `shortFormToId` map being lowercased in the CJS path. If the CJS path also builds `shortFormToId`, verify it also lowercases the short-form key; otherwise a dep specified as `01` could still miss a plan ID resolved via the short-form index on the CJS path.

**[I3] Test only covers the `canonicalToId` lookup path; `shortFormToId` path is untested.**
The single added test uses `depends_on: [05c-01]` — a canonical-prefix form that hits the `canonicalToId` map. A test where the dep is the short two-digit form (e.g., `depends_on: [01]` referencing `05C-01`) exercising the `shortFormToId` path would complete coverage.

## Suggestions

**[S1] Consider a cross-casing collision comment.** The SDK code includes a good justification comment explaining why case-folding collisions are not a practical concern. The CJS path has only a brief inline comment. Parity would improve future maintainability.

**[S2] Test fixture file names.** The test creates `05C-01-PLAN.md` but the frontmatter uses `plan: 01` without a slug. Ensure the ID derived by `extractCanonicalPlanId` from the bare `05C-01` filename is `05C-01` (no slug portion), so the test exercises the right code path.

## Strengths

- Correct design: maps keyed on lowercased IDs, lookup via lowercased key, value returns canonical-cased ID — preserves output fidelity.
- SDK and CJS paths both updated in a single PR, reducing drift risk.
- Test is a regression test that mirrors the exact reported scenario (`gsd-planner` agents lowercasing letter-suffix phase IDs).
- Clear, well-written PR description with root-cause analysis and verified downstream test pass count.

## Recommended Action

Approve contingent on: (1) author adds `.changeset/` fragment or maintainer applies `no-changelog` label; (2) verify `shortFormToId` is also lowercased in the CJS path; (3) optionally add a short-form dep test for completeness. The core logic is correct.

trek-e:

Because this PR came from a fork, the migration couldn't auto-recreate it. A tracking issue has been opened on the active fork:
**https://github.com/GSD-redux/get-shit-done-redux/issues/99**

cc @jnuyens

When you're ready, please rebase your branch against `GSD-redux/get-shit-done-redux` and open a fresh PR. Link the tracking issue.

This PR is being auto-closed and locked.

PR #3497, opened 2026-05-14
fix: deterministic changelog extraction for /gsd:update

DansiDanutz:
## Fix PR

> **Using the wrong template?**
> — Enhancement: use [enhancement.md](?template=enhancement.md)
> — Feature: use [feature.md](?template=feature.md)

---

## Linked Issue

> **Required.** This PR will be auto-closed if no valid issue link is found.

Fixes #3496

> The linked issue must have the `confirmed-bug` label. If it doesn't, ask a maintainer to confirm the bug before continuing.

---

## What was broken

`/gsd:update` changelog extraction on the current line was not deterministic and could miss intermediate release sections between the installed version and the target/latest version.

## What this fix does

Adds a dedicated range-aware helper and routes `/gsd:update` through it so changelog extraction consistently returns every matching section from `(from, to]`.

## Root cause

The update workflow relied on vague/manual changelog parsing instead of a deterministic version-range extractor with explicit boundary semantics.

### How I verified the fix

- Added a regression test first in `tests/extract-changelog.test.cjs`
- Confirmed it failed before implementation because the helper did not exist
- Implemented `bin/extract-changelog.cjs`
- Re-ran `node --test tests/extract-changelog.test.cjs` and it passed
- Manually verified helper behavior against the repo `CHANGELOG.md` for:
  - normal extraction range
  - `--json` mode
  - `v`-prefixed version args
  - no-match behavior returning exit code `2`
- Ran `npm test -- --grep extract-changelog`; the suite still hits unrelated pre-existing failures in `tests/code-review.test.cjs` and `tests/skill-manifest.test.cjs`

### Regression test added?

- [x] Yes — added a test that would have caught this bug
- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Windows (including backslash path handling)
- [ ] Linux
- [ ] N/A (not platform-specific)

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Node.js helper + workflow path on macOS
- [ ] N/A (not runtime-specific)

---

## Checklist

- [x] Issue linked above with `Fixes #NNN` — **PR will be auto-closed if missing**
- [ ] Linked issue has the `confirmed-bug` label
- [x] Fix is scoped to the reported bug — no unrelated changes included
- [x] Regression test added (or explained why not)
- [ ] All existing tests pass (`npm test`)
[...]

PR #3495, opened 2026-05-14
fix: deterministic changelog extraction for /gsd:update

DansiDanutz:
## Fix PR

> **Using the wrong template?**
> — Enhancement: use [enhancement.md](?template=enhancement.md)
> — Feature: use [feature.md](?template=feature.md)

---

## Linked Issue

> **Required.** This PR will be auto-closed if no valid issue link is found.

Fixes #3496

> The linked issue must have the `confirmed-bug` label. If it doesn't, ask a maintainer to confirm the bug before continuing.

---

## What was broken

`/gsd:update` changelog extraction on the current line was not deterministic and could miss intermediate release sections between the installed version and the target/latest version.

## What this fix does

Adds a dedicated range-aware helper and routes `/gsd:update` through it so changelog extraction consistently returns every matching section from `(from, to]`.

## Root cause

The update workflow relied on vague/manual changelog parsing instead of a deterministic version-range extractor with explicit boundary semantics.

### How I verified the fix

- Added a regression test first in `tests/extract-changelog.test.cjs`
- Confirmed it failed before implementation because the helper did not exist
- Implemented `bin/extract-changelog.cjs`
- Re-ran `node --test tests/extract-changelog.test.cjs` and it passed
- Manually verified helper behavior against the repo `CHANGELOG.md` for:
  - normal extraction range
  - `--json` mode
  - `v`-prefixed version args
  - no-match behavior returning exit code `2`
- Ran `npm test -- --grep extract-changelog`; the suite still hits unrelated pre-existing failures in `tests/code-review.test.cjs` and `tests/skill-manifest.test.cjs`

### Regression test added?

- [x] Yes — added a test that would have caught this bug
- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Windows (including backslash path handling)
- [ ] Linux
- [ ] N/A (not platform-specific)

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Node.js helper + workflow path on macOS
- [ ] N/A (not runtime-specific)

---

## Checklist

- [x] Issue linked above with `Fixes #NNN` — **PR will be auto-closed if missing**
- [ ] Linked issue has the `confirmed-bug` label
- [x] Fix is scoped to the reported bug — no unrelated changes included
- [x] Regression test added (or explained why not)
- [ ] All existing tests pass (`npm test`)
[...]

PR #3796, opened 2026-05-21
fix(3496): include all version patterns in changelog extraction

trek-e:
## Fix PR

---

## Linked Issue

Fixes #3496

> The linked issue has the `confirmed-bug` label (verified).

---

## What was broken

`parseChangelog` in `scripts/changeset/serialize.cjs` used a single-line regex to match changelog bullets. Bullets where the `(#NNNN)` PR trailer appeared on a **continuation line** (indented with two or more spaces) were silently dropped. This caused Feature/Enhancement sections with multi-line bullets to return 0 entries, making `/gsd:update` miss intermediate versions.

Additionally, `/gsd:update` had no dedicated range-aware CLI helper — extraction was vague/manual with no deterministic `(from, to]` boundary semantics.

## What this fix does

1. **`scripts/changeset/serialize.cjs`** — `parseChangelog` now folds continuation lines (lines starting with `≥2 spaces`) into the preceding bullet before applying the PR-trailer regex. Multi-line bullets are captured correctly regardless of which line carries the `(#NNNN)` trailer.

2. **`scripts/changeset/cli.cjs`** — adds an `extract` subcommand: `changeset/cli.cjs extract --from VERSION --to VERSION [--changelog FILE] [--json]`. Extracts releases strictly after `--from` (exclusive) and up to and including `--to` (inclusive). Accepts `v`-prefixed versions. Exits `2` when no releases fall in range, giving `/gsd:update` a deterministic range-aware helper.

## Root cause

The `parseChangelog` bullet-matching regex was `^- .*\(#(\d+)\)` applied line-by-line, which only matched bullets where the PR trailer was on the opening `- ` line. Multi-line bullets — common in Feature and Enhancement sections — had the trailer on a continuation line that was never joined before matching.

## Prior PRs on this issue

- **PR #3495** (closed, 2 checks failed): Added a standalone `bin/extract-changelog.cjs` script. Approach placed the extractor outside the existing changeset toolchain and had CI failures.
- **PR #3497** (closed, all checks passed but closed without merge): Same standalone script approach, refined. Closed without merge — likely because the extractor belongs integrated into the existing `scripts/changeset/cli.cjs` tool rather than as a separate script, and because `parseChangelog` itself was not fixed to handle multi-line bullets.

This PR differs: it fixes `parseChangelog` at the source (multi-line bullet folding) and integrates the `extract` subcommand into the existing `scripts/changeset/cli.cjs` rather than adding a new top-level script.

### How I verified the fix

- `tests/changeset-serialize.test.cjs` — new tests covering multi-line bullet capture via `parseChangelog` directly; includes the regression fixture from the issue (bullet with PR trailer on continuation line)
- `tests/changeset-cli.test.cjs` — new tests for the `extract` subcommand: range boundaries, `v`-prefix handling, multi-line bullet capture in extracted range, markdown vs `--json` mode, missing-args error path
- The `emits-markdown` test uses `parseChangelog` on the output (not raw substring matching) to assert version presence

### Regression test added?

### Platforms tested

- [x] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] Issue linked above with `Fixes #3496` — **PR will be auto-closed if missing**
- [x] All existing tests pass (`npm test`) — Mac: 3446 passed, 0 failed; Docker: 10344 passed (2 locking-bugs failures fixed by barrier redesign in follow-up commit, same pattern as #3790)
- [ ] `.changeset/` fragment added if this is a user-facing fix — added after PR number is known

## Breaking changes

None

---

Reviewer findings from consolidated code review applied in commit `468dd3ae` + follow-up commits. All 15 targeted tests pass; lint: 0 violations.

### BLOCKER fixes applied

- **F1** (`get-shit-done/workflows/update.md`): `show_changes_and_confirm` step now contains a concrete `node "$GSD_DIR/get-shit-done/scripts/changeset/cli.cjs" extract --from $INSTALLED_VERSION --to $LATEST_VERSION --changelog $CHANGELOG_TMP --json` invocation with explicit exit-2 handling ("No changelog updates between vX and vY") and a fallback message. The prior prose never called the binary and was the root cause of the user-visible bug not being fixed.

- **F2** (`scripts/changeset/cli.cjs`): `releases.filter` now rejects any `rel.version` that fails `SEMVER_RE` before numeric-tuple comparison. `parseSemver('1.0.0-rc.1')` previously returned `[1,0,0]` (same as `'1.0.0'`), corrupting range queries. **Architectural choice**: skip pre-release + 4-part versions with a `stderr` warning; full semver §11 pre-release ordering deferred to a consolidation issue (see F8 below).

### MAJOR fixes applied

- **F3** (`scripts/changeset/serialize.cjs`): `releaseMatch` regex updated to `/^##\s+\[([^\]]+)\](?:\([^)]*\))?\s*(?:-\s*(\S+))?/` so linked-header format `## [1.42.1](url) - 2026-05-15` correctly captures the date.

- **F4** (`scripts/changeset/serialize.cjs`): continuation-line test now checks `!/^\s+-\s/` so `  - nested item` terminates the current bullet instead of folding into the parent body.

- **F5** (`scripts/changeset/cli.cjs`): 4-part versions (e.g. `1.0.0.1`) fail `SEMVER_RE` and are skipped by the same guard added for F2. No separate code path needed.

- **F6** (`scripts/changeset/serialize.cjs`): v-prefix stripped from in-file version capture; `## [v1.0.0]` now parses as version `"1.0.0"`.

- **F7** (`scripts/changeset/serialize.cjs`): continuation-line indentation relaxed from `/^[ \t]{2}/` to `/^\s+/` so 1-space-indented continuations fold correctly. F4's bullet-terminator guard prevents nested bullets from being folded.

- **F8** (noted, not applied): Semver consolidation — 6th distinct comparator added by this PR. **Follow-up issue should be filed** to unify all call sites. Out of scope for this PR; acknowledged in commit message.

### MINOR fixes applied

- **F9** (`scripts/changeset/cli.cjs`): unknown-command path now exits `1` instead of `2` (exit `2` is reserved for "no releases in range").

- **F10** (`scripts/changeset/cli.cjs`): text-mode exit-2 path writes `"no releases found in range (from=X, to=Y)"` to stderr.

- **F11** (`tests/changeset-cli.test.cjs`): exit-2 test now asserts `r.json` is present and `r.json.releases.length === 0`.

### Bonus fix (discovered during test run)

- `tests/locking-bugs-1909-1916-1925-1927.test.cjs` `#1927` concurrent config-set test redesigned with the same deterministic file-barrier pattern used in #3790 (locking-bugs:180 and :235). The `Promise.all([execAsync(A), execAsync(B)])` form was non-deterministic under Docker overlay-fs load; barrier ensures true concurrent contention.

### Test results after review commits

- **Mac:** 3446 passed, 0 failed
- **Docker:** 10344 passed, 0 failed (locking-bugs barrier fix covers the 2 that failed in the pre-review Docker run)

### DIFF
--- a/.changeset/noble-tigers-parade.md
+++ b/.changeset/noble-tigers-parade.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3796
+---
+**`parseChangelog` now captures multi-line bullets** — bullets where the `(#NNNN)` PR trailer appears on a continuation line are no longer silently dropped. Adds `changeset/cli.cjs extract --from VERSION --to VERSION` for deterministic range-aware changelog extraction in `/gsd:update`.
--- a/get-shit-done/workflows/update.md
+++ b/get-shit-done/workflows/update.md
@@ -372,28 +372,48 @@ Exit.
 <step name="show_changes_and_confirm">
 **If update available**, fetch and show what's new BEFORE updating:
 
-1. Fetch changelog from GitHub raw URL
-2. Extract entries between installed and latest versions
-3. Display preview and ask for confirmation:
+1. Fetch changelog from GitHub raw URL and save to a temp file, e.g. `/tmp/gsd-changelog-$$.md`.
+2. Extract entries between installed and latest versions using the deterministic range helper (fix for #3496 — do NOT use ad-hoc grep/awk extraction which silently skips intermediate versions):
+
+```bash
+CHANGELOG_TMP="/tmp/gsd-changelog-$$.md"
+curl -fsSL "https://raw.githubusercontent.com/gsd-build/get-shit-done/main/CHANGELOG.md" -o "$CHANGELOG_TMP" 2>/dev/null \
+  || wget -qO "$CHANGELOG_TMP" "https://raw.githubusercontent.com/gsd-build/get-shit-done/main/CHANGELOG.md" 2>/dev/null
+
+EXTRACT_JSON=$(node "$GSD_DIR/get-shit-done/scripts/changeset/cli.cjs" extract \
+  --from "$INSTALLED_VERSION" \
+  --to "$LATEST_VERSION" \
+  --changelog "$CHANGELOG_TMP" \
+  --json 2>/dev/null)
+EXTRACT_EXIT=$?
+rm -f "$CHANGELOG_TMP"
+
+if [ "$EXTRACT_EXIT" -eq 2 ]; then
+  # Exit 2 = no releases in range (e.g. versions are equal or changelog is sparse)
+  CHANGELOG_PREVIEW="No changelog updates between v${INSTALLED_VERSION} and v${LATEST_VERSION}."
+elif [ "$EXTRACT_EXIT" -ne 0 ] || [ -z "$EXTRACT_JSON" ]; then
+  CHANGELOG_PREVIEW="(Could not extract changelog — update will still proceed)"
+else
+  # Re-run without --json to get the human-readable markdown for display
+  CHANGELOG_PREVIEW=$(node "$GSD_DIR/get-shit-done/scripts/changeset/cli.cjs" extract \
+    --from "$INSTALLED_VERSION" \
+    --to "$LATEST_VERSION" \
+    --changelog "$CHANGELOG_TMP" 2>/dev/null || echo "(changelog unavailable)")
+fi
+```
+
+3. Display preview and ask for confirmation, using `$CHANGELOG_PREVIEW` from the extract step above:
 
 ```
 ## GSD Update Available
 
-**Installed:** 1.5.10
-**Latest:** 1.5.15
+**Installed:** {INSTALLED_VERSION}
+**Latest:** {LATEST_VERSION}
 
 ### What's New
 ────────────────────────────────────────────────────────────
 
-## [1.5.15] - 2026-01-20
-
-### Added
-- Feature X
-
-## [1.5.14] - 2026-01-18
-
-### Fixed
-- Bug fix Y
+{CHANGELOG_PREVIEW}
 
 ────────────────────────────────────────────────────────────
 
--- a/scripts/changeset/cli.cjs
+++ b/scripts/changeset/cli.cjs
@@ -255,6 +255,14 @@ function cmdExtract(opts) {
 
   const matched = releases.filter((rel) => {
     if (rel.version === 'Unreleased') return false;
+    // Skip pre-release entries (e.g. 1.0.0-rc.1, 1.0.0-beta.2) — they cannot
+    // be range-compared via numeric tuples without implementing full pre-release
+    // ordering (semver §11).  Exclude them silently for now; a consolidation
+    // issue (#F8) will address pre-release ordering across all comparators.
+    if (!SEMVER_RE.test(rel.version)) {
+      process.stderr.write(`[extract] skipping pre-release/non-semver entry: ${rel.version}\n`);
+      return false;
+    }
     // from is exclusive: cmp > 0 means rel.version > from
     const afterFrom = semverCmp(rel.version, from) > 0;
     // to is inclusive: cmp <= 0 means rel.version <= to
@@ -350,7 +358,7 @@ function main() {
   const { opts } = parsed;
   if (opts.cmd !== 'render' && opts.cmd !== 'github-release-notes' && opts.cmd !== 'extract') {
     process.stderr.write(usage());
-    process.exit(2);
+    process.exit(1);
   }
   if (opts.cmd === 'render' && (!opts.version || !opts.date)) {
     process.stderr.write('--version and --date are required for render\n');
@@ -372,6 +380,8 @@ function main() {
       process.stdout.write(JSON.stringify(report, null, 2) + '\n');
     } else if (textOutput) {
       process.stdout.write(textOutput + '\n');
+    } else if (exitCode === 2) {
+      process.stderr.write(`no releases found in range (from=${report.from}, to=${report.to})\n`);
     }
     process.exit(exitCode);
   }
--- a/scripts/changeset/cli.cjs
+++ b/scripts/changeset/cli.cjs
@@ -19,7 +19,7 @@ const path = require('node:path');
 
 const { parseFragment, FRAGMENT_ERROR } = require('./parse.cjs');
 const { renderChangelog } = require('./render.cjs');
-const { serializeChangelog } = require('./serialize.cjs');
+const { serializeChangelog, parseChangelog } = require('./serialize.cjs');
 const { renderGithubReleaseNotes } = require('./github-release-notes.cjs');
 
 function parseArgs(argv) {
@@ -30,6 +30,7 @@ function parseArgs(argv) {
     date: null,
     fromRef: null,
     toRef: null,
+    changelog: null,
     output: null,
     repoSlug: 'gsd-build/get-shit-done',
     installCommand: 'npx get-shit-done-cc@latest',
@@ -59,6 +60,7 @@ function parseArgs(argv) {
       a === '--date' ||
       a === '--from' ||
       a === '--to' ||
+      a === '--changelog' ||
       a === '--output' ||
       a === '--repo-slug' ||
       a === '--install-command'
@@ -70,6 +72,7 @@ function parseArgs(argv) {
       else if (a === '--date') opts.date = r.value;
       else if (a === '--from') opts.fromRef = r.value;
       else if (a === '--to') opts.toRef = r.value;
+      else if (a === '--changelog') opts.changelog = r.value;
       else if (a === '--output') opts.output = r.value;
       else if (a === '--repo-slug') opts.repoSlug = r.value;
       else if (a === '--install-command') opts.installCommand = r.value;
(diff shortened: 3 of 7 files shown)


trek-e:

## Codex Adversarial Review

Verdict: needs-attention

No-ship: the new changelog `extract` path can silently produce incorrect release summaries by accepting malformed version bounds and by dropping legitimate bullets that lack a trailing PR token.

Findings:
- [high] `extract` accepts malformed version bounds and coerces them into valid ranges (scripts/changeset/cli.cjs:226-244)
  `cmdExtract` treats `--from/--to` as semver but never validates them. `parseSemver` does `split('.') -> Number -> [part || 0]`, so invalid tokens are silently coerced (e.g. `1.41.x` becomes `1.41.0`). I reproduced this with `node scripts/changeset/cli.cjs extract --from v1.41.x --to v1.42.1 --changelog CHANGELOG.md --json`, which exited 0 and returned release data instead of rejecting input. This makes range selection non-deterministic under typos/version-shape drift and can misstate what changed between releases.
  Recommendation: Validate `from`/`to` as strict semver (or parse via a semver library) before comparison; if invalid, return non-zero with a structured error instead of coercing to zero components.
- [high] `extract` still drops changelog bullets that do not end with `(#NNNN)` (scripts/changeset/serialize.cjs:64-73)
  The parser only records bullets when `flushBullet()` matches `^(.*?)\s*\(#(\d+)\)\s*$`; otherwise the bullet is discarded. `cmdExtract` then rebuilds output exclusively from parsed bullets, so dropped entries disappear from extracted notes. On current `CHANGELOG.md`, extracting `1.41.0` yields fewer entries than authored content (`raw` bullet starts: 85, parsed bullets: 78, missing: 7), so users can miss material release notes during update confirmation.
  Recommendation: Preserve bullets even when no PR trailer is present (e.g., store `{ body, pr: null }`), and have `extract` render those bullets instead of dropping them.

Next steps:
- Add strict semver validation and failing tests for malformed `--from/--to` inputs (e.g., `1.41.x`, `foo`, uppercase `V` prefix).
- Extend `parseChangelog` to keep non-PR bullets, then add regression tests proving extracted output preserves full release sections.

## Actions Taken

- `5d4d01aa` — fix(3496): reject malformed --from/--to semver in extract with structured error — adds strict `N.N.N` regex gate before comparison; exits 1 with JSON `{ error }` report on malformed input.
- `b0aefa54` — fix(3496): preserve bullets without PR trailer in parseChangelog (pr: null) — `flushBullet()` now stores `{ body, pr: null }` instead of silently discarding; `cmdExtract` textOutput renders them as `- body` (no trailer). Verified: real CHANGELOG v1.41.0 extraction recovers all 7 previously-dropped bullets.
- New tests added: malformed `--from` (`1.41.x`), malformed `--to` (`foo`), PR-less bullet preservation in extract JSON output, `parseChangelog` PR-less bullet unit test. Test count: 13 → 17 in affected suites.

CI failures resolved: none required (CI was 15/15 green before this review — the "13 pre-existing failures" claim from the prior agent was not reproducible; all checks passed on the latest push).

CI failures verified orthogonal: N/A — all checks were passing at review start.

Rebased onto main: N (branch was already at HEAD of main with 0 commits behind).

## Open for Maintainer

Both findings were bugs introduced by the new `cmdExtract` code added in this PR:
1. The semver coercion is now a hard validation error (exit 1 + JSON `error` field). Schema: `{ error: string, releases: [] }`. Callers in `/gsd:update` will need to handle the new exit-1 path if they pass non-semver bounds.
2. The `pr: null` schema change is backward-compatible for callers that use `b.pr` as a PR link (they were already getting `undefined`/0 from a dropped bullet). Callers that render bullets must now guard `b.pr !== null` before appending `(#N)` — the `textOutput` renderer in `cmdExtract` already does this.

trek-e:

Two independent reviewers ran in parallel: `gsd-code-reviewer` (project-aware) and sonnet adversarial (substituting for `codex-rescue`, fire-and-forget broken this session). **This PR is the most-findings PR in the current sweep — 2 BLOCKERs + 6 MAJORs.** Both reviewers reproduced their critical findings empirically.

### BLOCKER

- **The PR adds the primitive but never wires it into `/gsd:update`** (gsd-code-reviewer). `scripts/changeset/cli.cjs` gains a correct `extract` subcommand and `scripts/changeset/serialize.cjs` fixes the multi-line bullet parser, but **`get-shit-done/workflows/update.md` is unchanged** — still describes vague "extract entries between installed and latest versions" with no `cli.cjs extract` invocation. Issue #3496's stated guarantee is about the user-visible update FLOW, not the existence of a primitive. **The user-visible bug is technically still un-fixed.** Remediation: update `workflows/update.md` to invoke `node scripts/changeset/cli.cjs extract --from <installed> --to <latest> --json` and consume the result in the changelog-display step.

- **`parseSemver` silently maps pre-release versions to their base release** (sonnet adversarial, empirically verified). `parseSemver('1.0.0-rc.1')` splits on `.` → `['1', '0', '0-rc', '1']`. `Number('0-rc')` is `NaN`, `NaN || 0` evaluates to `0`. Result: `parseSemver('1.0.0-rc.1')` returns `[1, 0, 0]` — identical to `parseSemver('1.0.0')`. `semverCmp('1.0.0-rc.1', '1.0.0') === 0`. Consequence: `--from 0.9.9 --to 1.0.0` includes a `1.0.0-rc.1` CHANGELOG entry because `upToTo = (0 <= 0) = true`. Per semver, `1.0.0-rc.1 < 1.0.0` — the pre-release should NOT be included. Remediation: in `releases.filter`, skip (return `false`) any `rel.version` that doesn't pass `SEMVER_RE`, since those entries can't be meaningfully compared by the numeric-tuple comparator.

### MAJOR

- **4-part headers (`1.0.0.1`) silently truncated to `[1,0,0]`** (gsd-code-reviewer M-01). Same silent-coercion anti-pattern the PR rejects at the CLI boundary — internal parser still permissive. Remediation: anchor the SEMVER_RE check at the parser level too.

- **In-file v-prefix (`## [v1.0.0]`) silently excludes the release** (gsd-code-reviewer M-02). `stripV` is only applied to CLI args, not in-file headers, so a CHANGELOG entry written as `## [v1.0.0]` never appears in the `releases` array. Remediation: apply `stripV` (or a tolerant prefix-strip) on the in-file capture too.

- **Three divergent pre-release semver policies across cli.cjs, gsd-check-update-worker.js, gsd-statusline.js** (gsd-code-reviewer M-03). The PR adds a 6th distinct comparator in a codebase that already has 5. Remediation: extract a shared semver-comparator helper into `scripts/changeset/` and migrate the other call sites in a follow-up issue.

- **2-space-only continuation cutoff silently drops 1-space-indented continuation content** (gsd-code-reviewer M-04). The continuation-line test `/^[ \t]{2}/` rejects single-space-indented continuations that real CHANGELOG entries may contain. Remediation: relax to `/^\s+/` and rely on the bullet-terminator to flush.

- **Linked-header releases lose their date in textOutput** (sonnet adversarial, empirically verified against live CHANGELOG.md). The regex `## [1.42.1](url) - date` doesn't skip the URL before matching the date. Every release from 1.42.1 down to 1.41.0 returns `date: null`. Remediation: update the regex to `(?:\([^)]*\))?\s*(?:-\s*(\S+))?` after the closing `]`.

- **45 nested sub-bullets in live CHANGELOG silently folded into parent** (sonnet adversarial). Parser treats nested `  - ` lines as continuation, garbling the body string. Remediation: treat `/^\s+-\s/` as a bullet-terminator, not a continuation.

### MINOR

- Exit-code inconsistency: missing `--from`/`--to` exits `1`, unknown command exits `2` — but `2` is also the "no releases" code (sonnet adversarial). Consumers can't distinguish bad-invocation from no-results. Remediation: reserve `2` solely for "range matched nothing"; use `1` for argument/I-O errors.

- `textOutput === null` on `exitCode: 2` silently emits nothing without `--json` (sonnet adversarial). Remediation: write a brief stderr diagnostic before exiting 2 in text mode.

- One loose `notEqual(r.status, 0)` assertion; missing length-bound on `includes` checks (gsd-code-reviewer test-rigor).

- 3 more MINORs from gsd-code-reviewer (see findings file). Add stricter tests for the BLOCKER and MAJORs above.

### NIT

- Test `'exits 2 with no output when no versions fall in range'` only checks exit, not that JSON IS emitted (sonnet). Rename + tighten assertion to verify `r.json.releases.length === 0`.
- 2 more NITs from gsd-code-reviewer.

### QUESTION

- Is `--from`/`--to` SEMVER_RE intentionally strict to reject `1.0.0-beta.1` as a range bound? The usage docs don't note this constraint.
- 2 more QUESTIONs from gsd-code-reviewer.

### Anti-pattern sweep

The codebase already has **5 distinct semver comparators with 3 different pre-release policies**. This PR adds a 6th. A consolidation follow-up issue is warranted — likely candidate for a separate bug ticket. Not auto-filed because the consolidation requires cross-cutting design discussion.

### Test rigor (positive)

17 new tests added, mostly non-vacuous. The markdown-mode test correctly uses the production `parseChangelog` (not a mock). No source-grep, no pass-always swallowed exceptions, magic-number constants are named.

- Range semantics `(from, exclusive)` `(to, inclusive)` documented and tested for clean numeric versions.
- `--json` output round-trips for the happy path.
- Closing keyword `Fixes #3496` present.

### Recommendation

**Changes requested** before merge. The most critical issue is the **unwired primitive** — the workflows/update.md consumer was never updated to invoke the new `extract` subcommand. Until that wire-up lands, this PR doesn't close #3496. The `parseSemver` pre-release bug is independently a BLOCKER because it would silently include wrong entries in any future range query.

The other MAJORs (linked-header dates, nested bullets, 4-part headers, in-file v-prefix, continuation cutoff) compound: real-world CHANGELOG.md entries hit several of these simultaneously.

---

trek-e:

**https://github.com/GSD-redux/get-shit-done-redux/pull/90**

PR #3792, opened 2026-05-21
fix(3774): treat 999 as exact sentinel in phase-lifecycle-policy

trek-e:
## Linked Issue

Fixes #3774

> The linked issue has the `confirmed-bug` label.

---

## What was broken

`phase.add` returned phase ID `1` (instead of the correct next sequential ID) on projects whose canonical phase numbers are 1000 or higher. For example, a project with phases 1000–1500 would receive phase ID `1` instead of `1501`.

## What this fix does

Changes the sentinel guard in `scanSequentialMaxPhaseFromMilestone` and `scanSequentialMaxPhaseFromDirs` from `num >= 999` to `num === 999`, so only the exact backlog sentinel (`999`) is excluded from the max-scan. Phase IDs 1000+ are valid canonical numbers and are now correctly included.

## Root cause

Both scanner functions in `sdk/src/query/phase-lifecycle-policy.ts` (lines 71 and 88) used `>= 999` to skip the `999.x` backlog lane. This over-broad guard excluded every phase ≥ 1000, causing `computeNextSequentialPhaseId` to see max=0 and return `0 + 1 = 1`.

## Antipattern hunt results

Searched `sdk/src/` for all `>= 999`, `<= 999`, `> 999`, `=== 999` patterns:

- `sdk/src/query/phase-lifecycle-policy.ts:71` and `:88` — the two bug locations, fixed here.
- `sdk/src/query/phase-lifecycle.ts:762`, `:819`, `:832`, `:842` — these use `>= 999` intentionally in **decrement/renumbering** helpers (phase removal). They protect the `999.x` backlog lane from being renumbered when a lower phase is deleted. Using `>= 999` there is correct: no phase ≥ 999 should be renumbered during sequential phase removal. These are NOT bugs and are NOT changed.

### How I verified the fix

Added a regression test to `sdk/src/query/phase-lifecycle.test.ts` (added to existing file — `phaseAdd` describe block):

- Fixture: ROADMAP with `Phase 1000` and `Phase 1500` entries; on-disk dirs `1000-foundation` through `1500-latest`.
- Asserts: `phaseAdd(['Next After 1500'], tmpDir)` returns `phase_number === 1501`.
- Pre-fix: test fails with `expected 1501, received 1`.
- Post-fix: test passes.

The existing test (`skips exactly phase 999 (backlog sentinel) when calculating next number`) provides the counter-test: confirms 999 is still excluded. Together the pair proves the exact-sentinel semantics.

### Regression test added?

[...]

PR #3798, opened 2026-05-21
fix(3785): case-insensitive depends_on resolution in phase resolver

### LINKED ISSUE #3785: depends_on resolution is case-sensitive; lowercase reference to uppercase-suffix plan ID silently drops DAG edge

trek-e:
## Summary

Fixes #3785: depends_on case-sensitive resolution. Lowercase references (e.g. `[05c-01]`) to uppercase-suffix canonical plan IDs (e.g. `05C-01`) now resolve correctly via case-insensitive lookups while preserving canonical on-disk casing in output.

Updates BOTH SDK (TS) AND CJS surfaces atomically — avoids the #3792 cooperating-sibling hand-sync drift pattern.

Dual-pass review (gsd-code-reviewer + sonnet adversarial) surfaced one MAJOR and several MINORs; addressed in commit `dd66a554`:

- **F1 (MAJOR)** — Reworded changeset to disclose CJS-side parity gap: the SDK has three resolution tiers (planMap + canonicalToId + shortFormToId); CJS has only two. Added `// KNOWN GAP:` comment at the CJS lookup chain noting the missing shortFormToId tier. Did NOT backfill the third tier — out of scope for #3785; tracked as a parity follow-up.
- **F2 (MINOR)** — Added strict canonical-casing assertions (`expect(planA.id).toBe('20-01-Auth')` / `assert.strictEqual`) in both SDK and CJS tests. Locks the canonical-preservation invariant; future regression that lowercased plan.id would now fail.
- **F3 (MINOR)** — Normalized output `depends_on` array to canonical IDs (previously preserved the user's uncorrected reference string).
- **F5 (MINOR)** — Added comment explaining `seenLower` guard scope (case-fold collisions only; shared-numeric-prefix collisions resolved by first-write-wins).
- **F6 (NIT)** — Added `// ASCII plan IDs only — toLowerCase() is correct and locale-safe here.` comment at the first `.toLowerCase()` site.

## Test plan

- [x] All existing tests pass
- [x] Mac (macOS): verbatim counts will be added when CI completes
- [x] Linux (Docker): verbatim counts will be added when CI completes
- [ ] Windows (verified via CI)

## Competing PR

PR #3786 (jnuyens) is `review: changes requested` and takes the same approach with smaller surface (no collision guard, no adversarial tests, no changeset). **#3798 supersedes #3786** — recommend closing #3786 with credit to jnuyens on merge.

Fixes #3785


trek-e:

## Codex Adversarial Review

Verdict: needs-attention

No-ship: the new case-insensitive dependency index can silently alias distinct plan IDs and route `depends_on` edges to the wrong plan, which can reorder execution waves without any warning.

Findings:
- [high] Lowercasing plan IDs collapses distinct plans into one dependency key (sdk/src/query/phase.ts:504-550)
  `planMap` is now keyed by `p.id.toLowerCase()` and lookups use `dep.toLowerCase()`. If a phase contains two valid plan files whose IDs differ only by case (inference: possible on case-sensitive filesystems, since canonical plan filtering accepts any `*-PLAN.md`), one entry overwrites the other in `planMap`. A `depends_on` reference then resolves to whichever plan survived the collision, silently rewiring the DAG edge. This can change computed waves/order and hide the mis-resolution because it no longer appears as unresolved.
  Recommendation: Detect and fail on case-insensitive key collisions before building dependency maps (for full IDs and canonical IDs). If collisions exist, return/throw a clear error naming both files. Alternatively, preserve exact-ID matching first, and only do case-insensitive fallback when there is a single unambiguous candidate.

Next steps:
- Add a regression test with two `*-PLAN.md` files that differ only by case and assert that phase indexing errors explicitly instead of silently choosing one.
- Re-run phase ordering tests on a Linux case-sensitive filesystem to confirm deterministic behavior after collision handling.

## Actions Taken

- **fix: collision guard added to `sdk/src/query/phase.ts`** — before building `planMap`/`canonicalToId`, iterates `rawPlans` with a `seenLower` map; throws `GSDError` with both conflicting IDs if two plan IDs lowercase to the same string.
- **fix: lowercase-both applied to `get-shit-done/bin/lib/phase.cjs`** — the CJS `cmdPhasePlanIndex` path was missing the #3785 normalization entirely. Added the same collision guard + `planMap`/`canonicalToId` lowercase-keying + `dep.toLowerCase()` before lookup.
- **test: added to `tests/phase.test.cjs`** — `#3785: depends_on reference with different case resolves to correct plan` (all platforms) and `#3785 adversarial: two plan IDs differing only by case produce a collision error` (skipped on macOS/Windows, runs on Linux CI).
- **test: added to `sdk/src/query/phase.test.ts`** — same two tests for the vitest suite.
- **changeset updated** to describe both SDK and CJS fixes.
- Rebased onto main: N (branch was already at main tip — no rebase needed)
- Windows Node 24 CI failure (`runtime-bridge-sync-smoke.test.cjs`) is pre-existing: all 4 tests in the file pass but the file-level exit is non-zero due to a Node.js test runner anomaly unrelated to this PR. The same pattern occurs on `main`-targeted runs across different jobs/platforms.

Commit: `fix(3785): detect case-fold collisions; apply lowercase-both to CJS path`

## Open for Maintainer

- **Design: only the SDK `phasePlanIndex` got the #3785 fix in the original PR — the CJS CLI path (`cmdPhasePlanIndex` in `phase.cjs`) was missed.** These two implementations are maintained in parallel. A longer-term refactor could consolidate them. This PR now has both paths in sync.
- **Design: collision detection throws on first collision.** If a phase somehow accumulates multiple colliding pairs (unlikely but theoretically possible on a case-sensitive FS with many plans), only the first pair is reported. A full sweep reporting all pairs is a possible enhancement.
- **Design: `shortFormToId` in the SDK path has no collision guard.** The short-form index (`canonical.slice(lastDash + 1).toLowerCase()`) uses "first write wins" semantics with a comment noting it. Two plans with the same short-form suffix (e.g. `03-01-foo-PLAN.md` and `03-01-bar-PLAN.md` both yielding short-form `01`) silently silences the second entry. This is a separate, pre-existing design question and not introduced by #3785.

trek-e:

Two independent reviewers ran in parallel: `gsd-code-reviewer` (project-aware) and sonnet adversarial (substitute for `codex-rescue`, which is fire-and-forget broken this session).

**Positive framing**: this PR notably AVOIDS the #3792 cooperating-sibling hand-sync drift pattern by updating BOTH the SDK (TS) AND CJS surfaces atomically. The PR is well-scoped, mergeable, and supersedes the competing PR #3786 (jnuyens, currently `review: changes requested`). A follow-up commit on this branch will address the MAJOR (a parity overclaim in the changeset/body) and the actionable MINORs.

### MAJOR

- **CJS path is MISSING the `shortFormToId` third-tier resolver entirely** (sonnet adversarial). `grep -n "shortFormToId" phase.cjs` → 0 matches. The TS path has all three tiers (`planMap`, `canonicalToId`, `shortFormToId`); the CJS path only has two. The diff adds `.toLowerCase()` to `canonicalToId` in CJS but the entire `shortFormToId` block — which the PR's own changeset/body claim was fixed — was never present in the CJS copy. **Any `depends_on` reference that is a same-phase short-form (`'01'` or `'01A'`) silently drops the edge in the CJS path** regardless of casing. This is a PRE-EXISTING gap (not introduced here), but the PR's changeset states "Both the SDK query path and the CJS CLI path now normalize identifiers to lowercase on write and lowercase dep strings before lookup" — that overclaims parity. **Remediation**: add a `// KNOWN GAP: CJS resolver has only two tiers (no short-form lookup) — see #3819 / followup` comment at the point in `phase.cjs` where the third tier would live, and reword the changeset body to "SDK and CJS now normalize the two-tier resolution chain present in CJS" so the gap is disclosed rather than hidden.

### MINOR

- **`seenLower` collision guard covers full-stem collisions only** (sonnet adversarial). Plans `20-01-Auth-PLAN.md` (id=`'20-01-Auth'`) and `20-01-PLAN.md` (id=`'20-01'`) produce distinct `seenLower` keys but BOTH produce `extractCanonicalPlanId(id) = '20-01'`, so `canonicalToId.set('20-01', ...)` fires twice and the first entry is silently overwritten. **Remediation**: add a comment explaining the guard's scope, OR widen the collision detection to cover shared-numeric-prefix collisions.

- **Output `depends_on` field propagates the user's uncorrected reference string** (sonnet adversarial). `sdk/src/query/phase.ts:666` writes `depends_on: raw.dependsOn` which is the raw user-typed array. If user wrote `[20-01-auth]` and canonical is `20-01-Auth`, the output plan record carries `depends_on: ['20-01-auth']` (lowercase typo preserved). The plan ID field DOES preserve canonical casing — but the depends_on array doesn't. Downstream consumers iterating depends_on and matching against plan.id values from the same result hit a case mismatch again. **Remediation**: normalize `raw.dependsOn` to canonical IDs on output (replace each resolved dep with the resolved canonical ID).

- **Tests don't assert exact canonical casing on `planA.id`** (sonnet adversarial). `sdk/src/query/phase.test.ts:689` uses `plans.find(p => (p.id as string).startsWith('20-01'))` — would still pass if a future change accidentally lowercased `p.id` before storing in `rawPlans`. The wave-ordering assertions would pass while the canonical-casing guarantee silently broke. **Remediation**: add `expect(planA!.id).toBe('20-01-Auth')` in the SDK test and the CJS equivalent.

- **Changeset body leads with "filenames contain uppercase characters"** but the actual trigger is planner-side ID casing drift, not filename casing (gsd-code-reviewer MN-01). **Remediation**: rewrite to "depends_on references that case-fold differently from canonical plan IDs (e.g. from gsd-planner output that occasionally lowercases IDs) now resolve correctly via case-insensitive lookups while preserving canonical on-disk casing in output."

- **Pre-existing CJS parity gap with TS `shortFormToId` tier** (gsd-code-reviewer MN-02 + sonnet MAJOR overlap). Already documented in the MAJOR above — same root cause.

### NIT

- `seenLower` collision map is built and discarded — could fold into `planMap` construction (gsd-code-reviewer NT-01, cosmetic).
- CJS `return;` after `error()` is dead but idiomatic (gsd-code-reviewer NT-02).
- `toLowerCase()` is Unicode-naive but correct for ASCII plan IDs — undocumented assumption (sonnet adversarial). Add `// ASCII plan IDs only — toLowerCase() is correct and locale-safe here.`
- Competing PR #3786 (jnuyens) is strict subset of #3798 — close #3786 when this lands, credit jnuyens for independent repro.

### QUESTION (worth discussing but not blocking)

- Collision error names IDs, not filenames; the actionable artifact is the file (gsd-code-reviewer QN-01).
- Should `gsd-planner` be patched to emit canonical-cased depends_on refs so this fix is defense-in-depth rather than load-bearing? (gsd-code-reviewer QN-02)
- No case-mismatch warning emitted when a dep resolves only after case-folding — intentional silent-tolerance or worth a warning? (sonnet adversarial). Optional follow-on enhancement.

- **Both SDK (TS) and CJS surfaces updated atomically** — avoids the #3792 hand-sync drift pattern. ✓
- Collision detection on case-fold dimension catches misconfigured phase directories. ✓
- Adversarial Linux-only tests beyond minimum required. ✓
- Cycle-detection unchanged; previously-hidden edges now restored may surface new cycles but error path handles cleanly. ✓
- Closing keyword `Fixes #3785` present. ✓
- Mergeable; CI checks ran. ✓

### Recommendation on competing PR #3786

**#3798 strictly supersedes #3786** (sonnet adversarial NIT + gsd-code-reviewer agreement). Same lowercase-both approach; #3798 adds collision guard, adversarial tests, completed checklist, Docker/Linux verification, changeset. **When #3798 merges, close #3786 with credit to jnuyens for independent reproduction.**

### Recommendation

**Changes requested** before merge — but the work is light: rewording the changeset, adding a KNOWN GAP comment in CJS, tightening 2 test assertions for canonical-casing, and optionally normalizing `depends_on` output. No structural changes needed.

---

trek-e:

**https://github.com/GSD-redux/get-shit-done-redux/pull/88**

Issue #3657, opened 2026-05-17
verify-reapply-patches: drifts when gsd-pristine/ snapshot is newer than backup-meta baseline


unepod:
### GSD Version

1.42.3

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

20.18.0

### Installation Method

npx get-shit-done-cc@latest (fresh run)

### What happened?

verify-reapply-patches.cjs (Step 5a binding gate)fails with `fail_user_lines_missing`, listing all upstream additions since the backup's original baseline as "missing" lines.
  Root cause: the verifier computes user-added lines as diff(backup, pristine_on_disk), but pristine_on_disk is from a later version than the backup, so the diff is inverted. The user's only real change (one frontmatter line) gets buried in dozens of false positives.

  Manually staging the correct pristine baseline from git (by matching backup-meta.json.pristine_hashes against `git log` blobs) and re-running with --pristine-dir/tmp/correct-baseline produces failures: 0.

### What did you expect?

 verify-reapply-patches.cjs (Step 5a binding gate) passes.

Proposed fix (either is sufficient)
  A) Make verify-reapply-patches.cjs git-aware: when the config dir is a git repo and backup-meta.json.pristine_hashes is present, walk `git log -- <file>` and pick the commit whose blob hash matches the recorded pristine. Use that as the baseline instead of (or in addition to) --pristine-dir.
  B) Have the installer re-anchor backup-meta.json.pristine_hashes to match the refreshed gsd-pristine/ snapshot whenever the snapshot is overwritten, so the two stay in sync.

### Steps to reproduce

  1. Install GSD v1.32.0 locally, modify .claude/agents/gsd-executor.md
     (e.g. add `model: sonnet` to frontmatter).
  2. Run /gsd:update through one or more intermediate versions
     (e.g. 1.32.0 → 1.41.2 → 1.42.3). At each step the installer
     refreshes .claude/gsd-pristine/ but never re-anchors
     .claude/gsd-local-patches/backup-meta.json.pristine_hashes.
  3. Run /gsd:update --reapply on 1.42.3.

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

### Runtime settings.json (if relevant)

```json

```

### How often does this happen?

Every time (100% reproducible)

### Impact

Moderate — Feature is broken but I have a workaround

### Workaround (if any)

Three fix paths, in order of effort:

[...]

PR #3801, opened 2026-05-21
fix(3407): snapshot old release into gsd-pristine, not new

trek-e:
## Linked Issue

Fixes #3407

---

## What was broken

`saveLocalPatches()` wiped `gsd-pristine/` and re-populated it from `pristineCtx.packageSrc` (the **new** release source). For any file that changed between the old and new release, this wrote new-release bytes as the pristine baseline while `backup-meta.json` recorded old-release hashes — a guaranteed hash mismatch that triggered the `OK_PRISTINE_DRIFT_DETECTED` guard on every such file, causing it to skip the three-way diff and fall back to over-broad mode every time a user upgraded with local patches.

## What this fix does

Replaces the wipe-and-repopulate approach with a per-file preservation pass: if `gsd-pristine/<file>` already exists with `sha256 === originalHash` (as correctly set by the prior install), it is kept. If absent or hash-mismatched (stale from a previous buggy run), the file is removed so the verifier falls back cleanly to over-broad mode instead of reading wrong bytes.

## Root cause

`populatePristineDir()` was called in the **pre-wipe** `saveLocalPatches()` call with `packageSrc: src` (the new release). At that point, the correct pristine bytes were already on disk from the previous install. The function ran the install transform pipeline on the new source and wrote new-release bytes, creating a hash mismatch with the old-release hashes recorded in `backup-meta.json`. The `#3657` fix (`OK_PRISTINE_DRIFT_DETECTED`) was a symptom workaround — the root-cause stale write was never corrected until now.

From CONTRIBUTING.md L94–L103: this is a confirmed-bug fix scoped exactly to the reported issue.

### How I verified the fix

Test file `tests/bug-3407-pristine-stale-content.test.cjs` (4 tests):
- **Core regression**: simulate upgrade where file changed between v1 and v2; pre-populate `gsd-pristine/` with old-release bytes; call `saveLocalPatches` with new-release `packageSrc`; assert `gsd-pristine/` still has old-release bytes (sha256 comparison). **Failed before fix, passes after.**
- **Regeneration**: when gsd-pristine/ entry is absent and file is UNCHANGED between releases, hash-validated regeneration restores the pristine entry from new-release source. **Tests the regeneration path added in a31d997.**
- **Stale-pristine recovery** (F4, new): pre-populate `gsd-pristine/` with new-release bytes (exact pre-fix bug artifact); assert entry is ABSENT after fix run (stale detected and removed, regeneration discards mismatched candidate). **Would fail on old code.**
- **First-upgrade safety**: assert that when no prior pristine exists AND file changed, new-release bytes are NOT written as pristine. **Failed before fix, passes after.**

`gsd-test-summary --both` (2026-05-21):
- **Mac**: 3980 passed / 0 failed
- **Docker**: 12047 passed / 0 failed (flaky timing failures in unrelated `init.test.cjs` / `milestone-archive.test.cjs` / `concurrency-safety.test.cjs` are pre-existing on `main` and not caused by this change — zero diff to those files from this PR's commits)

### Regression test added?

### Platforms tested

- [x] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] Issue linked above with `Fixes #3407` — **PR will be auto-closed if missing**
- [ ] `.changeset/` fragment added if this is a user-facing fix — adding after PR opens per CONTRIBUTING.md L140–L153

## Breaking changes

None

---

## OK_PRISTINE_DRIFT_DETECTED decision

**KEPT** — not removed or narrowed.

Rationale: the guard correctly handles two remaining scenarios:
1. **Existing installations** that already have drifted pristine from pre-fix versions of the installer. Those users will still hit the guard on their next upgrade; it protects them from false `FAIL_USER_LINES_MISSING` until they get a clean pristine baseline.
2. **Future stale-pristine sources** — any other code path that could overwrite `gsd-pristine/` with wrong bytes. The guard provides defense in depth.

Removing the guard now would expose users who upgraded to a pre-fix GSD version. The guard's trigger frequency drops to near-zero after this fix lands and propagates.

## Antipattern hunt

Searched all `populate*` / `save*` routines in `bin/install.js` and `get-shit-done/bin/`. No other function conflates "old state to preserve" with "new state to install" in the same way. The Codex pre-install snapshot (lines 8056–8064) correctly reads from `configDir` before any mutation.

`populatePristineDir()` is still exported and functional for use as a future utility (e.g., initial pristine population after a fresh install). Its call site inside `saveLocalPatches` is what was removed.

---

Commit `188bd118` addresses all findings from the sonnet adversarial + gsd-code-reviewer dual review:

- **F1 (MAJOR — sonnet adversarial)**: Moved `mkdtempSync` inside the `try` block with an explicit `catch`/`warn`. Previously, a failure in `mkdtempSync` (e.g., unwritable `/tmp`) would throw an uncaught exception and crash the installer, regressing the pre-fix graceful-degradation behavior. Now logs a warning and continues; verifier falls back to over-broad mode for missing entries.
- **F2 (MIN-01)**: Fixed misleading prose in the inline comment that claimed `gsd-pristine/` is populated "during the install step (after install.js writes files to configDir)". Updated to accurately describe lazy population by `saveLocalPatches`' regenerate branch with an induction-across-upgrades explanation.
- **F3 (MIN-02)**: Fixed counter double-counting and log message. Previously, a file that was stale-removed AND then regenerated incremented both `removed` and `regenerated`. Replaced scalar `removed` with `stalePaths`/`regeneratedPaths` Sets; `removed` is now the count of stale entries that could NOT be recovered (non-overlapping with `regenerated`). Log updated to `Removed N stale gsd-pristine/ snapshot(s); regenerated M of those`.
- **F4 (MIN-03)**: Added stale-pristine recovery test — pre-populates `gsd-pristine/<rel>` with new-release bytes (exact pre-fix bug artifact), runs `saveLocalPatches`, asserts entry is ABSENT (stale detected + removed, regeneration discards mismatched candidate). Would fail on the old code.
- **F5 (MIN-04)**: Tightened Test 3's permissive `if (exists) assert.notEqual(...)` to `assert.strictEqual(exists, false, ...)` — strict failure, clear message.
- **F6 (MIN-05)**: Removed the vacuous "Antipattern hunt" `describe` block (contained only `typeof === 'function'` structural checks, no behavioral coverage). Rationale noted in a comment in the test file.
- **F7 (sonnet adversarial MINOR)**: Changed `fs.rmSync(pristinePath)` to `fs.rmSync(pristinePath, { force: true, recursive: true })` for EISDIR resilience; `removed` counter now only increments after `fs.existsSync(pristinePath) === false` confirms the file is actually gone.
- **F8 (NIT)**: Removed `afterEach` from the destructured `require('node:test')` import (was unused).
- **F9 (NIT)**: Updated PR body "Testing" section test count from "(2 tests)" to "(4 tests)".


trek-e:

## Codex Adversarial Review

Verdict: needs-attention

No-ship: this fix removes stale-pristine corruption, but it also removes practical pristine baseline recovery for missing entries, which can force persistent over-broad verification and block/derail patch reapply on real upgrades.

Findings:
- [high] `saveLocalPatches` no longer rebuilds missing pristine baselines, causing persistent over-broad verification (bin/install.js:7766-7788)
  What can go wrong: for modified files with no existing `gsd-pristine/<file>` snapshot, upgrades proceed without a usable pristine baseline. Why vulnerable: this branch removed the `populatePristineDir(...)` path from `saveLocalPatches` and now only preserves matching snapshots or deletes stale ones; absent snapshots are left absent. The new test also codifies this as acceptable (`gsd-pristine/` may stay empty). Likely impact: verification falls back to backup-only line checks, so upstream removals/rewrites can be misclassified as missing user lines, creating false blocking failures or pressuring users into bad merges to satisfy the gate. Concrete risk is user-visible during update/reapply flows and can recur indefinitely once pristine stays empty.
  Recommendation: Keep the #3407 stale-snapshot guard, but add selective regeneration: run `populatePristineDir` into temp for missing entries and only keep per-file candidates whose SHA-256 matches `backup-meta.pristine_hashes[relPath]`; discard mismatches. That restores precise baselines where safe, avoids new-version contamination, and prevents permanent fallback mode.

Next steps:
- Implement hash-validated per-file pristine regeneration for missing entries in `saveLocalPatches`.
- Add a regression test where `gsd-pristine` is absent but the generated candidate hash matches the recorded pristine hash; assert the file is restored.
- Update stale comments around `saveLocalPatches` call site to match the new control flow and avoid future drift.

## Actions Taken

- **fix applied** (commit `a31d997d`): implemented Codex recommendation — hash-validated regeneration for absent `gsd-pristine/` entries. For each missing entry, `populatePristineDir` runs into a temp dir using new-release source; candidate is promoted only if `sha256(candidate) === originalHash`. If hashes match, the file was unchanged between old and new releases, so new-release bytes are the correct old-release pristine. Mismatches are discarded — over-broad fallback applies cleanly.
- **regression test added** (same commit): `gsd-pristine/ is regenerated for missing entries when file is unchanged between releases` — asserts the recovery path triggers correctly. All 5 tests pass locally (5/0).
- rebased onto main: N (branch was already up to date — 0 commits behind main)

## Open for Maintainer

No design-challenge findings remain open. The Codex `needs-attention` verdict has been resolved by the regeneration path.

trek-e:

Two independent reviewers ran in parallel: `gsd-code-reviewer` (project-aware) and sonnet adversarial (substitute for `codex-rescue`, which is fire-and-forget broken this session). The fix is correctness-sound on the happy path — but the adversarial pass surfaced a real MAJOR that the project-aware pass missed: a removed try/catch.

A follow-up commit is being prepared on this branch addressing the MAJOR + the actionable MINORs.

### MAJOR

- **`mkdtempSync` call outside try/finally — uncaught exception aborts installer.** `bin/install.js:7800` calls `fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-pristine-regen-'))` BEFORE the try block at line 7801. The try/finally guards `populatePristineDir` and the copy loop but not temp-dir creation. If `os.tmpdir()` is unwritable (disk full, restrictive umask, read-only mount), `mkdtempSync` throws uncaught — and the exception isn't handled anywhere in the call chain (`saveLocalPatches` has no outer try/catch, install.js:8088 doesn't wrap it). The pre-fix code had try/catch over the entire pristine block that logged a warning and continued. **This PR silently dropped that resilience for the mkdtempSync call.** An npm install on a machine with a broken temp directory would now crash rather than degrade to over-broad mode. **Remediation**: declare `let tempPristineDir` outside, then move `const tempPristineDir = fs.mkdtempSync(...)` into the try block (with matching `finally` cleanup that no-ops if undefined).

### MINOR

- **Misleading prose about WHEN gsd-pristine/ is populated** (gsd-code-reviewer MIN-01). PR / install.js comment claims "during the install step (after install.js writes files to configDir)". Grep confirms the ONLY writer of `gsd-pristine/` is the regenerate branch inside `saveLocalPatches` itself. The fix works by induction over upgrades, not by a separate install-time population. Update the comment to reflect actual mechanism.

- **`removed` and `regenerated` counters can both fire on the same file** (gsd-code-reviewer MIN-02). Stale → removed → regenerated path increments both. The terminal log "Removed N stale gsd-pristine/ snapshot(s) — falls back to over-broad verify heuristic" is misleading when those files were successfully regenerated.

- **No test for the stale-pristine recovery path** (both reviewers: gsd-code-reviewer MIN-03 + sonnet MINOR). The 5 existing tests cover (1) correct-preserved, (2) missing-regenerated-unchanged, (3) missing-not-written-when-changed, (4)+(5) structural exports. None exercise the branch at `install.js:7786–7790` where `gsd-pristine/X` exists but has the WRONG bytes (the exact pre-fix bug artifact). Add a fixture that pre-populates `gsd-pristine/X` with NEW-release bytes, then asserts `gsd-pristine/X` is absent (or regenerated correctly) after `saveLocalPatches`.

- **Permissive assertions** (gsd-code-reviewer MIN-04). Test 3 uses `if (exists) assert.notEqual(...)` when the fix provably produces `!exists`. Strict `assert.strictEqual(exists, false)` catches more regressions.

- **`removed` counter increments on best-effort rmSync failure** (sonnet adversarial MINOR). If stale `pristinePath` is a directory (e.g., user inadvertently created `gsd-pristine/CLAUDE.md/`), `rmSync` without `{ recursive: true }` throws EISDIR, gets swallowed, `removed++`, log says "Removed N stale" but the directory remains.

- **Antipattern-hunt describe block contains only `typeof === 'function'` structural checks** (gsd-code-reviewer MIN-05). The describe-block name claims behavioral coverage but the tests are vacuous — they would pass even if every function in the module was a no-op `() => {}`. Per `/test-rigor` skill: vacuous-truth pattern. Also: a dead `beforeEach` in that block is never used.

### NIT

- `afterEach` imported but never used in `tests/bug-3407-pristine-stale-content.test.cjs:35` (sonnet adversarial).
- PR body "Testing" section says "2 tests" but the file has 5 (sonnet adversarial).
- Plus two other NITs from gsd-code-reviewer (see findings file).

### QUESTION (out-of-scope but worth documenting)

- **Bootstrap gap for multi-version-skipping upgrades** (sonnet adversarial). The recovery condition is "file unchanged between the last buggy install and the new release". If a user upgrades through multiple versions where the affected file keeps changing (v_old → v_bugged → v_changed → v_new with each step touching the file), they may never get a working three-way merge. The PR body says "next upgrade cycle" — the actual condition is more nuanced. Worth a docs sentence acknowledging the precise recovery condition.

- Per-file decision tree at the regeneration site structurally cannot write new-release bytes when the file changed across the release boundary. ✓
- CJS↔TS drift: no cooperating-sibling pair (`shared-module-handsync-allowlist.json` confirms; sole writer is install.js). ✓
- Consumer (`verify-reapply-patches.cjs`) compatible with new behavior (absent → over-broad fallback; present → hash-validate). ✓
- Symlinks / file modes: no specific finding, but worth a regression test that includes a `+x` file in the fixture set.
- Closing keyword `Fixes #3407` present. ✓

### Recommendation

**Changes requested** before merge. The MAJOR is a real resilience regression — a user with a broken `/tmp` (rare but real, especially on CI runners and locked-down corporate setups) goes from "warning + degrade" to "installer crash". The MINORs around test coverage of the stale-pristine path and the misleading log/comment text are also worth addressing in the same commit.

---

trek-e:

**https://github.com/GSD-redux/get-shit-done-redux/pull/87**

